PR #11, opened 2017-10-09
fix bug in --out flag of export

tsatsoulis:
Fix bug when using the --out flag in export --bibtex to correctly append the output to a file.


alejandrogallo:
Excellent catch! Thank you very much!

PR #12, opened 2017-10-15
Fix the filetype buffer recognition

tsatsoulis:


alejandrogallo:
Yes that is right, before it would really fail to recognise if it is a binary buffer or a path,
it would complain in os.path.exists, good catch! This deserves a new version.

PR #13, opened 2017-11-11
Add limited use of mimetypes for file extensions

trueselfdao:
Hi

Turns out it was somewhat less useful than I anticipated. Since
[mimetype -> extension] is a one-to-many map, we can only use it
when the mapping is unique as with pdf and djvu files.

However, the non-mimetype output of magic.from_file() provides more
useful information. May consider using this and just parsing output.


alejandrogallo:
Hello, first of all thank you very much for your great work, I really appreciate it.
I have some questions and requests:
  - Have you checked out the functions at the end of ``papis.util`` to try to merge them

trueselfdao:
I'm new to mimetypes as well so I'm also figuring this out

- I was also thinking of placing this in `papis.utils` (see the TODO in the pull request). Seems we're on the same page. 

  Not sure about the standard way to get file types but parsing the mimetype works to some extent. However `chm` and `mobi` files both have the non-descriptive mimetype `application/octet-stream`. The more descriptive information (passing `mime=False`) from magic seems to be enough (`MS Windows HtmlHelp Data` vs `Mobipocket E-book...` respectively).

- A brief search hasn't led me to any definitive standard, but [Wikipedia media-type](https://en.wikipedia.org/wiki/Media_type) and [Debian-MIME](https://wiki.debian.org/MIME) claim this to be a UNIXism.

- Alright. Consider adding a linter to the tests and a CONTRIBUTING file to project root.

For starters, I'll merge mimetype support into `papis.utils` and see what difficulties arise when determining file type and extension.

Cheers

alejandrogallo:
Hi! Do you need some help? Feel free to engage in a discussion!

alejandrogallo:
I'll be closing this for the moment

PR #15, opened 2017-11-13
New mutt integration

michaelplews:
This PR allows you to access papis from within mutt, which required a new flag (`--attach`) to be added to `papis open`. 

While a similar effect could be achieved by writing a papis-script, this (I believe) requires the least amount of code changes. However, if you can think of a better way to do this I am all ears.

Ignore a5c70d0 and e8c450e, I added them to the wrong branch.


alejandrogallo:
Hi Michael, it is a great idea and I think your solution is wonderful. Specially also thank you
for including documentation.

I have one thing that I want to ask your opinion though. Maybe it is more consistent
to transfer the behaviour of your
```
papis open --attach
```
into the `export` command, because it is more like "exporting" at the end, so maybe you could implement
```
papis export --file --out /tmp/papis.pdf
```
instead, or do you think people would find it more natural to do as you first thought?
It would go in line with the already existing behavior of exporting a folder
```
papis export --folder --out folder_name
```
Thanks again!

michaelplews:
Alejandro, you're absolutely right. I will make those changes and update the PR.

Thanks for your input!

alejandrogallo:
Wonderful ! Thanks!

PR #25, opened 2017-12-07
papis-scihub: Add new scihub domain (.tw)

nico202:
Today sci-hub io, bz, biz, ac and cc seems down, while tw is working


alejandrogallo:
Wonderful, thanks!

PR #32, opened 2017-12-11
New config flag: ref-format

### LINKED ISSUE #16: Allow `ref` preferences to be set in config
Some of the larger bibliography programs allow the style of the `ref` flag to be set in Preferences. The `ref` flag is used in referencing entries from bibtex files, and would be useful to remember off hand. Personally, I use the style AuthorYear e.g. `Plews2019`. It may be useful to allow users to add a preferred format which is used if the information is available when using the `--from-doi` flag (or similar).

An Example:
```
[settings]
ref-format = {doc[author]}{doc[year]}
```
Right now it defaults to the doi, which is good fallback behaviour.

I have a branch I was going to PR where I added a `--ref` flag (310e096), but this might be worth adding. I'm happy to work on this, I just can't figure out where/how you pull information from the config file, and how I would translate it into `papis add`.

michaelplews:
This PR brings the implementation of the `ref-format` flag. This allows the user to set their preferred format to reference when using in conjunction with .bib files. Or for remembering a article by an easy to remember combination of information (e.g. AuthorYear) rather than the doi.

The default behaviour is for the ref to be set to the doi.

Closes #16

### DIFF
--- a/papis/config.py
+++ b/papis/config.py
@@ -1,5 +1,4 @@
-"""
-General
+"""General
 *******
 
 .. papis-config:: local-config-file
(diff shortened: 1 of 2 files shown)


alejandrogallo:
Looks fine to me!
Thank you very much Michael!

PR #36, opened 2017-12-15
Introduce file-name option and flag

### LINKED ISSUE #22: pdf import: rename file in a meaningful way
When sharing all the papers, with zotero I did "find zotero_dir -iname "*.pdf" -exec cp {} output_dir \;".
Using ZotFile, the files where named after the paper year - author - title.

Now with papers downloaded from scihub, for example, I have a lot of "output.pdf". This is ugly also if using the "history" function of pdf viewers and so on, since it's difficult to remind which paper was it (on emacs interleave, notes are named like the source pdf, so having almost-unique name is a must).

I'd suggest to name them with doc[ref] (especially if it will be possible to customize the ref format) upon import and then don't touch them anymore

michaelplews:
This PR creates the `file-name` option and flag. 

`file-name` defaults to `ref-format`. Users can set `file-name` in the config file (papis-format) just like other options. If multiple files are imported, filenames are amended so to not overwrite other files. Closes #22.

`ref` tags now make themselves unique (an update of #32)


michaelplews:
Huh. Travis CI failed due to `itertools NameError`, but I never call itertools directly. Is it because of the example in the docstring? The only other time it's called is `form itertools import product, count`.

alejandrogallo:
Nice! Hmm yeah it is basically the example in the docstring that is creating this error, you
can simply import it before the doctest line where you use the `itertools` package, in principle
if you change it in your master and push it to your master then the pull request will be updated automagicallly I think.

alejandrogallo:
Hi Michael nice thanks.
I really like the approach with the generators and using the next iterator function.
 I was reading the commits and I'll change a couple of things:
- Merge the last two commits into one (git rebase)
- Some linting.
- There is a problem when giving it the ``--file-name`` flag, because you are not formating
  the string right now
```python
            # Rename the file in the staging area per options or flags
            if self.args.file_name: # Use args if set
                new_filename = papis.utils.clean_document_name(
                    self.args.file_name
                )
            else:                   # if not use naming format
                new_filename = papis.utils.clean_document_name(
                    self.get_file_name(data) + string_append + '.pdf'
                )
            new_file_list.append(new_filename)
```
and therefore you just need actually to set the config value to the passed arg
```python
if self.args.file_name: # Use args if set
            papis.config.set("file-name", self.args.file_name)
```
for instance.

- Also all this should be done before the ``--edit`` flag ,so that you see the names if you edit.
- I do not think it is a good idea to put the ``ref-format`` as a default because it will break
if you try to add a document just with author and title and no other information. So I will just leave
that the default behaviour is just cleaning the names of the files you're actually getting.
- I also think it is a good idea not to suppose that the documents are `pdf`s, so I wrote also a function
go guess the extension of the to-be-added docs.
- I also think it is a good idea to have the suffixes separated by a `-`.

I hope it is ok for you that I did these changes, we can discuss about it and revert it if you think it does
not make any sense ;)

Thank you for your great work!

alejandrogallo:
Hi Michael, I did not know I can push directly to your branch on github, so I merged your commits manually  and the commit hashes changed, so your code is already in the master, can you cancel then this pull request? sorry for the inconvenience

michaelplews:
Sounds good. Glad everything is sorted out. Thanks for your help and input.

PR #49, opened 2018-02-10
Fixes -c not working

nico202:
on master -c option was not working, and "Overwriting previous configuration was printed".

### DIFF
--- a/papis/config.py
+++ b/papis/config.py
@@ -63,7 +63,7 @@
 .. papis-config:: xeditor
 
     Sometimes papis might use an editor that uses a windowing system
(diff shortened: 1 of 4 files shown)


nico202:
No wait still not working, will fix it

nico202:
The problem is, during the import of `bibtex`, `utils` and others you are calling `papis.config.get`, which calls `get_configuration`, which creates `CONFIGURATION`. I fear a great part of the logic should be changed in order to allow custom configs

alejandrogallo:
We have to take a look at this, let's see if I manage today to push the whoosh backend for the database ;)

Issue #21, opened 2017-11-23
Rofi GUI open wrong file, sometimes


nico202:
Hi, I don't know how to debug this right now. However, opening a file with the gui --rofi not always open the correct paper (both for edit/show file) while the vim GUI for the same entries open the correct file.

I'll look if I can reproduce this with a smaller library

alejandrogallo:
Thanks, I'll keep an eye on it, the thing is that I do not use the rofi gui so much, and I was thinking
about reimplementing it in form of a papis external script, but I'll use it now more regularly to see if 
I also notice something weird.

alejandrogallo:
I'll go ahead and close this issue for the time being, we can open it again when
we find some similar behaviour again.
Thanks!

nico202:
I'm still experiencing it, I'll try to find out a simple way to reproduce and why it's happening tomorrow. The issue should be re-opened

alejandrogallo:
Ok! Let me know!

nico202:
Just this two:

```
author: Ritterbush, Jon
title: Supporting Library Research with LibX and Zotero
year: '2008'
```
```
author: Xue Lu and Niall Adams and Nikolas Kantas
title: On Adaptive Estimation for Dynamic Bernoulli Bandits
year: '2017'
```

are enough to make rofi gui go cazy (it show 4 elements)

PR #52, opened 2018-02-11
Fix rofi + housekeeping

nico202:
Fix issue #21.

Also delete some extra whitespace and spellcheck config. I'm spellchecking files as I edit them so probably more of those to come


nico202:
The "fix empty selection crash" is what happens when you press esc on papis open with rofi

alejandrogallo:
Thanks a lot nico!

PR #53, opened 2018-02-11
link DOIs to preferred resolver

katrinleinweber:
Hello!

The DOI foundation recommends a [new, secure resolver](https://www.doi.org/doi_handbook/3_Resolution.html#3.8).

Cheers :-)

### DIFF
--- a/papis/crossref.py
+++ b/papis/crossref.py
@@ -195,7 +195,7 @@ def validate_doi(doi):
     :type  doi: str
     """
     from urllib.request import urlopen, Request
-    handle_url = "http://dx.doi.org/" + doi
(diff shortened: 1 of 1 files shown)


alejandrogallo:
Hi Katrin, thanks so much!
Cheers!

katrinleinweber:
You're welcome :-)

katrinleinweber:
PS: Does the function docu need changing as well? https://github.com/papis/papis/blob/f8f57cc30b8885b24d5f5804578458bb7f548308/papis/crossref.py#L408-L410

alejandrogallo:
That's right, you cann add a doctest if you want without the dx, but leave the dx test also in place.
Thanks again!

PR #55, opened 2018-02-15
NixOS repository package

nico202:
papis is now on nixpkgs! It's on the unstable channel (meaning that it's available to those that are on the bleeding edge, while I think it will become available on stable release that usually happens on March). 

I also removed a couple of whitespaces and fixed a wrong capital word (added the reference to the grammar, I'm not english but I'm quite sure it's right)


alejandrogallo:
It sounds great!! Thank you very much!

nico202:
Ops there was a typo, I meant "if you *like* papis (instead of run)"

I'm not at my computer now I can't fix it

alejandrogallo:
Don't worry, I just fixed it ;)
Thanks!

PR #56, opened 2018-02-17
[WIP] Adding zotero-connector support

nico202:
# WORK IN PROGRESS

Hi, I noticed that zotero translators are inside the connector, so I thought we can gain a lot from the connector itself (the firefox plugin).

This is a draft on the implementation of `papis serve`, a command that create an http server on the same port zotero does and listen for zotero-connector requests. The great part of the server itself is implemented and is in a working state.

I'm opening the PR because I need help on the papis part. Here the function `papis_add(item)` is passed a dictionary containing info like author, title and so on, we should call here the papis method to add the info.yaml, but I still don't know how to do this, if you could point me to the right direction.

You can start testing it by installing the connector from [here](https://www.zotero.org/download/). If you go to a supported site (like scholar.google.com or almost any scientific journal website) and click on the connector icon.

Thanks, Nicolò

### DIFF
--- a/papis/commands/__init__.py
+++ b/papis/commands/__init__.py
@@ -25,6 +25,7 @@ COMMAND_NAMES = [
     "run",
     "git",
     "gui",
+    "serve",
     "sync"
 ]
 
--- a/papis/commands/serve.py
+++ b/papis/commands/serve.py
@@ -9,10 +9,12 @@ https://raw.githubusercontent.com/zotero/zotero/master/chrome/content/zotero/xpc
 """
 
 import json
+import os
 import argparse
 import logging
 import http.server
 import papis.config
+import papis.document
 
 logger = logging.getLogger("serve")
 
@@ -61,8 +63,13 @@ def papis_add(item):
     # This should just call a papis function that given a dict of
     # properties spit out the correct info.yaml and saves it.
     # I still have to figure out where PDFs are taken
-    print(item)
-    print("PAPIS add NOT IMPLEMENTED.")
+    logger.debug("Adding the document to your library")
+    document = papis.document.from_data(item)
+    lib_dir = os.path.expanduser(papis.config.get('dir'))
+    outdir = os.path.join(lib_dir, papis.utils.clean_document_name(item["title"]))
+    os.makedirs(outdir)
+    document.set_folder(outdir)
+    document.save()
 
 # HTTPRequestHandler class
 class PapisRequestHandler(http.server.BaseHTTPRequestHandler):
@@ -128,8 +135,6 @@ class PapisRequestHandler(http.server.BaseHTTPRequestHandler):
         data = json.loads(self.read_input())
         source_uri = data['uri'] # source page
         # debug info
-        print(data.keys())
-        print(len(data['items']))
         # Add all papers
         for item in data['items']:
             papis_add(item)
--- a/papis/commands/serve.py
+++ b/papis/commands/serve.py
@@ -33,7 +33,8 @@ class Command(papis.commands.Command):
             "--port",
             help="Port to listen to",
             action="store",
-            default=23119
+            default=23119,
+            type = int
         )
 
         self.parser.add_argument(
--- a/papis/commands/serve.py
+++ b/papis/commands/serve.py
@@ -1,0 +1,161 @@
+"""Start a web server listening on port 23119. This server is
+compatible with the `zotero connector`. This means that if zotero is
+*not* running, you can have items from your web browser added directly
+into papis.
+
+This is a rough port of
+https://raw.githubusercontent.com/zotero/zotero/master/chrome/content/zotero/xpcom/server_connector.js
+
+"""
+
+import json
+import argparse
+import logging
+import http.server
+import papis.config
+
+logger = logging.getLogger("serve")
+
+class Command(papis.commands.Command):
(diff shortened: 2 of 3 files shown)


nico202:
Ok now adding works, even if I don't know how to call the folder, but I like the 1-click add

alejandrogallo:
Hi Nico, sorry for the delay. This is really awesome, I'm going to create a `serve` branch in the main repository and we can work on it

nico202:
TODO:
- [ ] change the command name?
  serve might be used to share the library itself in the future 

- [ ] Add web page screenshot capabilities
  If the connector finds a page that it's not/ he think it's not a paper, it will ask us to save the page itself, do we want this? Is there a simple way to download a webpage with images/css and save it to a single file from python without having big dependencies?

- [ ] Manage downloads
  files to download are now saved under 'attachments' as:
  mimeType: application/pdf
  title: Fulltext
  url: http://link.to/file.pdf
  
  we might want to replace this with the actual downloaded file

- [ ] Manage duplicates
  This is probably the easiest thing but I should better read the papis code to discover how to manage this. It might be useful to refactor a bit the `add.py` code, I see many things that might be moved to their own functions

alejandrogallo:
I agree on everything. I'm very excited of what it will be able to do and also it will be
useful for implementing some of the rpc methods for the server that will be used for the webapp.
I am quite disappointed though that the zotero connectors do not download the pdfs,
I thought they also downloaded the pdfs and so on, do they?
Many journals need the robots confirmation and this is very annoying... And we could set scihub as a flag, that's for sure.

About the `add.py` refactor, the method `papis_add` that you implemented in `serve.py` rings a bell for me... I have implemented this a couple of times now, and this is bad. I think we should separate the cli parsers from the functionalities, don't you think? This is, there should be a python function or module that does the adding in general, and then the cli connects the user input (through the command line) to the more general function. This could be also then recycled by all guis and also used in user scripts.
So I think this is something we have to do for all commands, does this make sense?

alejandrogallo:
Hi, as you can probably see I have been busy refactoring the commands on the refactor branch.
Also the tests for the databases are ready, I just have to get up-to-date to the documentation and I
think it would be ready for a v0.6.0 for papis, where different databases are supported.
What do you think should come also to the v0.6.0?

michal-h21:
I was going to write support for [Zotero translation server](https://github.com/zotero/translation-server) and then I found this project, which uses much more sensible approach, as Zotero translators in a browser are more powerful than the standalone version. It is really important, because Zotero supports huge number of sources of information, from library catalogues to information in blog posts.

I've tried to run the `papis serve` command from the `serve` branch. It runs after commenting our `sync` command in `papis/commands/__init__.py`. Here are some notes:

- the library is not updated  after addition of a new document, so it is not visible in `papis edit` and other commands
- the directory names are a different  (`papis add` uses `hash-author` scheme, `papis serve` use `title`)
- the `ref` field is not created (when `ref-format` is set in the config file)

There is also issue that Zotero JSON uses quite different data scheme than Papis expects, for example different fields are used for author and year, so `papis edit` is not able to print them. It is also difficult to export it to BibTeX, the author list is missing. 
I've tried  to set the `author_list` field, but it isn't exported to BibTeX. 

Here is my updated `papis_add` function:

```
def papis_add(item):
    # This should just call a papis function that given a dict of
    # properties spit out the correct info.yaml and saves it.
    # I still have to figure out where PDFs are taken
    logger.debug("Adding the document to your library")
    if "creators" in item.keys():
        authors = []
        for author in item["creators"]:
            if "firstName" in author.keys() and "lastName" in author.keys():
                authors.append(
                    dict(given_name=author["firstName"], surname=author["lastName"])
                )
        item["author_list"] = authors
        item["author"] = ",".join(
            ["{a[given_name]} {a[surname]}".format(a=a) for a in authors]
        )
    item['ref'] = papis.utils.format_doc(papis.config.get("ref-format"), item)
    document = papis.document.from_data(item)
    lib_dir = os.path.expanduser(papis.config.get('dir'))
    outdir = os.path.join(lib_dir, papis.utils.clean_document_name(item["title"]))
    os.makedirs(outdir)
    document.set_folder(outdir)
    document.save()
    papis.api.clear_lib_cache()
```

It sets up the `author_list` and `ref` fields and clears the cache, so the document shows up in the search. I am not sure if it is a correct way.

alejandrogallo:
Hi @michal-h21, thanks so much for your interest. Yes indeed this an important todo and I'll be taking a look at it from now on since I was one month away.

michal-h21:
Thanks @alejandrogallo, you are doing great job!

michal-h21:
Hi @alejandrogallo, what is the status of the Zotero-connector support?

alejandrogallo:
Hi @michal-h21 , sorry for the delay, now v0.7 is out as you see, the conectors will be the next step,
it would be a separate script too, so that people can install it separately also using pip3, maybe even include the firefox plugin conector from zotero ?

michal-h21:
Thanks for the info @alejandrogallo, that sounds like a great plan!

alejandrogallo:
It should be working now

https://github.com/papis/papis-zotero

The papis-zotero project however could benefit of active zotero users and people
willing to use the https://github.com/urschrei/pyzotero project in the `papis-zotero`.

PR #57, opened 2018-02-17
NEW: Add Downloader for IEEEXplore

dbruggner:
Download document and bibtex from IEEEXplore
papis add --from-url http://ieeexplore.ieee.org/document/7353481/


alejandrogallo:
Awesome, thanks! Also about the cookiejar!

PR #68, opened 2018-05-29
Removal of whitespace from ref tags

### LINKED ISSUE #67: 'ref' tag with spaces causes errors on LaTeX compiling
Bibtex needs refs that do not contain whitespace. Refs should be collapsed to remove whitespace during `papis add`

I can take care of this.

michaelplews:
ref tags with whitespace cause errors if they are used for
bibtex. Now, ref tags are run through command `collapse_whitespace()`
to remove any whitespace put in by the user (in ~/config/papis) or
from anything pulled from scrapers.

Fixes #67


alejandrogallo:
There is a lot of commits here that I have the feeling have nothing to do with the removal of whitespaces.
Could you create an ad hoc branch with a couple of commits introducing the change ?

michaelplews:
You're right, my bad. I'll close this and open a new one with the relevant commits

PR #69, opened 2018-05-30
Improved search citations (hubation)

### LINKED ISSUE #65: Search citations
I see that some papers I add `--from-doi` have a list of citations in the info.yaml. It would be awesome if, from gui or command line, could be able to see references with the picker. Hit enter, and, if you already have them in your library open the file, else open a web browser of `https://dx.doi.org/[DOI]`. 

Might take a bit of work for the extra functionality but I wanted to add it while I remembered.

michaelplews:
1.  `papis hubation` now searches the users library to see if
documents are already present.

2. Some users are unable to use the scihub script. These
modifications to the hubation script allow users to tell hubation what
to do after a citation is picked, defined in `config` as
'hubation'. So far, `open`, `browse`, and `export` have been tested
and are working. With changes described in (1), `open` will try to
open the document file if the document is already present in the users
library. If not, a browser is opened using the scraped url.

3. Added `has_citations` property to `papis.document` to check if the
document entry has citations associated with it. This can be set in
any 'header' option and returns 'citation-string' if True.

4. Added 'citation-string' config option. If `has_citation`, defined in
(2), is True, this string is returned allowing an indicator to show up
if a document has citations.

Closes #65


michaelplews:
It looks like I did a similar thing here. Let me make a better branch and a PR for that.

Thanks for your patience!

alejandrogallo:
Thank you Michael! Take your time! ;)

Issue #66, opened 2018-05-05
Why does papis browse not default to using the doi?


michaelplews:
Is there a reason that `papis browse` tries to use different methods (pubmed, airvx, etc.) to open the paper in a browser rather than using the digital object identifier (DOI)? DOIs are designed to last for a very long time, if not essentially forever. It also, in some cases, does not redirect to the original article location.

 Example:
```bash
$ papis browse Cha2008
# opens https://www.ncbi.nlm.nih.gov/pubmed/18062687
```

[This is only the abstract](https://www.ncbi.nlm.nih.gov/pubmed/18062687), not the full article.

```bash
$ papis export Cha2008 --yaml | grep 'doi:' | awk '{print "https://doi.org/"$2}'

https://doi.org/10.1021/ic701570z
```

[This redirects to the original article](https://doi.org/10.1021/ic701570z)

EDIT: At the time of writing, `pubs.acs.org` is down.

michaelplews:
Hi Alejandro, 

I like this idea. I also like the idea of adding `browse-doi = True` option to the config so I can set it globally. I can work on this.

PR #70, opened 2018-05-31
Allow browse-key to be set in config

michaelplews:
`browse-key` lets the user define the key that `papis browse` uses when constructing the URL.

Solves #66


alejandrogallo:
Hi Michael,

thanks, it's a little bit unclear some commits, I'm merging the changes in 7b8efec

Cheers!

PR #74, opened 2018-06-05
Added `bibtex-journal-key` option

### LINKED ISSUE #73: Add bibtex-journal-key option
## Problem
Currently, `papis export --all --bibtex` shows the full journal title in the `journal` key. When writing articles, some publishers prefer journal abbreviations in the bibliography.

## Solution
There should be a config option `bibtex-journal-key`, and, possibly, a command line option, to export bibtex with the journal key defined by the user, allowing bibtex files to be made with the journal as either `full_journal_title` or `abbrev_journal_title`. Default should just be `journal` seeing as it will need to be backwards compatible with entries not containing the new fields. 

I can take care of this

michaelplews:
`bibtex-journal-key` lets the user define the key that `papis
export --bibtex` uses when assigning the `journal` bibtex
entry. This allows the user to define `full_journal_title` or
`abbrev_journal_title` depending on the publishers requirements.

Closes #73

### DIFF
--- a/papis/config.py
+++ b/papis/config.py
@@ -168,6 +168,15 @@ Tools options
 Bibtex options
 ^^^^^^^^^^^^^^
 
+.. papis-config:: bibtex-journal-key
+
+    Journal publishers may request abbreviated journal titles. This
+    option allows the user to set the key for the journal entry when using
+    ``papis export --bibtex``.
+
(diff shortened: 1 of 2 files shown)


alejandrogallo:
Thanks, useful!

PR #79, opened 2018-06-19
papis-abbrev script

### LINKED ISSUE #78: Do our own journal abbreviations?
Hi @alejandrogallo, 

This one is more of a question/idea. I've noticed that some of the journal abbreviations picked up from crossref are incorrect sometime. e.g.

```bash
papis add --from-doi '10.1038/ncomms11683'
```

results in `abbrev_journal_title: Nat Comms` which should be `Nat. Commun.`

I think Zotero and other citation software calculates their own abbreviations. I wondered whether this is a direction you think papis should go in?

michaelplews:
papis-abbrev takes 'full_journal_title' field and attempts to
create 'abbrev_journal_title' via [ISSN standards](http://www.issn.org/services/online-services/access-to-the-ltwa/). To use, users
must copy (or symlink) `papis-abbrev` and `LTWA.json` to their
`~/.config/papis/script` directory.

The script runs as expected, allowing users to pick documents and use
search functions. However, doing `papis abbrev '' --all` allows all
documents in the library to be updated through papis-abbrev.

Closes #78


alejandrogallo:
Hi michael, thank you very much for this, I'll take take a look at it at some point.
Take care!

PR #85, opened 2018-07-28
[hubation] Added an optional reference number argument

### LINKED ISSUE #84: [hubation] Implement immediate reference redirecting
## Problem
Right now, If I use the `hubation` script to search the references, I have to go to the references section of the journal, look up the bibliographic information, run `papis hubation`, and then search the results for the bibliographic information.

## Solution
I've noticed that the references caught by `hubation` are usually in order of how they appear in the paper. It might be useful to add an additional argument e.g. `papis hubation ref='Plews2019' 28` where 28 is the inline reference. Papis could then download information for all the references, stopping when it got to 28 and then returning the result.

I can work on this.

michaelplews:
Now by adding an optional reference number from the article as a
command line argument, papis hubation will stop searching when it gets
to that reference and return the result.

Closes #84

Also some other minor bug fixes

### DIFF
--- a/examples/scripts/papis-hubation
+++ b/examples/scripts/papis-hubation
@@ -46,7 +46,7 @@ doc = papis.api.pick_doc(
         documents
     )
 
-if 'citations' not in doc.keys():
+if 'citations' not in doc.keys() or doc['citations'] == []:
     print_no_citations_and_exit()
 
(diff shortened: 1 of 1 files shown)


alejandrogallo:
Thanks as always

PR #101, opened 2018-09-05
Changed 'read and write' to just 'read'

### LINKED ISSUE #100: papis add has permission issues
`papis add` is causing permission issues when reading from `/tmp/`. 

https://github.com/papis/papis/blob/aeddcaf857488d37bb10de13da23dbf63b19c707/papis/commands/add.py#L160

Asks to open in binary read **and write** mode, when that isn't necessary.

michaelplews:
fixes #100


alejandrogallo:
True! Thanks for catching that one!
Cheers!

PR #111, opened 2018-10-05
Fixed zotero-sql to work for 5.0.55.1

avonmoll:
- zotero-sql was not working for me because my SQL database had a slightly different structure
- After inspecting my database I made some changes to some of the table names in the script, but then the author list was out of order
- Then, updated the ORDER BY criteria for creators

### DIFF
--- a/examples/scripts/papis-zotero-sql
+++ b/examples/scripts/papis-zotero-sql
@@ -77,20 +77,19 @@ def getCreators(connection, itemId):
   itemCreatorQuery = """
     SELECT
       creatorTypes.creatorType,
-      creatorData.firstName,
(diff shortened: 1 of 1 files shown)


alejandrogallo:
Is it backwards compatible?

avonmoll:
I'm assuming that it's not backwards compatible since the script, as is, doesn't seem to match the zotero database structure in 5.0.55.1 (which is why I fiddled with it in the first place). I'm not sure what the right thing to do here is.

alejandrogallo:
I think that's alright, then people should update zotero before importing it, I think it's fair to say
that they should have the latest zotero Zotero.

Thank you very much again!

PR #112, opened 2018-10-05
Import zotero collections to project

avonmoll:
- When importing from Zotero SQL, the items' 'collections' should also be imported
- Collection in Zotero corresponds to `project` metadata for papis
- Does not respect collection hierarchy from Zotero (projects are all at the same level in papis)
- Treats collections with the same name as the same collection


alejandrogallo:
Thanks!! Looks awesome!

PR #115, opened 2018-10-11
Add support for conference papers to crossref parsing

ARamsey118:
Conference papers, such as [this one](https://ieeexplore.ieee.org/document/6758978) currently return 
`Can't locate metadata`. This allows them to be parsed by crossref.py and added to papis like journal papers.


alejandrogallo:
Sounds good! Thank you and sorry for the delay in the merge!

PR #117, opened 2018-10-15
Fixed DocMatchers query string for jinja2

avonmoll:
`DocMatcher` needs to be able to handle jinja2 templating when a query of the form `key=value` is given to a papis command.


alejandrogallo:
Thanks so much and sorry for the delay in the merge

PR #120, opened 2018-10-22
Extended config parser interpolation

avonmoll:
When using jinja2, it's nice to be able to use conditionals `{%- if %}`, etc., for `add-name` and `file-name` for example. However, the `%`s are used by `configparser`'s basic interpolation scheme. In order to deconflict, we can have `configparser` use extended interpolation, which utilizes the `$` symbol.

### DIFF
--- a/papis/config.py
+++ b/papis/config.py
@@ -988,7 +988,7 @@ class Configuration(configparser.ConfigParser):
     logger = logging.getLogger("Configuration")
 
     def __init__(self):
-        configparser.ConfigParser.__init__(self)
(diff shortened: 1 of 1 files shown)


alejandrogallo:
Thanks for your contribution! I understand.
This would break backwards compatibility for already existing configurations.
Have you tried writing `%%` instead ofr `%` ? I would not recommend breaking backwards compatibility.

avonmoll:
That works for me! I should have read the configparser docs a bit more closely before submitting a PR 😄

alejandrogallo:
Don't mention it! ;)
It's nice to get PR anyways, thank you so much for your engagement.

PR #121, opened 2018-10-26
Slightly more robust bibtex handling

avonmoll:
I was having issues where I had papis field names that didn't match a corresponding bibtex key. So I made two conversion dictionaries (one for types, one for fields). Also, since I'm using XeLaTeX most of the time I added a configuration variable to allow unicode in the bibtex output.

### DIFF
--- a/papis/bibtex.py
+++ b/papis/bibtex.py
@@ -24,6 +24,12 @@ bibtex_types = [
   "unpublished"
 ] + re.sub(r" *", "", papis.config.get('extra-bibtex-types')).split(',')
 
+bibtex_type_converter = {
+  "conferencePaper" : "inproceedings",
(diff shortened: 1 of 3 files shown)


avonmoll:
I think that the reason this is necessary is for use with the Zotero browser connectors (i.e. using `papis-serve`). If it were only a matter of cleaning my library (which I migrated from Zotero) I'd be fine with writing a script.

alejandrogallo:
Hi Alexander, thank you very much for this feature.
I added some basic documentation for `bibtex-unicode` and then  I merged it.
It makes a lot of sense to implement this anyways, so thank you again.

I will be releasing `v0.7.5` today since we have some bug fixing since last version.

PR #123, opened 2018-11-13
Added `papis-tags` script

### LINKED ISSUE #122: Picker for tags
It would be really nice if we could have a picker for 'tags'. For example:

`papis tags` would allow picking of all tags used in a library. Once picked, the entries tagged with that tag are shown.

Is there a quick script we could write for this or would it need some more attention?

michaelplews:
The papis tags script adds the ability to pick by tags from a
    library. Closes #122


alejandrogallo:
Thanks, that's nice, as you see I also changed some things in the script, to simplify some parts
I hope you like it ;)

Issue #141, opened 2019-03-05
feedback on first try


teto:
I've tried papis a bit in real conditions. Here is some feedback, sometimes subjective, sometimes objective, sometimes actionable, sometimes without a clear path to resolution.

1. ~~my main concern is that upon `papis export --bibtex --all` citation keys are generated even though I imported from zotero. I could have coped with it if I were starting from a cleanslate but right now this breaks the compilation of all my papers. The keys got exported from zotero `extra: 'Citation Key: Heuschkel2018'` but are not used. I wonder if this actionable ? or if the generated citation key is part of papis' logic ?~~
2. ~~in nixos, usually it's enough to put the zsh completion in share/zsh/site-functions/ and it's picked up by zsh.
With papis, I added  `eval "$(_PAPIS_COMPLETE=source_zsh papis)"` on the advice of @dtz but it gave me some python errors on spawning shell (which I can't find it anymore), I actually wonder if this why my shells are now slow to start. I might need to investigate.~~
3. ~~I tried running `papis update --from-doi` and it asked me if I wanted to override fields even when they were equal.
For instance I got asked `skip user conflict (year conflict) Replace '2018' by '2018'`~~
4. ~~in the documentation, it says cache-dir is None by default but the default is $XDG_CACHE_HOME. Maybe it doesn't reflect the code but I think it should.~~
5. ~~I would really prefer if running just `papis` would print help instead of doing nothing. Might be just me though.~~
6. I use a solarized theme for my terminal and papis colors look really bad, to the point it's kinda unreadable. I don't think the default config could accomodate everyone but something I usually enjoy is when there is an exhaustive config with all options available somewhere. wireshark does it. Astroid generates such a config as well when none exists. It makes it easier to discover options. Also not everything is overridable in the TUI theme so I still have blue text on bluish background xD
7. Because I am used to fzf, I wanted to use it as a picktool but from the doc it was not clear how picktools work or how I can add one. I was expecting papis to just pipe everything out but seems like you need some adapter ? The help message:
```
This is the program used whenever papis asks you to pick a document or
options in general.
```
8. ~~First I tend to share my dotfiles across computers so I ended up with a papis config that pointed at a non-existing folder. I originally thought the folder existed so I ended up confused when running `papis open`:
`No documents found with that name.` 
I thought ok let's try verbose mode:~~
```
52-DEBUG:config:Merging configuration from /mnt/ext/Zotero/Documents/.papis.config
54-DEBUG:db:cache:Initializing
[...]

PR #144, opened 2019-03-13
ref-format: use dynamically

teto:
I would like to change the ref according to ref-format dynamically rather than on import
see https://github.com/papis/papis/issues/141.

For the behavior, I wonder what to do if it can't generate the ref : fallback on a ref if it exists but otherwise ? use a default key "unknown" ?
```
Using ref-format = {doc[author_list][0][surname]}{doc[year]}
on <papis.document.Document object at 0x7f2abbe9d0b8>
  File "/run/user/1000/tmp.6p8bvkMjc8/bin/papis", line 11, in <module>
    load_entry_point('papis', 'console_scripts', 'papis')()
  File "/home/teto/papis/papis/main.py", line 10, in main
    papis.commands.default.run()
  File "/nix/store/83glni8a99pkal667qckk0ibb6rc0jca-python3.7-click-7.0/lib/python3.7/site-packages/click/core.py", line 764, in __call__
  File "/nix/store/83glni8a99pkal667qckk0ibb6rc0jca-python3.7-click-7.0/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/nix/store/83glni8a99pkal667qckk0ibb6rc0jca-python3.7-click-7.0/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
  File "/nix/store/83glni8a99pkal667qckk0ibb6rc0jca-python3.7-click-7.0/lib/python3.7/site-packages/click/core.py", line 956, in invoke
  File "/nix/store/83glni8a99pkal667qckk0ibb6rc0jca-python3.7-click-7.0/lib/python3.7/site-packages/click/core.py", line 555, in invoke
  File "/home/teto/papis/papis/commands/export.py", line 195, in cli
    text=text
  File "/home/teto/papis/papis/commands/export.py", line 100, in run
    papis.document.to_bibtex(document) for document in documents
  File "/home/teto/papis/papis/commands/export.py", line 100, in <listcomp>
    papis.document.to_bibtex(document) for document in documents
  File "/home/teto/papis/papis/document.py", line 122, in to_bibtex
    document
  File "/home/teto/papis/papis/utils.py", line 94, in format_doc
    return python_format.format(**{doc: document})
IndexError: string index out of range
```

### DIFF
--- a/examples/scripts/papis-zotero-sql
+++ b/examples/scripts/papis-zotero-sql
@@ -96,7 +96,7 @@ def getCreators(connection, itemId):
   creatorCursor.execute(
     itemCreatorQuery.format(itemID=itemId)
   )
-  creators = {}
+  creators = {} # type: ignore
 
   for creatorRow in creatorCursor:
     creatorName = creatorRow[0]
@@ -217,9 +217,14 @@ parser = argparse.ArgumentParser(
 parser.add_argument('zotero_dir', type=str, default=".",
     help="Path to zotero's documents directory")
 
+parser.add_argument('papis_lib_dir', type=str, default="Documents",
+    help="Path towards papis' library directory")
+
+
 args, unknown = parser.parse_known_args()
 
 inputPath = args.zotero_dir
+outputPath = args.papis_lib_dir
 
 connection = sqlite3.connect(os.path.join(inputPath, "zotero.sqlite"))
 cursor = connection.cursor()
--- a/papis/commands/export.py
+++ b/papis/commands/export.py
@@ -176,7 +176,10 @@ def cli(
     """Export a document from a given library"""
 
     logger = logging.getLogger('cli:export')
-    documents = papis.database.get().query(query)
+    if all:
+        documents = papis.database.get().get_all_documents()
+    else:
+        documents = papis.database.get().query(query)
 
     if json and folder or yaml and folder:
         logger.warning("Only --folder flag will be considered")
--- a/papis/document.py
+++ b/papis/document.py
@@ -113,7 +113,17 @@ def to_bibtex(document):
     if not bibtexType:
         bibtexType = "article"
 
-    ref = document["ref"]
+    # REFERENCE BUILDING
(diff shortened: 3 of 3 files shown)


alejandrogallo:
Hi @teto,

I would propose the following pseudo code

```python
if document.has('ref'):
    ref = document['ref']
elif papis.config.get('ref-format'):
    # Construct your reference
elif document.has('doi'):
    ref = document['doi']
else:
    ref = # folder name of the document
```

What do you think?

teto:
I think it makes sense if the ref building is adjusted. I assume the yaml always gets generated with a ref but if it can't generate a ref, it should leave it empty instead of "noreference". In the short run, I would like to fix 
https://github.com/papis/papis/pull/150 first.

teto:
@alejandrogallo so I fixed the ref generation to match your logic (which I agree with). I wonder if that should go into its own function. 
I plan to write a test (and clean up the code). Is there a database of document or some specific test I could inspire myself from ?

As for the zotero-sql part, I wonder if it would be fine to pass the sqlite database instead of just the folder and assume that the database is named "zotero.sqlite". It would allow to let argparse check/open for the file and might be more robust in case some people renamed it ( like I did to zotero-backuup.sqlite). That would be a separate PR.

alejandrogallo:
Hi! Thank you very much for everything.
I merged it with the exception for commit abdf875
since the `--all` flag in export just means that you export all the selected documents with the query.
If you want to export everything with whoosh, the logic is
```
papis export --bibtex --all '*'
```

we can discuss the logic..

teto:
haaa you merged before I had time to cleanup xD ! I will add tags [WIP] / [READY] next time :p 

`papis export --bibtex --all '*'` works but `papis export --bibtex --all` won't. I think it doesn't use `default-query-string = '*'` as for open so that's confusing.

What about `As for the zotero-sql part, I wonder if it would be fine to pass the sqlite database instead of just the folder and assume that the database is named "zotero.sqlite".` ?

PR #147, opened 2019-03-13
nc flag: convert to --color=always|auto|no

teto:
more traditional. I am not sure which command outputs colors in order to test.


alejandrogallo:
Right now is just the logging outputing color, this would not work for the `tui`, but since the `tui` is experimental now, let's leave it like this, that's very good.

Thanks!

PR #150, opened 2019-03-14
zotero-sql: import mendeley keys when present

teto:
This looks for "Citation Key" in the "extra" key of zotero and generates a ref from it.
The ref is correctly found when I look at the output of the conversion script yet in the info.yaml, it's always zotero's key that appears.
I don't know if it's overwritten or not ?
I would love some tips in order to get my .tex compile without  touching every ref :)

### DIFF
--- a/examples/scripts/papis-zotero-sql
+++ b/examples/scripts/papis-zotero-sql
@@ -8,6 +8,7 @@ import os
 import shutil
 import glob
 import dateutil.parser
+import argparse
 
 # path to zotero SQLite database "zoter.sqlite" and "storage" to be imported
 inputPath = "."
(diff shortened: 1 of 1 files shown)


alejandrogallo:
Ok yes that's useful. Thanks a lot!

teto:
sorry if I was not clear but this is not fully functional. It does find the mendeley key and convert it to ref but somewhere it gets erased in the info.yaml in my test but I don't know why. Maybe the info.yaml gets rewritten later.
well I can still improve it later. The argparse parser is a nice addition IMO.

teto:
Nevermind it worked correctly but I was looking at wrong folder I think. I've added an outPath to the conversion script parser to make it clearer where it exports to. will submit later.

PR #156, opened 2019-03-22
[RDY] poc for export plugins

teto:
Just a Proof of concept to show how it would work, the plugin can be a class instance but here a function was enough.
I added a --format parameter to the export command since the exporters are now dynamically loaded.
The current syntax is kinda strange since you don't really know what to expect from `export --json --yaml --text`, better to enforce a single export format.
`papis export --format=bibtex --all`.

Whit this design, if you install a package that registers an entry_point for 'papis.exporter', it would appear in papis `export --format` options.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@ VERSION v0.9
 ============
 
 ## Add command ##
+- The configuration settings `export-text-format` has been removed along with
+  the export --text command. papis now support plugins so you should write your
+  own instead.
+- The flags --bibtex/--json/--yaml of the `export` command have been replaced
+  by `export --format=bibtex/json/yaml`
 
 - The configuration settings `file-name` and `folder-name` are now
   ```
--- a/doc/source/default-settings.rst
+++ b/doc/source/default-settings.rst
@@ -73,10 +73,6 @@ General settings
     The name of the library that is to be searched when ``papis``
     is run without library arguments.
 
-.. papis-config:: export-text-format
-
-    The default output papis format for ``papis export --text``.
-
 .. papis-config:: format-doc-name
 
     This setting controls the name of the document in the papis format strings
--- a/papis/commands/export.py
+++ b/papis/commands/export.py
@@ -55,57 +55,73 @@ import papis.api
 import papis.database
 import papis.strings
 import logging
+from stevedore import extension
 
+logger = logging.getLogger('cli:export')
+
+def stevedore_error_handler(manager, entrypoint, exception):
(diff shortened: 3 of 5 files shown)


teto:
I guess it would make more sense as a positional argument: `export json` / `export bibtex `

alejandrogallo:
Hi!

yes, I see it, I think it would be a very nice thing to have.
I guess it is not necessary to add as plugins the existing code, but
explain in the documentation how one would go ahead and create a plugin, right?

alejandrogallo:
About the `--format` flag, I agree, it makes more sense like that and it's more extendable.

teto:
I am thinking of removing --format and passing the format as a positional argument. Otherwise keep --format but set a default (bibtex ?). If there is no default than positional makes more sense.
Dunno if you want to keep the --bibtex / --json flags for backwards compatibility ? As the syntax is not great, I would be in favor of removing them.

The text export seems kinda brittle and would typically look like sthg that should be out of the core ? would allow to remove the `export-text-format` config.

alejandrogallo:
I agree on both things.
Let's do it with `--format` and `-f` flags, so for instance:
```
paps export -f json
```

This break of backwards compatiblity will be noted on the `changelog.md` file then
for version `v0.9`.

teto:
I've added an entry in the changelog and removed support for --text.
I've tested manually each format and it seemed to work. ready for merge.

alejandrogallo:
Why do we have to put all the plug-ins in the setup script. I thought the internal plug-ins we don’t have to put in the setup script, it is only for the external don’t you think? Otherwise we are going to have a very big set up. I don’t know what’s the standard here.

teto:
`Otherwise we are going to have a very big set up.` you mean setup.py ? 
I don't know if there is any standard here. I prefer it like that (that's how I use it in my project too) because it results in only one codepath => fewer bugs (I don't need to distinguish between native vs plugins). 
One perk too is it makes easier to disable some features though it's not true here but in my codebase, some plugins depend on external libraries and when there is a bug or for refactoring, I can just disable the feature by commenting it out in setup.py. Feel free to reject or adapt.

alejandrogallo:
Ok, fair enough, let's try it!

alejandrogallo:
So in principle now I can just declare a plugin for `papis-html` from the `papis-html` project and
use it in the normal papis as `papis export --format html `, let me tell you, that's brilliant!
You just made papis much better

teto:
Yes that's correct. In fact i believe this approach is superior to just dumping the plugins into.config/papis/scripts as it forces you to write a setup.py explicit about extra dependencies. The downside is that you have to write a setup.py xD

PR #159, opened 2019-03-25
Default command

teto:
Displays the help if you just type "papis". I find this behavior better though it's debatable.

Also references similar software as it's good practice. I didn't know about xapers, looks interesting though no release since 2017.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -125,3 +125,10 @@ See the ``AUTHORS`` list for a list of authored commits.
    :target: https://asciinema.org/a/3otatlbqXIsI102uAywMhT4fP
 .. |Code_Quality| image:: https://img.shields.io/lgtm/grade/python/g/papis/papis.svg?logo=lgtm&logoWidth=18
    :target: https://lgtm.com/projects/g/papis/papis/context:python
+
+Related software
+----------------
+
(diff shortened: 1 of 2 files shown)


alejandrogallo:
Hi @teto, I have no inconvenient, I only use the default to clear cache, but maybe we can also create a command to clear the cache, what do you think, maybe something like
```
papis db clear
```
?

alejandrogallo:
About xapers, I knew about it, specially xapian, however I think whoosh is not worse for our case than xapian, maybe for full text search is xapian better, but it is not pure python.

teto:
for full search text, xapian might be better true. 
I don't think the database command should be exposed. The --clear-cache command should be needed rare enough that it doesn't deserve a top level command.

alejandrogallo:
The only problem is that now `papis --clear-cache` does not work, because it expects a command...

teto:
oh I understand now why you were speaking of adding a db command xD
I still don't think the DB should be exposed. The --clear-cache seems like an "import" function to me. It's like importing from the current library folder. If we really want to go that way I would propose `papis import --from-lib=lib`. Otherwise just leave as is. The feature may not be worth breaking backwards compatibility.

alejandrogallo:
Let's not change it then !

PR #162, opened 2019-03-28
Importers as plugin

teto:
Very much WIP. I believe the import area is where plugins would shine the most as there are so many websites etc.
There is an attempt to guess what plugin to load from based on some heuristic (depending on if the string is a local file/folder/DOI/ url with "arxiv" in it etc).
But I have a conceptual problem as some of these are "mass" importers while others are specific to a file.
On current master if I do `add --from-doi <DOI> myfile myfile2` then myfile2 is ignored ?!
Same if I import a full library or a folder. I believe there are different semantics which make it harder to propose something coherent.


teto:
I am on holiday and as I never really used the add  command yet, it's kind of tricky to wrap my head around all the parameters so feel free to take over/start from scratch xD
There are some interesting improvements possible in this `import/add` area for sure.

alejandrogallo:
OK perfect I’ll take a look next week
Thank you very much

alejandrogallo:
I'm willing to work on this, but I think the best way to deal with it would be the following cli

```
papis add --from url 'google.com/' --from doi '1011.123123/123'
```

etc, so that always the `--from` flag needs two arguments, because always an importer will need an argument, right?

alejandrogallo:
I have already started, I have some ideas also to improve the downloaders with this, and then everything will be set for `v0.9`, if you have time and want to write something about plugins it would be great.
I thought also about doing an example repo of a papis plugin, but maybe I can do this with the `papis-octopart` script.

teto:
I don't think I will have any time soon. The thing is I am regularly confused by the papis UI, the use of positional vs optional arguments doesn't follow the usual pattern. The import command seems to have many syntaxes where sometimes it ignores extra arguments and in other cases processes all of them.  I would like to write a follow up on my previous feedback as I don't necessarily have the time to write code right now.

papis/importer.py
```diff
@@ -0,0 +1,108 @@
...
+        """
+        raise NotImplementedError()
+
+    def __str__(self):
+        return 'Importer({0}, uri={1})'.format(self.name, self.uri)
+
+
+def stevedore_error_handler(manager, entrypoint, exception):
+    logger.error("Error while loading entrypoint [%s]" % entrypoint)
+    logger.error(exception)
```
teto: I think we should move this to __init__ as it is used in almost every plugin-based code.

teto:
I'll try to test it, could you share some typical commands ? One of my concerns was how to proceed if several importers match the document (in smart mode for instance). Do we try all of them until one succeeds/ask the user/provides some ranking (so instead of returning true, we could return some kind of a confidence integer to rank the importer candidates).

papis/importer.py
alejandrogallo: Agreed, originally I wanted to do a `plugin.py` central package, however I messed up with stevedore, stevedore is sensitive about some things, also specially about initializing twice an extension manager, that's why everywhere we should write a function like _create_importer_mgr
that only creates the manager if has not been created previously.

alejandrogallo:
As of now, it should work like
```
papis add --from url https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.122.086802 --from doi 10.1103/PhysRevLett.122.086802
```
etc as before.

However even things like this
```
papis add https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.122.086802
```
work now, where it enteres into `smart` mode and it tries to match the importers.

I thought about doing this matching of the importers, fetch the information for
whichever importers match and then ask the user for input like, should we use
all files, merge all data, use only the data fetched from crossref, etc..

teto:
I am glad you do it since I would have been unsure about the UI. Question: couldn't we consider the zotero importer as an (imorter) plugin ? as in if the argument is an SQLite database, then import documents from it ?

teto:
so I tried on master `papis add --from-doi 10.1145/3241539.3241571 https://web.eecs.umich.edu/~hyunjong/pubs/mobicom18_raven.pdf` which failed with `Error: Invalid value for "[FILES]...": Path "https://web.eecs.umich.edu/~hyunjong/pubs/mobicom18_raven.pdf" does not exist.`.
I tried my chance with this PR but got 
```
  File "/run/user/1000/tmp.cdsRnawTlw/bin/papis", line 11, in <module>
    load_entry_point('papis', 'console_scripts', 'papis')()
  File "/nix/store/pv9x3fbb3nww5gd3shvhdj9scra7bc68-python3.7-bootstrapped-pip-19.0.3/lib/python3.7/site-packages/pkg_resources/__init__.py", line 489, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/nix/store/pv9x3fbb3nww5gd3shvhdj9scra7bc68-python3.7-bootstrapped-pip-19.0.3/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2793, in load_entry_point
    return ep.load()
  File "/nix/store/pv9x3fbb3nww5gd3shvhdj9scra7bc68-python3.7-bootstrapped-pip-19.0.3/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2411, in load
    return self.resolve()
  File "/nix/store/pv9x3fbb3nww5gd3shvhdj9scra7bc68-python3.7-bootstrapped-pip-19.0.3/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2417, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/home/teto/papis/papis/commands/default.py", line 32, in <module>
    import papis.api
  File "/home/teto/papis/papis/api.py", line 6, in <module>
    import papis.utils
  File "/home/teto/papis/papis/utils.py", line 12, in <module>
    import papis.crossref
  File "/home/teto/papis/papis/crossref.py", line 4, in <module>
    import doi
ModuleNotFoundError: No module named 'doi'
```

alejandrogallo:
Yes, I outsourced the `doi` package because I need it for other projects, so
`pip3 install python-doi` ;)

teto:
I had searched the papi repositories for "doi" but not your own. Will it stay there (asking because I am packaging papis for nixos so I would rather get the good urls straightaway)

alejandrogallo:
@teto, true, I think I'll put it as a papis project in the organization, anyways it will be for v0.9, which by the way I really intend to finish up this weekend.

teto:
Cool ! great work ! 

Some more feedback ignore if it's normal during WIP, it's blind tries without checking the new code:
I've packaged python-doi and retried `papis add --from-doi 10.1145/3241539.3241571 https://web.eecs.umich.edu/~hyunjong/pubs/mobicom18_raven.pdf`
```
WARNING:cli:add:
The flag --from-doi '10.1145/3241539.3241571' is deprecated, please use in the future
         --from doi '10.1145/3241539.3241571'
ERROR:cli:add:'doi'
  File "/home/teto/papis/papis/commands/add.py", line 634, in cli
    importer = import_mgr[importer_name].plugin(uri=resource)
  File "/nix/store/rdxcgir6sabkbly1zrm98kcn5v6lvfz2-python3.7-stevedore-1.30.1/lib/python3.7/site-packages/stevedore/extension.py", line 326, in __getitem__
    return self._extensions_by_name[name]
KeyError: 'doi'
  File "/run/user/1000/tmp.qV2zZkqU2u/bin/papis", line 11, in <module>
    load_entry_point('papis', 'console_scripts', 'papis')()
  File "/nix/store/8vb16vb9d27171gxxzzsljlpvsgpdmy6-python3.7-click-7.0/lib/python3.7/site-packages/click/core.py", line 764, in __call__
  File "/nix/store/8vb16vb9d27171gxxzzsljlpvsgpdmy6-python3.7-click-7.0/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/nix/store/8vb16vb9d27171gxxzzsljlpvsgpdmy6-python3.7-click-7.0/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
  File "/nix/store/8vb16vb9d27171gxxzzsljlpvsgpdmy6-python3.7-click-7.0/lib/python3.7/site-packages/click/core.py", line 956, in invoke
  File "/nix/store/8vb16vb9d27171gxxzzsljlpvsgpdmy6-python3.7-click-7.0/lib/python3.7/site-packages/click/core.py", line 555, in invoke
  File "/home/teto/papis/papis/commands/add.py", line 671, in cli
    link=link
  File "/home/teto/papis/papis/commands/add.py", line 274, in run
    raise IOError('Document %s not found' % p)
OSError: Document https://web.eecs.umich.edu/~hyunjong/pubs/mobicom18_raven.pdf not found

```
I retried with `papis add --from doi 10.1145/3241539.3241571 https://web.eecs.umich.edu/~hyunjong/pubs/mobicom18_raven.pdf`
which ended up with
```
Usage: papis add [OPTIONS] [FILES]...

Error: Invalid value for "--from": invalid choice: doi. (choose from )


```

alejandrogallo:
Stevedore doesn't update automatically when you pull, try installing papis again, like
```
python3 setup.py develop
```
etc. You're not getting the entry points.

I'm preparing now a difflib based tool to diff dicts, I think it'll be ok then
    https://asciinema.org/a/SssvDApZuMQiJfZ6EgGBzkPKJ

alejandrogallo:
I'm really testing a lot many concepts, the difftool is working now but I'm not sure
is the best user experience,

I'm experimenting now with not making a dfiference between importers and downloaders, which seems to me to make the most sense.

I send you an example:

https://asciinema.org/a/T8ylXdX5vB1ZRic8A95tqDXBL

Right now the idea would be, that if you want to do something in the `smart` way,
you just type in a resource, and the matcher for the importers
matches the resource against their function, if they match (for instance in the example the fallback downloader and the hal downloader match), so they both fetch information and at the end you decide with the diff tool, which now supports splitting.

What do you thing about this?

Maybe this also interests you @gaalcaras

teto:
so in case you have several matches, they all fetch information (in parallel ?) and the user decides between the different results ? or does it choose which one to run ?

I was not familiar with the old UI which is why I refrained from pursuing a new UI but seems like you are open to change it completely.

alejandrogallo:
They don't fetch the information in parallel yet, it could be done easily but it's not my priority.
The user now will gradually decide between different results, maybe you merge
the abstract of the fallback downloader and the rest from the doi downloader.
Etc.

Yes I'm willing to change it because it was a bad UI. Now it's better actually, I'm then generalitzing 
the update command as well to work in the auto mode like this.

alejandrogallo:
Hi @teto , 

this branch turned into branch `v0.9`, which you can already fiddle around, I'm trying to solve the windows bug and refine the add and update diff based tui interface.

Cheers!

PR #163, opened 2019-03-28
Fix get_scripts()

zlangley:


zlangley:
This was failing for me locally when trying to run `make test`.

alejandrogallo:
Hi @zlangley ,
sorry for the delay, interesting, I don't know reall which method to take, from the stevedore source code
```python
    def names(self):
        "Returns the names of the discovered extensions"
        # We want to return the names of the extensions in the order
        # they would be used by map(), since some subclasses change
        # that order.
        return [e.name for e in self.extensions]
```

and 

```python
    def entry_points_names(self):
        """Return the list of entry points names for this namespace."""
        return list(map(operator.attrgetter("name"), self.list_entry_points()))
```

I think taking the `names` method would make more sense, as you suggest, I'd like to hear @teto 's 
opinion on the matter, if possible.

Thanks a lot.

teto:
plugins are defined as A=X, I think names refer to A while entry_point refer to X thus A would make more sense,X being an implementation detail.

zlangley:
For more context, this is just currently broken. In the loop, the code `commands_mgr[command_name]` results in a `KeyError` if you iterate over `entry_points_names`.

alejandrogallo:
I'll change also in the other managers the `entry_point_names` method for the `names`.

Issue #167, opened 2019-04-09
python_format.format(**{doc: document}) fails with missing keys


teto:
We should better handle errors when a key is not available:
for instance I have `ref-format = {doc[author_list][0][surname]}{doc[year]}` which triggered:

```
INFO:cli:update:Try using doi 10.1109/ICC.2018.8422950
Traceback (most recent call last):
  File "/run/user/1000/tmp.arEYFH6UA4/bin/papis", line 11, in <module>
    load_entry_point('papis', 'console_scripts', 'papis')()
  File "/nix/store/8vb16vb9d27171gxxzzsljlpvsgpdmy6-python3.7-click-7.0/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/nix/store/8vb16vb9d27171gxxzzsljlpvsgpdmy6-python3.7-click-7.0/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/nix/store/8vb16vb9d27171gxxzzsljlpvsgpdmy6-python3.7-click-7.0/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/nix/store/8vb16vb9d27171gxxzzsljlpvsgpdmy6-python3.7-click-7.0/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/nix/store/8vb16vb9d27171gxxzzsljlpvsgpdmy6-python3.7-click-7.0/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/teto/papis/papis/commands/update.py", line 367, in cli
    doidata = papis.crossref.doi_to_data(query)
  File "/home/teto/papis/papis/crossref.py", line 188, in doi_to_data
    results = get_data(dois=[doi])
  File "/home/teto/papis/papis/crossref.py", line 172, in get_data
    for d in docs
  File "/home/teto/papis/papis/crossref.py", line 172, in <listcomp>
    for d in docs
  File "/home/teto/papis/papis/crossref.py", line 127, in crossref_data_to_papis_data
    papis.config.get("ref-format"), new_data
  File "/home/teto/papis/papis/utils.py", line 94, in format_doc
    return python_format.format(**{doc: document})
KeyError: 'surname'
```
[...]

PR #168, opened 2019-04-10
User format throws if missing key

teto:
Catch it and provide a default that names the missing key instead.

Not sure that's a fantastic fix but at least it helped me go through the issue in https://github.com/papis/papis/issues/167

### DIFF
--- a/doc/source/default-settings.rst
+++ b/doc/source/default-settings.rst
@@ -92,27 +92,6 @@ General settings
 
     Default format that is used to show a document in order to select it.
 
-.. papis-config:: format-jinja2-enable
-
-    This setting is to enable the `jinja2 <http://jinja.pocoo.org//>`_ template
-    engine to render the papis templates being used, as ``header-format``,
-    ``match-format`` etc...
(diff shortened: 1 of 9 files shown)


alejandrogallo:
I don't really understand what's happening, maybe you could write a couple of tests about this in
```
tests/test_utils.py
```
at the function
```python
def test_format_doc():
```
and run it with
```
pytest tests/test_utils.py
```
?

teto:
I explained it in the issue: ref-format  is configured by the user in `$XDG_CONFIG_HOME/papis/config` so the input can't be trusted: it can also fail when the document doesn't contain a used key , in my case, `doc[author_list][0][surname]}`. We need to warn the user.
The stacktrace in https://github.com/papis/papis/issues/167 shows the function was invoked by papis.crossref.doi_to_data . I am not a fan of unit testing (except for complex functions) as papis codebase is moving fast, I rather prefer integration testing (makes also tests easier to write IMO).

I will add a commit for the jinja2 removal.

PR #183, opened 2019-06-11
tandfonline improvements

alexfikl:
Simplified and improved the author_list scraping in `fetch_data` a bit:
* removed some superfluous spaces around names
* get affiliation without some other noise

It seems to work a bit better for the implemented test cases and a couple of additional ones. The rest of it just updates the test resources.


tests/downloaders/resources/tandfonline_1_out.json
```diff
@@ -1,25 +1,14 @@
...
-  "number": "9-11",
-  "publisher": "Taylor & Francis",
-  "pages": "1092-1099",
+  "abstract": "Contrarily to what happens with the Epstein–Nesbet (EN) zeroth-order Hamiltonian, the Møller–Plesset (MP) perturbation operator has diagonal matrix elements, the expression of which is recalled. It is a balance between hole–hole and particle–particle repulsions on one hand and of hole–particle attractions on the other hand. For the double excitations, which dominate the correlation effects, the attractive terms prevail and the second-order MP energy is underestimated, at least for atoms of the first rows of the periodic table. It will be shown that when the perturbation expansion reaches multiple excitations, the diagonal terms of the MP perturbation operator may become larger than the zeroth-order MP excitation energy and creates an oscillating divergence of the series. Several situations of this type will be presented. This divergence is linked to the non-additivity of excitation energies, while this additivity is an underlying assumption for the linked cluster theorem and the coupled cluster method. Th...",
   "author": " Jean-Paul   Malrieu  and  Celestino   Angeli ",
```
alexfikl: This is still a bit annoying. In `fetch_data`, it gets a nicely formatted author string, but then that gets overwritten by the bibtex data [here](https://github.com/papis/papis/blob/v0.9/papis/downloaders/base.py#L168)

papis/downloaders/tandfonline.py
```diff
@@ -22,41 +25,41 @@ def get_data(self):
...
+
+                affiliation = self.re_comma.sub(', ',
+                        self.re_clean.sub('', affiliation))
+                affiliation = [dict(name=affiliation)]
+
+            # find href="/author/escaped_fullname"
+            fullname = author.find_all('a', attrs={'class': 'entryAuthor'})
+            fullname = fullname[0].get('href').split('/')[-1]
+
+            fullname = re.sub(r'%2C', ',', re.sub(r'\+', ' ', fullname))
```
alexfikl: I'm not that familiar with this sort of stuff, but is there a more generic way to replace these HTML codes (like `%2C`) with their unicode equivalent?

tests/downloaders/resources/tandfonline_1_out.json
alejandrogallo: I agree, we have to figure out a way of doing it in a controlled way, but I'm not sure yet what's best,
I think the bibtex information if the rest is good enough could be ignored.

alejandrogallo:
That's some great work, thank you very much @alexfikl , tell me whenever you're ready to merge,
I really appreciate you updating the tests, that way everyone can sleep well at night ;)

alexfikl:
The tests seem to be failing on the CI, but they're passing locally. Is that a problem?

Besides that, I think it should be good to go from my side!

PR #184, opened 2019-06-13
ACS improvements

alexfikl:
Tried to make this a bit more robust in finding the correct affiliations for each author.

I found a couple of cases where it still doesn't work that well, like

- [this](https://pubs.acs.org/doi/10.1021/jp035055c), because the affiliations aren't assigned
- [this](https://pubs.acs.org/doi/10.1021/acs.jpcc.5b08054), because they're all in a single line
- and probably quite a few more


alexfikl:
This should now handle the second case too, although the code is a bit nasty :(

teto:
is it ready for review ? could you share a command to check how it improved papis ?

alexfikl:
It should be ready for review, but unfortunately the improvements don't actually make it out. The problem is that in [fetch_data](https://github.com/papis/papis/blob/v0.9/papis/downloaders/base.py#L148) it first gets the web scrapped data (worked on in this PR) and then just overwrites it with the bibtex data a few lines below. This sort of worked when I started this PR because the bibtex data didn't have an author list, but then I added it in #193, so now it gets overwritten..

The way I checked that it's working as expected was with the tests in `test_acs.py`, which block out the bibtex data retrieval and just look at the `author_list`. You can look at the `acs_X_out.json` files to see the improvements. Basically
* makes use of `parse_meta_headers` to remove a bunch of code.
* a few cases got affiliations scrapped better, e.g. [this](https://pubs.acs.org/doi/10.1021/acsphotonics.9b00250) in `acs_3_out.json` and similar pages with popups weren't working at all before.
* parsing author names assumes a single family name and multiple given names, which should be the more common case (?), instead of the other way around.
* some more cleaning of the returned text.

With the issue in `fetch_data`, there isn't much actual benefit in merging this at this point, sort of like #183. Are there any plans for addressing that somehow?

PR #186, opened 2019-06-27
Add lxml dependency

alexfikl:
This is required [here](https://github.com/papis/papis/blob/v0.9/papis/downloaders/base.py#L210) and not installed on the CI, so it caused a few failures for the downloader tests.

I didn't add a minimum version because I'm not sure what a good one would be.


alejandrogallo:
Agreed, thanks!

PR #192, opened 2019-07-11
Ensure `files` key is present in document

avonmoll:
Sometimes when adding a document with no files to a library and then
later trying to use `addto` there would be an error because the document
(i.e. the YAML file) had no `files: []` entry. This commit ensures that
in the event that `files` key is not present that the key is added and
instantiated with an empty list.

### DIFF
--- a/papis/commands/addto.py
+++ b/papis/commands/addto.py
@@ -78,6 +78,8 @@ def run(document, filepaths):
         )
         shutil.copy(in_file_path, endDocumentPath)
 
+    if not "files" in document.keys():
+        document["files"] = []
     document['files'] += new_file_list
     document.save()
     papis.database.get().update(document)


papis/commands/addto.py
```diff
@@ -78,6 +78,8 @@ def run(document, filepaths):
         )
         shutil.copy(in_file_path, endDocumentPath)
 
+    if not "files" in document.keys():
+        document["files"] = []
```
teto:
```suggestion
        document["files") = document.getdefault("files", []) + new_file_list
```

papis/commands/addto.py
```diff
@@ -78,6 +78,8 @@ def run(document, filepaths):
...
     document['files'] += new_file_list
```
teto: I would rather see `document["files"] = document.getdefault("files", []) + new_file_list`

avonmoll: I'm not quite sure what you mean. I see that the `Document` object has a method called `get_files` (but I'm not seeing anything like `getdefault`). The issue with `get_files` is that it returns the `files` list with full file paths. I'm using a papis library across macOS and linux and thus only want to store relative file paths in `files`.

After re-looking at `document.py` I'm questioning why this is even necessary.  The `__setitem__` method adds the key if its not present. Will try to reproduce and re-debug.

teto:
Just had the problem which reminded me of this sry. Thanks for the contribution

PR #193, opened 2019-07-14
Get author_list from bibtex data

alexfikl:
Makes use of `keyconversion_to_data` in `bibtexparser_entry_to_papis`.


papis/bibtex.py
```diff
@@ -140,10 +158,10 @@ def bibtex_to_dict(bibtex):
     """
     from bibtexparser.bparser import BibTexParser
 
-    parser = BibTexParser(common_strings=True)
-    parser.ignore_nonstandard_types = False
-    parser.homogenise_fields = False
-    parser.interpolate_strings = True
+    parser = BibTexParser(common_strings=True,
+            ignore_nonstandard_types=False,
+            homogenize_fields=False,
```
alexfikl: Unrelated change, but there was a typo: `homogenize` instead of `homogenise` (see [docs](https://bibtexparser.readthedocs.io/en/master/bibtexparser.html#module-bibtexparser.bparser)).

teto:
might be good to have a test. Would be nice to have examples of what the commits do in the commit messages.

alexfikl:
> might be good to have a test. 

@teto Added a simple test case. Does that look ok?

> Would be nice to have examples of what the commits do in the commit messages.
 
I'd be happy to change them, but I'm not quite sure what you mean. Can you give me an example?

tests/resources/bibtex/1_out.json
```diff
@@ -0,0 +1,67 @@
+{
+  "abstract": "... to 100 {\\%}(concurrent intercalation)...",
+  "author": "Shi, % some comments here\nFazhan and Rong, Xing and Xu, Nanyang and Wang, Ya and Wu, Jie and Chong, Bo and Peng, Xinhua and Kniepert, Juliane and Schoenfeld, Rolf-Simon and Harneit, Wolfgang and Feng, Mang and Du, Jiangfeng",
```
alexfikl: This seems to be a problem with bibtexparser. It looks like it's not great at parsing comments (see [this issue](https://github.com/sciunto-org/python-bibtexparser/issues/133))

teto:
oh great thanks. I am not a maintainer of papis so don't take my comments too seriously.
For instance, for the commit: `support multiple given names`, I would be curious to see (in the commit message) an example of an author list that didn't work before but works now with your commit.

alexfikl:
> oh great thanks. I am not a maintainer of papis so don't take my comments too seriously.

More reviews are always welcome, so thanks!

> For instance, for the commit: support multiple given names, I would be curious to see (in the commit message) an example of an author list that didn't work before but works now with your commit.

Ah, I see. That was just an improvement over my previous commit, where it was doing
```
given, family = re.split(r'\s+', author)
``` 
which couldn't unpack if there were several whitespace-separated names. I'll see if I can expand the commit message a bit.

PR #195, opened 2019-07-23
Fix git flags in documentation

olantwin:
Update documentation to use the correct git flag: `-a` doesn't exist. `-A` does (and was probably meant), but its long version `--all` is IMHO more understandable.

### DIFF
--- a/doc/source/git.rst
+++ b/doc/source/git.rst
@@ -101,7 +101,7 @@ When editing a document's info file:
 
     ::
 
-      papis git add -a
+      papis git add --all
 
   - Commit
 


teto:
hum I think the -a is supposed to be passed to git and the manpage for git commit does mention -a. Seems like it's valid.

olantwin:
Could this depend on the `git` version? With `git` 2.21.0, the -a flag does not exist for `git-add`.

teto:
was looking at the wrong manpage (git commit instead of git add). Sry LGTM

PR #198, opened 2019-07-31
Non-existent library folder causing crash

j-c-w:
Hi,

While setting up papis, I encountered some crashes when the library folder I specified in the config file didn't yet exist.  This addresses that by creating the folder if it doesn't exist rather than crashing.

Steps to reproduce crash:
In the config, create a library where the `dir` field doesn't exist.  Then, run papis -l library open "".

Changes pass the tests with no additional failures.

Jackson

### DIFF
--- a/papis/config.py
+++ b/papis/config.py
@@ -484,6 +484,10 @@ def get_lib_from_name(libname):
             raise Exception('Library {0} not defined'.format(libname))
         try:
             paths = [expanduser(config[name]['dir'])]
+            for path in paths:
+                if not os.path.exists(path):
+                    raise Exception(("Path '%s' for library '%s' does not "
+                                    "seem to exist") % (path, name))
         except KeyError:
             try:
                 paths = eval(expanduser(config[name].get('dirs')))
--- a/papis/library.py
+++ b/papis/library.py
@@ -1,6 +1,7 @@
 import os
 import glob
 import logging
+import sys
 
 logger = logging.getLogger("library")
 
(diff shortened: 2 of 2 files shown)


teto:
I don't think we should create the folder. It is errorprone. Instead, papis should warn the user and not crash when the folder doesn't exist.

j-c-w:
Hi,

Yeah, seems reasonable.  I've updated this to crash with a helpful error when using a directory that doesn't exist.  Mirrors the behaviour of papis -l <non-existent file path>


Jackson

alejandrogallo:
Hi, can you squash all the commits into one? I'm reviewing all the prs and it'd help.

papis/library.py
```diff
@@ -8,6 +8,7 @@ def __init__(self, name, paths):
         assert(isinstance(name, str)), '`name` must be a string'
         assert(isinstance(paths, list)), '`paths` must be a list'
         self.name = name
+
```
teto:
```suggestion
```

papis/config.py
```diff
@@ -484,6 +484,10 @@ def get_lib_from_name(libname):
             raise Exception('Library {0} not defined'.format(libname))
         try:
             paths = [expanduser(config[name]['dir'])]
+            for path in paths:
+                if not os.path.exists(path):
+                    raise Exception(("Path '%s' for library '%s' does not "
+                                    "seem to exist") % (path, name))
```
teto: woulnd't exist work for files as well as directories ? Shouldn't it be moved to the Library constructor so that it becomes more generic ?

j-c-w:
I think this has been addressed in 081ff55c33c1eb32c958476cc4d0e71a7652df77

PR #199, opened 2019-07-31
whoosh: autoupdate on schema change

j-c-w:
Hi,

I noticed that papis crashed if whoosh-schema-fields was changed after initializing a library.  I've attempted to address this issue by checking if the user fields are the same as the database fields.

If they are not, the database is cleared and recreated.

Steps to reproduce the crash:

In the config file, set:
whoosh-schema-fields = ['<something not currentlly in whoosh-schema-fields']
Then, use papis add <pdf> --from-doi <doi>

This passes with no additional test failures.

Is this OK?

Jackson

### DIFF
--- a/papis/database/whoosh.py
+++ b/papis/database/whoosh.py
@@ -218,10 +218,44 @@ class Database(papis.database.base.Database):
         """Function to be called everytime a database object is created.
         It checks if an index exists, if not, it creates one and
         indexes the library.
+
+        If the schema fields have been changed, it updates the database.
         """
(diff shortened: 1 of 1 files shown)


teto:
haven't tested but Looks good. thanks

teto:
could you change the commit message to:
`whoosh: autoupdate on schema change`
or something that respects https://chris.beams.io/posts/git-commit/ .
I will merge afterwards.

teto:
thanks

PR #200, opened 2019-08-01
Different opentools for different file types

j-c-w:
Hi,

This patch uses the file extension on the files opened to select a different opener program.

An example usecase is:
I download a .pdf paper, and want to use evince to open it.
I download an RFC (which is a .txt file), and want to use a text editor to open it.

So, in the config file, I can write:

opentool-pdf = evince
opentool-txt = gvim

And the appropriate program will be used.  This doesn't destroy any old behaviour: if a filetype without a specified opentool is opened, it will fallback to the old behaviour of using the `opentool' program.

Documentation has been updated to reflect this.

There are no additional test failures.

Is this OK?

Jackson

### DIFF
--- a/doc/source/configuration.rst
+++ b/doc/source/configuration.rst
@@ -37,7 +37,9 @@ A more complete example of a configuration file is the following
   # "cygstart"
   #
   opentool = rifle
-  # Specify a PDF editing program for PDF filetypes
+  # Specify a PDF editing program for PDF filetypes.
+  # More generally, a tool for a filetype .XYZ can be specified as 
+  # "opentool-XYZ = <tool>"
   opentool-pdf = evince
   # Use ranger as a file browser, a nice python program
   file-browser = ranger
--- a/doc/source/configuration.rst
+++ b/doc/source/configuration.rst
@@ -37,6 +37,8 @@ A more complete example of a configuration file is the following
   # "cygstart"
   #
   opentool = rifle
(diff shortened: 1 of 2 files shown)


teto:
Independently of the PR, I am not sure it's a good idea to choose the application in papis, systems have their own way of opening file. On linux, I know xdg-open or rifle from ranger's.

j-c-w:
Hi teto,

So for me, this is motivated because I don't want papis to use the same program as everything else by default.  So, for a "normal" pdf, I would like it to open in chromium, because I usually already have it open,  it has a relatively smooth PDF viewer, and it is easy for me to do things I am subsequently likely to do (e.g. print).   When I'm opening a PDF in papis, I'm likely to want to add notes, so I'd rather use qpdfview.

Also, when papis is replicated between machines, I don't need to worry about keeping machine-wide defaults in sync to see consistent behaviour: all I need to do is make sure they have the same papis config file and I'll see the same behaviour.

That's why I found papis to be the right place to do this.

teto:
Do you use different OSes on different machines ? if you are on linux you could wrap the papis binary so that xdg-open use a different mime configuration.

j-c-w:
I use the same OS.  That said, your approach sounds more difficult than using this patch to me.

However, I do understand that papis might not be the right place to do this, so I'm willing to let this die and just keep it in my local branch.

Thanks for the comments,

Jackson

PR #201, opened 2019-08-01
Add rm --all option

j-c-w:
Hi,

I noticed many of the commands have an --all option to repeat the command for all matches.  This adds an identical option to the rm command.

There are no additional test failures.

Any thoughts?

Jackson

### DIFF
--- a/papis/commands/rm.py
+++ b/papis/commands/rm.py
@@ -48,10 +48,17 @@ def run(document, filepath=None):
     is_flag=True,
     default=False
 )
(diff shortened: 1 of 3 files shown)


teto:
thanks

alejandrogallo:
It looks fine, since we've got the `--force` flag going on, I think it should be safe enough.
I'm worried always that I read somewhere `remove all`, however in this case it should be ok.i
Thank you!

PR #202, opened 2019-08-03
Print the library name when `papis list --libraries` is called

j-c-w:
Hi,

This changes the behaviour of the `papis list --libraries` command to print the name of the library in addition to its location.

I found it useful because I couldn't remember the exact names of some of the libraries I have.  It was helpful to be able to print them out with a single command.

Is this OK to go in?

Jackson


teto:
I agree. Thanks

PR #208, opened 2019-08-26
Print paper name and author list when fetching from DOI

j-c-w:
Hi,

A typical workflow for me is to do something like:

    papis add --from-doi <something> Downloads/bland_filename.pdf

Followed by an attempt to run:

    papis open "some search phrase"

I have found myself forgetting a unique search phrase I can use to subsequently open the document.

To address this, this patch prints the title and author list (provided they exist in the DOI information).  (So, I can immediately do `papis open "authors"`)

This patch also helps with cases where the DOI might have been misread or miscopied.

Is this OK?

Jackson

### DIFF
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -670,7 +670,12 @@ def cli(
         logger.info("using doi {0}".format(from_doi))
         doidata = papis.crossref.get_data(dois=[from_doi])
         if doidata:
-            data.update(doidata[0])
+            doidata = doidata[0]
+            data.update(doidata)
+            if 'title' in doidata and 'author' in doidata:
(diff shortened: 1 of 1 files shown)


teto:
I am not too fan of printing stuff. It's not clean to mess with stdout. I believe the issue at hand is linked to a more fondamental issue of UI in papis.
A quickfix could be to plug some hooks: for instance postAddHook where you can run the command of your choice.
A better long term fix would be to rework the UI. I would like to write something with fzf but I lack the time.

alejandrogallo:
I tend to agree with @teto, I think you can achieve some workarounds scripting your way out of the problem. Thank you so much however!

PR #213, opened 2019-09-06
jinja2: remove support

teto:
It makes little sense to have a separate path specifically for jinja2.
Same behavior should be achievable via plugins.

### DIFF
--- a/doc/source/default-settings.rst
+++ b/doc/source/default-settings.rst
@@ -92,27 +92,6 @@ General settings
 
     Default format that is used to show a document in order to select it.
 
-.. papis-config:: format-jinja2-enable
-
-    This setting is to enable the `jinja2 <http://jinja.pocoo.org//>`_ template
-    engine to render the papis templates being used, as ``header-format``,
-    ``match-format`` etc...
-
-    For instance you could set the option ``header-format`` to
-
-    .. code:: html
-
-        <span color='#ff00ff'>{{doc.html_escape["title"]}}</span>
-        <span color='#00ff00'>  {{doc.html_escape["author"]}}</span>
-        <span color='#00ffaa'>   ({{doc.html_escape["year"]}}) </span>
-        {%- if doc.has('tags') %}<span>[<yellow>{{doc['tags']}}</yellow>] </span>{%- endif %}
-        {%- if doc.has('citations') %}<red>{{doc['citations']|length}}</red>{%- endif %}
-        {%- if doc.has('url') %}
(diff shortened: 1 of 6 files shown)


alejandrogallo:
I agree, it looks fine, let us remove it for version 0.9, only @avonmoll is really using it to my knowledge, so we can develop other plugins for coming versions as you point out. This pr is a go for me.

teto:
@avonmoll do you have a backup plan if I merge this or should I wait ?

avonmoll:
I would say go ahead and merge. I'm currently using a version of 0.9 from a month or so ago with all of my PRs merged in. I would like to better understand how to achieve finer control over the templates (filenames, ref formats, etc.). In my naive mind, there would be some way to hook in a plugin or even just a single script that gets called when papis needs to generate a filename or refformat. Right now, we can only specify a textual template in the config. Without jinja2, there is much less control over how these are generated.

My use case is using papis as a master repository, tagging various documents (using the `project` list field) for a particular paper, then exporting all documents with `project:newPaper` to bibtex. When writing LaTeX documents I'd like to avoid the scenario where the cite keys are ambiguous or uninformative (e.g. `johnson2015the` or `johnson2015a`). A lot of times, I cite multiple works from the same author in the same year. It gets confusing sometimes (even now, with what's possible in jinja2). I don't know if this helps to explain why I want more control over things like ref format.

teto:
We should have a flag to either prevent duplicate keys (by adding a counter for instance) or warn the user.
As for more control over the format, the current system looks powerful enough, apart from the duplicate problem. What I do is that I add the reference manually in info.yaml, then they are not generated. Have you tried that ?
Can't you just manually ed

avonmoll:
Manually editing the `refformat` field is always an option. It's much more annoying for things like file and folder names though. My concern is less with duplicate keys (although that should be avoided somehow) and more with `refformat`s with stop-words like "a" or "the". This has nothing to do with jinja2 (as it's not possible there either). This is what prompted my original interest in being able to call a script to generate names. I don't think it should be a high priority since the current naming template configurations are probably sufficient for the vast majority of users.

teto:
I think editing the info.yaml with `ref: BertotYves2015` is a good solution.
As for ref-format and your ambiguous example `johnson2015the` vs `johnson2015a`, setting ref-format to use the title should fix the ambiguity ?

alejandrogallo:
Can I merge this?

teto:
yes go ahead please.

Issue #234, opened 2019-11-06
Port code to typed python from 3.5 onwards.


alejandrogallo:

alejandrogallo:
# Files to be edited

- [ ] papis/utils.py
- [ ] papis/crossref.py
- [ ] papis/library.py
- [ ] papis/strings.py
- [ ] papis/json.py
- [ ] papis/api.py
- [ ] papis/git.py
- [ ] papis/database/base.py
- [ ] papis/database/whoosh.py
- [ ] papis/database/cache.py
- [ ] papis/database/__init__.py
- [ ] papis/downloaders/fallback.py
- [ ] papis/downloaders/thesesfr.py
- [ ] papis/downloaders/get.py
- [ ] papis/downloaders/aps.py
- [ ] papis/downloaders/hal.py
- [ ] papis/downloaders/worldscientific.py
- [ ] papis/downloaders/springer.py
- [ ] papis/downloaders/tandfonline.py
- [ ] papis/downloaders/base.py
- [ ] papis/downloaders/scitationaip.py
- [ ] papis/downloaders/sciencedirect.py
- [ ] papis/downloaders/iopscience.py
- [ ] papis/downloaders/acs.py
- [ ] papis/downloaders/frontiersin.py
- [ ] papis/downloaders/annualreviews.py
- [ ] papis/downloaders/__init__.py
- [ ] papis/downloaders/ieee.py
- [ ] papis/bibtex.py
- [ ] papis/config.py
- [ ] papis/pubmed.py
- [ ] papis/document.py
- [ ] papis/exceptions.py
- [ ] papis/yaml.py
- [ ] papis/docmatcher.py
- [ ] papis/base.py
- [ ] papis/pick.py
- [ ] papis/dissemin.py
- [ ] papis/isbnplus.py
- [ ] papis/importer.py
- [ ] papis/isbn.py
- [ ] papis/tui/widgets/diff.py
- [ ] papis/tui/widgets/command_line_prompt.py
- [ ] papis/tui/widgets/__init__.py
- [ ] papis/tui/widgets/list.py
- [ ] papis/tui/app.py
- [ ] papis/tui/__init__.py
- [ ] papis/cli.py
- [ ] papis/arxiv.py
- [ ] papis/__init__.py
- [ ] papis/plugin.py
- [ ] papis/commands/addto.py
- [ ] papis/commands/external.py
- [ ] papis/commands/update.py
- [ ] papis/commands/browse.py
- [ ] papis/commands/git.py
[...]

PR #237, opened 2019-11-19
Add type information to api.py

alexfikl:
Hopefully helping out a bit with #234 and learning about typing in python.

What's a good way to check that the additions are correct? For this, I just ran `mypy` over the file and checked that there are no errors specifically in it.


alejandrogallo:
Hi! Thank you very much. About the way to check, this is pretty much also what I do, so it looks fine. I'll merge it.

PR #238, opened 2019-11-22
Add typing to arxiv.py

alexfikl:
Another addition to #234. Checked with
```
mypy --disallow-untyped-defs --ignore-missing-imports --show-error-codes papis/arxiv.py
```

### DIFF
--- a/papis/arxiv.py
+++ b/papis/arxiv.py
@@ -25,24 +25,26 @@ import click
 import papis.downloaders.base
 import arxiv2bib
 import os
+import sys
+from typing import Optional, Sequence, Dict, Any
 
 
 logger = logging.getLogger('arxiv')
 
 
 def get_data(
-    query="",
-    author="",
-    title="",
-    abstract="",
-    comment="",
-    journal="",
-    report_number="",
-    category="",
-    id_list="",
-    page=0,
-    max_results=30
-):
+        query: str = "",
+        author: str = "",
+        title: str = "",
+        abstract: str = "",
+        comment: str = "",
+        journal: str = "",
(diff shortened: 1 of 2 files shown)


papis/importer.py
```diff
@@ -66,7 +66,7 @@ def match_data(cls, data: Dict[str, Any]) -> None:
         raise NotImplementedError(
             "Matching data not implemented for this importer")
 
-    def fetch(self) -> str:
+    def fetch(self) -> Any:
```
alexfikl: Not quite sure about this change. I changed it to `Any` because the docs say it could return a dict, but the method in `ArxividFromPdfImporter` doesn't return anything.

papis/arxiv.py
```diff
@@ -73,7 +75,7 @@ def get_data(
     url = urllib.request.Request(
         req_url,
         headers={
-            'User-Agent': papis.config.get('user-agent')
+            'User-Agent': papis.config.get('user-agent')  # type: ignore[dict-item]
```
alexfikl: Not sure ignoring the type warning is the best idea here. The issue seems to be that `get` returns an `Optional[Any]` but `Request` expects a string.

alejandrogallo:
Hi Alex,

I did not have time to really go through your changes thoroughly since I was changing some stuff
and I had the objective of finishing with the typing as soon as my time was still available.

I already did arxiv and changed some things, maybe you'd like to take a look at it.

Thanks so much.

alexfikl:
No problem! Glad you found some time to work on papis!

I'll take a look when I find some time. Thanks!

PR #243, opened 2019-12-02
Remove upper limits on dependencies

### LINKED ISSUE #242: Upgrade dependencies?
I recently upgraded papis for Arch Linux to v0.9; however, this has caused a number of issues as Arch Linux packages are usually up to date, and for some reason papis puts upper limits on a number of dependencies.  In particular:

- [ ] `python-slugify<4,>=1.2.6` -- Is there a known incompatibility with version 4?  Looking at the [`python-slugify` changelog](https://github.com/un33k/python-slugify/blob/master/CHANGELOG.md), the only change between v3 and v4 is that they've dropped some older Python versions.
- [ ] `lxml<4.3.5` -- Again, the main breaking change for v4.4 I can see is a drop in support for older Python versions, and one deprecation.
- [ ] `prompt_toolkit<3` -- This seems to be a bigger change, but perhaps work towards `v3` should be considered?  (On Arch Linux, the latest version is still `2.0.10`, so I'm guessing the change to v3 is not completely trivial).
- [ ] `isbnlib<4` -- This is not an issue for now as `v4` hasn't been released; but is there a reason to preemptively be incompatible with the next version?

It appears to me, from the limited testing I have done, that removing the upper restrictions on the versions does not introduce any regressions (but this is from very limited testing).

JP-Ellis:
Upper limits can be re-instated if/when an incompatibility is discovered (and
until papis can be updated to be compatible with the newer version).

For python 3.4, conditional dependencies are still used to place upper bounds on dependencies which have dropped support for 3.4.

This resolves #242

PS: Might it be worth considering dropping support for Python 3.4 given 3.8 is out now?


teto:
I am not fond of upperbounds like haskell people do: either the bump works (so no point in putting a limit) or it breaks (and you will see it).

teto:
for the tests, seems like we should disable python 3.3 but there seems to be an error in your PR for python 3.4 onm appveyor, that's strange:
```Complete output from command python setup.py egg_info:

    C:\Python34\lib\distutils\dist.py:260: UserWarning: Unknown distribution option: 'python_requires'
      warnings.warn(msg)
    error in papis setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers; Expected ',' or end-of-list in lxml<=4.3.5 ; python_version<='3.5' at  ; python_version<='3.5'
```

JP-Ellis:
> for the tests, seems like we should disable python 3.3 but there seems to be an error in your PR for python 3.4 onm appveyor, that's strange

I didn't pay too close attention to the appveyor builds as they are all failing on all the pull requests; however, indeed Python 3.4 does have a different error and it is simply because the conditional syntax for dependencies was introduced in 2015, which is after Python 3.4 was released.

I would actually recommend we stop supporting Python 3.4 as it is no longer updated (that ended on 2017-08-09) nor patched for security fixes (that ended on 2019-03-18).

teto:
thanks

PR #244, opened 2019-12-02
Update supported Python version to 3.5, 3.6, 3.7 & 3.8

JP-Ellis:
It's probably safe to drop support for Python <= 3.4 as they are neither maintained nor receiving security fixes.  (Unless there's another reason for keeping this support)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,10 @@ language: python
 os:
 - linux
 python:
-- '3.4'
 - '3.5'
 - '3.6'
 - '3.7'
-#- '3.3'
+- '3.8'
 script:
 - python -m pytest papis/ tests/ --cov=papis
 install:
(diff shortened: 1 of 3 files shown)


teto:
I think we can ignore appveyor errors. As for travis, warnings make the tests fail ? Strange behavior :s 
That would be fantastic if you could fix it as well if you have time, otherwise, we'll make it later.

JP-Ellis:
I'll fix these warnings, though it might be better to do this in another pull request (as it's unrelated).  I'll do this tomorrow if I get time.

JP-Ellis:
Looking the failure on Python 3.8, I don't believe the test failed due to the warnings.  I believe it failed because it timed out.  There is a bug with `pytest` whereby it hangs indefinitely.  I'm not entirely sure why this is happening but it has been documented elsewhere online.  I could not find any definite clue as to what to do to fix it though.

See for example: pytest-dev/pytest-xdist#60, pytest-dev/pytest-xdist#110, and [a number of StackOverflow questions](https://stackoverflow.com/search?q=pytest+hang).

teto:
yes it timed out.

Issue #210, opened 2019-08-31
Request for Comment: thoughts on sorting documents


j-c-w:
Hi,

I'd like to implement sorting of results (e.g. with `papis open "*"`), so that results can be sorted on e.g. author name.

Eventually, I'd like to implement add a "time added" field, so that papis could sort by recency.  That's my use-case for this anyway.

Regardless, there seems to be some choice for how best to sort.

I see two ways:

1.  Press <C-s> and have a field appear in which the user can enter a field (with a prompt).
2.  Just type "sort:<field> <rest of query>" into the search bar.

I'm leaning towards 1, but am keen to hear your thoughts on it  and whether there might be another way to do this.

Thanks,

Jackson

teto:
That would be a useful feature. Maybe we can focus on sorting by author name first (the "time added" can come afterwards).
I believe you could add a `--sort` option to papis/commands/open.py and/or pass a sort flag in base.py as well `def query(self, query_string [, sort]):` depending on if some backends can leverage it (can whoosh sort results natively for instance ?).

I think we should only sort according to one field, papis is simple, let's keep it simple :)

j-c-w:
Yep, a flag makes sense, I'll proceed that way.  Thanks for the feedback.

PR #247, opened 2019-12-24
Sorting

j-c-w:
Hi,

This pull request introduces document sorting in the picker as discussed in https://github.com/papis/papis/issues/210.

Use is:

papis <command> <search> --sort <field>

e.g.

papis open "*" --sort title

For whoosh, sort uses the database to sort the results if possible  (i.e. the field is indexed) and otherwise sorts in Python.  For cache, the sorting is always done in python.

Sorting produces the following sort:

1.  Numbers (sorted smallest to largest)
2.  Strings (sorted alphanumerically)
3.  Empty fields.

So, for example if you have the following list of "year" fields: [2014, 1998, Early 10th Century, Late 10th Century, 1055-1058, None], (where None means no 'year' was specified in the yaml file).

And you sort with '--sort year', you get:

[1998, 2014, 1055-1058, Early 10th Century, Late 10th Centure,  None]

I have added a couple tests to check that this all works.

Is this OK to go in?

Jackson


alejandrogallo:
I'll check this on the weekend, but looks promising! Thx!

alejandrogallo:
Thank you again for your contribution.
Please in the future try to submit pull requests that are near or on top of the current master's HEAD.
I have put together the sort option into a decorator as you can see, so that we have to write
less and less. There is a check of papis database that is not passing, I'll
take a look at it later.

I've also added a configuration option `sort-field`, we still have to write
some documentation for it.

There is something else though, shouldn't we add also a '--reverse'
option whenever we offer a '--sort' flag? Since I tend to think so,
it will be a little bit cumbersome to add all this in the database,
I think it would be best to define a different function outside of the database
which does the sorting on the documents. Like this the database code does not
get more complicated and also it's easier to test.

j-c-w:
Hi,

I agree a --reverse flag would be useful, I will get around to that.  I agree that perhaps sorting within the database is not the best idea, partially because according to the whoosh documentation, it's not clear whether there is actually much performance benefit (https://whoosh.readthedocs.io/en/latest/facets.html#making-fields-sortable ).   I did that in line with the comments on the RFC #210 , but I agree that it may not make sense.

IMO papis probably won't be returning more documents than can be sorted effectively in Python.  I'll get around to removing the whoosh sorting part in the next few days.

Thanks,

Jackson

alejandrogallo:
Hi Jackson,

I just had time today and I went and did it.
Now the sort is allways accompanied by `--reverse` and everything
is in the `papis.cli`.

Also, the sorting of numbers and text was "quite" slow with the try
and except block. Positive Numbers converted to string will also get sorted
according to the number order, so we might as well sort everything sorted
in the string sense.

I modified everything to use only a short
sort function in `papis.document.short`. I think your way was also
good however all things considered is easier to do it through the
sort function in the document module.

The results that I get are the same as in your implementation.

If you have time you can take a look at it, but after implementing a test
and add the documentation I think is good enough to merge it to master.

Thanks so much for the great idea and work.
Alejandro

j-c-w:
Yeah, LGTM :), thanks for your help!

PR #251, opened 2020-01-01
Fix sort ordering

j-c-w:
Hi,

With the latest changes to the sorting algorithm, the sorting order was no longer as intuitive as it could be.

For example, in the current code, sorting by 'year' with years: [914, 1066, 1929, 2019] results in [1066, 1929, 2019, 914].

Similarly, the current sorting algorithm brings 'None' values to the top of the sort, rather than the end, where IMO they more intuitively belong.

Unlike my first attempt at this sorting algorithm, this doesn't sort negative numbers properly, which means it can use the (faster) isnumeric() function rather than the slower try/except block.

Is this OK to go in? 

Jackson

### DIFF
--- a/papis/document.py
+++ b/papis/document.py
@@ -395,6 +395,19 @@ class Document(dict):
         """
         return key in self
 
+    def sort_for_key(self, key):
+        # The tuple represents:
+        # (is not None?, is an integer?, integer value, string value)
+        if key in self.keys():
+            if str(self[key]).isdigit():
+                return (False, True, int(self[key]), self[key])
+            else:
+                return (False, False, 0, self[key])
+        else:
+            # The key does not appear in the document, ensure
+            # it comes last.
+            return (True, False, 0, '')
+
     def save(self):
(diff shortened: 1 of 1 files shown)


alejandrogallo:
Hi,
thank you for this, it sounds reasonable. I'll take a look at it and merge it when I'm done
implementing typing for papis, which hopefully will be this weekend.
Cheers,
Ale

j-c-w:
Since performance was a concern before, I thought I would go through and check the performance of these two approaches to sorting, for 100,000 items to sort (i.e. far more than one would reasonably expect papis to handle) is:

Current approach: 0.085 to 0.090 seconds
Approach involving sorting numbers correctly: 0.160 to 0.166 seconds

Script for the comparison is here:
[sorting_speed_comparison.py.txt](https://github.com/papis/papis/files/4049173/sorting_speed_comparison.py.txt) (rename to .py then run `python3 sorting_speed_comparison.py`)

alejandrogallo:
I meant to close this! sorry

PR #256, opened 2020-01-19
Document timestamps

j-c-w:
Hi,

This pull request introduces several changes.  The aim is to allow papis to sort documents by how recently they were added, with the most recently added documents appearing first.

First, it builds on sorting algorithm introduced in #251; I can't figure out how to make Github's diff actually show that this is just building on that, but commit 81901e is just that patch.

a8f2a4a adds a timestamp field to each document as it is added with the `papis add` command.  The format is human readable (as opposed to UNIX time) since that seems most in-line with papis's design.

4f15f24 updates the sorting algorithm to parse these dates properly and sort appropriately on them.  This enables sorting on time added.

c766995 again changes the sorting algorithm: it changes --reverse to not mean true reversal, but rather, to mean reversal of documents with a field instead.  Here's an example of what I mean:

Lets say we have documents with 'year' values of 'None', 1988, 2000.

In the old algorithm, this would sort as: [1988, 2000, None], and using --reverse, this would become [None, 2000, 1988]. 

Using this version, this sorts as: [1988, 2000, None], and using --reverse turns the sort into [2000, 1988, None].  I think this is a more intuitive reverse option.

This last part is relevant to this patch in particular, because to find the most recently added document, you can now do `papis edit "*" --sort time-added --reverse` in libraries where many documents do not already have added times.

Is this good to go in?
(Sorry, don't know why there is a merge conflict)

Jackson

### DIFF
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -94,6 +94,8 @@ import re
 import tempfile
 import hashlib
 import shutil
+import subprocess
+import time
 import papis.api
 import papis.pick
 import papis.utils
@@ -101,6 +103,8 @@ import papis.config
 import papis.document
 import papis.importer
 import papis.cli
+import papis.yaml
+import papis.strings
 import click
 import colorama
 import papis.downloaders
@@ -540,6 +544,8 @@ def cli(files, set_list, subfolder, folder_name, file_name, from_importer,
     if not ctx:
         logger.error('there is nothing to be added')
         return
+
+    ctx.data['time-added'] = time.strftime(papis.strings.time_format)
 
     return run(
         ctx.files,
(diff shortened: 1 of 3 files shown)


alejandrogallo:
Ok, I accept this, thank you very much, I think it is a good Ansatz.
However I simplified and typechecked your code. There is no need to make that function a
method of the Document class, we can pack eeryhing in the sort function. I'll merge it to master.

alejandrogallo:
commit Does the trick ec99deb981dbd24b5e04566ed7a3bdaf8f05619f

j-c-w:
I'll rebase this onto your changes to #251, probably sometime this weekend.

j-c-w:
See #259

PR #257, opened 2020-01-22
Move the sort call in `papis list` so it is applied before the picker

j-c-w:
Hi,

Previously,

`papis list "*" --sort ...` did not actually sort anything, becuase the sorting function was applied after the picker.

This patch moves the application of the sorting function to before the picker to fix this bug.

Good to go in?

Jackson


alejandrogallo:
brilliant thank you!

PR #258, opened 2020-01-24
Add papis --profile to enable debugging when papis is slow

j-c-w:
Hi,

papis was  acting a bit slow for me the other day, so I added a command to debug that.  Command use is:

`papis --profile <filename to dump into> command`

It doesn't work very well with things like `open "*"` where the picker can take a lot of time but  is good for working out why the backend is being so slow.

OK to go in?

Jackson

### DIFF
--- a/papis/commands/default.py
+++ b/papis/commands/default.py
@@ -30,7 +30,7 @@ import os
 import difflib
 import sys
 import logging
-from typing import Optional, Tuple, List
+from typing import Optional, Tuple, List, Callable
 
 import papis
 import papis.api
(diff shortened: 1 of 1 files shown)


papis/commands/default.py
```diff
@@ -107,6 +110,19 @@ def get_command(
         return cli
 
 
+def generate_profile_writing_function(profiler: cProfile.Profile,
+                                      filename: str) -> Callable[[], None]:
+    def _on_finish():
```
alejandrogallo: Add type annotation for  return value (None)

alejandrogallo:
Very Good idea. Please add the type annotation -> None to the internal function so that mypy does not complain anymore and then I’ll merge it.

j-c-w:
Thanks for the review, fixed :)

PR #259, opened 2020-01-26
Document Timestamps

j-c-w:
Hi,

(Replaces #256 )

This patch series introduces timestamps when documents are added to papis, using a `time-added` field in the document metadata.
The big benefit of these timestamps is that they allow document sorting by recency of addition.

The patches are as follows:

4bdb4b:  In add.py, put a timestamp in the document metadata.

9baf197 and 031b46c modify the sorting algorithm in a way that makes sense regardless, but makes the following changes easier.

a660ce7: Adds a special condition to the sorting algorithm to check if each document has a field called `time-added`.  If it does, that document is sorted by date instead.

Notes: There are a couple things I have left out here that could go in.  First, `time-added` is a human-readable date.  An alternative that would not require change 2cb2ec9 is to use Unix time, but IMO this is much less readable.  IMO there could also be some config file settings to enable/disable this, any thoughts on that?

Is this good to go in?

Jackson

### DIFF
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -99,6 +99,7 @@ import shutil
 import click
 import colorama
 
+import time
 import papis.api
 import papis.pick
 import papis.utils
@@ -108,6 +109,7 @@ import papis.config
 import papis.document
 import papis.importer
 import papis.cli
+import papis.strings
 import papis.downloaders
 import papis.git
 
@@ -570,6 +572,8 @@ def cli(
     if not ctx:
         logger.error('there is nothing to be added')
         return
+
+    ctx.data['time-added'] = time.strftime(papis.strings.time_format)
 
     return run(
         ctx.files,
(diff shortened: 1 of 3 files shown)


j-c-w:
> This pull request **introduces 1 alert** when merging [2cb2ec9](https://github.com/papis/papis/commit/2cb2ec91f00e9d83389bbf55933feb494dd132d8) into [cbedcd8](https://github.com/papis/papis/commit/cbedcd87a3c3c123b447e57fb41073a8077c891d) - [view on LGTM.com](https://lgtm.com/projects/g/papis/papis/rev/pr-5729af71cfe55eabe3cbc6a60a7789b73b5e45b2)
> 
> **new alerts:**
> 
> * 1 for Unused import

Addressed

j-c-w:
Actually fixed and good to be reviewed now :)

alejandrogallo:
Sorry for the delay.
I reviewed and it looks very nice.
I made it that users can also opt out of this feature,
but it's good because it is not too invasive.
I thought about setting the flag like
`papis-time-added` but it would be overly verbose. It's ok like this
I think, and it's for sure very useful, thanks mate!

PR #264, opened 2020-02-24
Modernized PR #177

### LINKED ISSUE #134: Exporting --to-bibtex is semi-broken
https://github.com/papis/papis/blob/b023ca0e551a29c0c15f73fa071addd3e61fa36d/papis/document.py#L129

I admit this is partly my fault from an earlier pull request wherein I wanted to explicitly map some papis fields to their proper bibtex equivalents. In doing so, I wanted to add/modify as few lines as possible. Everything was working okay until the `sorted` call was added in the line referenced above. Calling `sorted` actually creates a new list. Thus when I add these 'translated' keys back into the `document` dict, the _sorted_ version remains unchanged, and these fields are never added to the bibtex string, breaking the intended functionality.

I agree that having the fields sorted would be nice, which is why I'm opening the issue to see how best to resolve the dilemma. Also, I'm an engineer and am not very good at writing "real" software :smile:.

avonmoll:
Performed bugfix on current code base. Original PR message:

> Moved the translation of keys to proper bibtex keys to its own loop.
>
> Avoids ambiguity with the loop iterator in the following loop for bibKey in sorted(document.keys())
> Fixes #134

### DIFF
--- a/papis/bibtex.py
+++ b/papis/bibtex.py
@@ -231,12 +231,12 @@ def to_bibtex(document: papis.document.Document) -> str:
     logger.debug("Used ref=%s" % ref)
 
     bibtex_string += "@{type}{{{ref},\n".format(type=bibtex_type, ref=ref)
-    for bibKey in sorted(document.keys()):
-        logger.debug('%s : %s' % (bibKey, document[bibKey]))
+    for bibKey in document.keys():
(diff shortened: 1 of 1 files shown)


alejandrogallo:
Looks good. Thanks!

PR #269, opened 2020-03-29
Update PubMed Importer

alexfikl:
The [current url](http://pubmed.macropus.org) for retrieving pubmed data seems to be dead. This PR switches the importer to use https://api.ncbi.nlm.nih.gov/lit/ctxp instead. This seems to be offered by the NCBI so hopefully it works well enough. 

There's also [TexMed](https://www.bioinformatics.org/texmed/), which seems to offer results directly in bibtex. Would you prefer querying that?

### DIFF
--- a/papis/pubmed.py
+++ b/papis/pubmed.py
@@ -1,34 +1,117 @@
-from typing import Optional
+from typing import Any, Dict, List, Optional
+
+import json
+import urllib
+import requests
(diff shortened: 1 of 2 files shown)


alejandrogallo:
Thanks it looks fine, I'll make sure everything holds for `0.10`

PR #272, opened 2020-04-26
Mercurial support

xaltsc:
I've started working on mercurial support for Papis, which may be used as a replacement or alongside with git.

I'm not really respecting the coding style at the moment, I'm just advertising the fact that I'm working on it.

Comments are welcome, especially, mercurial provides a python library which could perhaps be used instead of calling the commands with `subprocess.call`.

As I'm merely adding functionnality for the moment and Mercurial has no notion of staging, there may be uncessary adds in the code. I'm trying to model the git module, but git and hg behave differently.

### DIFF
--- a/papis/cli.py
+++ b/papis/cli.py
@@ -93,6 +93,17 @@ def git_option(
         return click.decorators.option('--git/--no-git', **attrs)(f)
     return decorator
 
+def hg_option(
+        help: str = "Add mercurial interoperability",
+        **attrs: DecoratorArgs) -> DecoratorCallable:
+    """Adds a ``hg`` option as a decorator"""
+    def decorator(f: DecoratorCallable) -> Any:
+        attrs.setdefault(
+            'default',
+            lambda: True if papis.config.get('use-hg') else False)
+        attrs.setdefault('help', help)
+        return click.decorators.option('--hg/--no-hg', **attrs)(f)
+    return decorator
 
 def bypass(
         group: click.core.Group,
--- a/papis/cli.py
+++ b/papis/cli.py
@@ -93,6 +93,7 @@ def git_option(
         return click.decorators.option('--git/--no-git', **attrs)(f)
     return decorator
 
+
 def hg_option(
         help: str = "Add mercurial interoperability",
         **attrs: DecoratorArgs) -> DecoratorCallable:
@@ -104,6 +105,7 @@ def hg_option(
         attrs.setdefault('help', help)
         return click.decorators.option('--hg/--no-hg', **attrs)(f)
     return decorator
+
 
 def bypass(
         group: click.core.Group,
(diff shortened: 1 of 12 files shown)


alejandrogallo:
Hello and thank you for this.

I'm quite wondering if we couldn't use the `git`
functionality by just changing `git` into `hg`, what
do you think as a `mercurial` user, as far as I see and know
the only real difference between the two apis for `mv, commit` and `add` is that `-C` turns to `--pwd`.

However this is also quite easily fixed at some later point,
maybe it's better to keep everything separated for clarity.

On a different note, I was also thinking of another approach.
To create a general `cvs` functionality.
I'm assuming that every library is only on one kind of `cvs`.
Then we could have templates for committing, adding and so on.
So in the case of git it would be

```
cvs-commit-cmd = git commit -m "{msg}" {files}
cvs-add-cmd = git add {files}
...
```

for `svn`, `hg`, `darcs` it would be similar.
Since we're not targeting at a full-blown compatibility
layer, we only want simple renaming, edit etc...
Maybe this would be enough.

In this case, on a library basis we could have a `hg`
library, a `git` library...

alejandrogallo:
however as a first step towards that direction, your approach seems good to me

xaltsc:
> Hello and thank you for this.
> 
> I'm quite wondering if we couldn't use the `git`
> functionality by just changing `git` into `hg`, what
> do you think as a `mercurial` user, as far as I see and know
> the only real difference between the two apis for `mv, commit` and `add` is that `-C` turns to `--pwd`.
> 
> However this is also quite easily fixed at some later point,
> maybe it's better to keep everything separated for clarity.
> 
> On a different note, I was also thinking of another approach.
> To create a general `cvs` functionality.
> I'm assuming that every library is only on one kind of `cvs`.
> Then we could have templates for committing, adding and so on.
> So in the case of git it would be
> 
> ```
> cvs-commit-cmd = git commit -m "{msg}" {files}
> cvs-add-cmd = git add {files}
> ...
> ```
> 
> for `svn`, `hg`, `darcs` it would be similar.
> Since we're not targeting at a full-blown compatibility
> layer, we only want simple renaming, edit etc...
> Maybe this would be enough.
> 
> In this case, on a library basis we could have a `hg`
> library, a `git` library...

I was wondering the same thing, however, I decided against it because I thought it would be useful to be able to make the papis folder both a git repository and a mercurial repository.

Also, `hg` really behaves in a different way regarding `add` and `commit` commands: `add` is only useful to add untracked files, and `commit` by default adds and commits everything that has changed in the repository, so I thought it would be better to keep things separate.

xaltsc:
Unless there are some changes you wish to request, I think the PR is ready to merge. I've been using it for a week now, without any problems.

PR #276, opened 2020-06-25
Remove { } and \ from the characters translated to latex

j-c-w:
Hi,

When exporting to bibtex, I often find that my documents need a post-export fixup.

This is OK if I only ever export once, but if I export multiple times (e.g. creating different .bib files for different documents), the task of manually fixing the bibtex each time is a bit tiring.

This patch aims to solve this problem by removing key formatting characters from the auto-translator; allowing me to use "{", "}" and "\" directly in the yaml files.  This means I can format my documents properly for latex once in the yaml file, and they are correct each time I run `papis export`.

I'm not certain that this is the best way to achieve this, because it removes some of the format-independence of the papis yaml format.  For my usecase, this is not an issue --- I exclusively use bibtex outputs.  However, I am interested to hear if anyone has any thoughts about how to do this while maintaining output format independence.

Is this OK to go in?

Jackson

EDIT: I've bumped this to include "$" and "_", two other characters I was running into difficulties translating.

### DIFF
--- a/papis/bibtex.py
+++ b/papis/bibtex.py
@@ -291,20 +291,23 @@ def unicode_to_latex(text: str) -> str:
 
     # from six import text_type
 
+    # Some characters we choose not to translate to allow for latex
+    # formatting within yaml files.
+    #    u"\u007B": r"\lbrace{}",
+    #    u"\u007D": r"\rbrace{}",
+    #    u"\u005C": r"\textbackslash{}",
+    #    u"\u005F": r"\_",
+    #    u"\u0024": r"\textdollar{}",
+
     unicode_to_latex_table_base = {
         u"\u0023": r"\#",
-        u"\u0024": r"\textdollar{}",
         u"\u0025": r"\%",
         u"\u0026": r"\&",
         u"\u0027": r"'",
         u"\u002A": r"\ast{}",
-        u"\u005C": r"\textbackslash{}",
         u"\u005E": r"\^{}",
-        u"\u005F": r"\_",
         u"\u0060": r"\textasciigrave{}",
-        u"\u007B": r"\lbrace{}",
         u"\u007C": r"\vert{}",
-        u"\u007D": r"\rbrace{}",
         u"\u007E": r"\textasciitilde{}",
         u"\u00A0": r"~",
         u"\u00A1": r"\textexclamdown{}",


j-c-w:
I've discovered the bibtex-unicode flag, setting it to true seems to be good enough for me, so I'm closing this for now.

avonmoll:
One potential use case for this is the following:

A lot of my papis yaml files have unicode in them (e.g. in peoples' last names). Sometimes I work with LaTeX class files that only really work well with pdfLatex and thus unicode characters need to be "texified" (e.g. the IEEE conference template). Sometimes my yaml file will have a title like `{PILCO}: {A} model-based and data-efficient approach to policy search`. The curly braces are meant to enforce capitalization in BibTeX -- otherwise, certain bibliography styles impose capitalization schemes that may render as `Pilco: a model-based...`, which is undesirable. So I've run into cases where I set `bibtex-unicode = false` and then end up with a bunch of `\lbrace{}` and `\rbrace{}` everywhere. From this standpoint, your originally proposed change is helpful.

On the other hand, I don't really like having BibTeX-related 'quirks' inside my YAML files, and I wonder if the capitalization brackets can be inserted at export time. For example, any word in the title that is all uppercase in the YAML should be wrapped with `{}` when exporting to BibTeX. Apologies for my ramblings.

j-c-w:
Hi,

Yes, I completely agree re your use-case.  I don't think that Bibtex's heuristics for what to leave capitalized (i.e. nothing) are sensible ones, but finding a sensible one is actually quite difficult, e.g., if I have a document named: "An exploration of Facebook's approach to computational accelerators", I would quite like to have the exported bibtex be "An exploration of {Facebook's} approach to computational accelerators.

Since bibtex is a sole-target export for me, I think I'm going to integrate the bibtex formatting into my references.

Basically, the flaw I see here is that the YAML formatting is less expressive than bibtex formatting --- so moving to bibtex requires some heuristics like the one you suggest.  Ultimately, I think that the problem could be solved either with a markup language that is as expressive as bibtex's, or by just using bibtex's markup language.  I should also note that there are bugs with the current character-by-character translator:

"μIR: ...." turned into "\mathrm{\mu}IR", which causes an error because \mathrm should only exist in mathmode.  So something slightly 'smarter' is needed to convert to bibtex as-is --- my current solution to this is to write out the $\mathrm{\mu}$ directly into my YAML file, but I also agree that the automatic character conversions seem quite helpful.

PR #278, opened 2020-07-23
escape strings in get_regex_from_search

d70-t:
When a string with regex-relevant data is passed into this function, these characters should be escaped. Otherwise, invalid regexes could appear and lead to strange errors.

I think the double `.*` at the beginning is also not ideal, but I am not entirely sure if my edit is the most sensible choice.

The problem which occured in my use case was that I had some bibtex which I wanted to import and that included ISBN numbers which where placed in curly braces. In combination with this function, this lead to regexes like `.*{123456}` which are not valid.

### DIFF
--- a/papis/database/cache.py
+++ b/papis/database/cache.py
@@ -139,7 +139,7 @@ def match_document(
 
 
 def get_regex_from_search(search: str) -> str:
-    """Creates a default regex from a search string.
+    r"""Creates a default regex from a search string.
 
     :param search: A valid search string
     :type  search: str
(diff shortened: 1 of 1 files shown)


alejandrogallo:
I think you're right, this is a good fix, thanks so much, this actually fixed a problem
I had with dois where the dois had parenthesis in them and I could not match them.

About the two `.*` at the beginning, it doesn't actually matter since `.*.* = .*` in *regex arithmetic*.
Thanks a bunch!

PR #294, opened 2020-11-27
Allow to escape dots and underscores with a backslash in ref-format

prataffel:
Hey,

first of all: thanks for writing papis, its a very useful tool for me. 

While the cleanup when generating a ``ref`` introduced in 216616839754771602ec29f4f2a10d111ea409a7 is very useful, I'd like to place a dot in my refs (e.g. to divide author and year). Therefore, I would like to suggest the possibility to escape characters in `ref-format` -- here implemented for ``.`` and ``_``.

Specifying ``ref-format = {doc[author]:.10}\.{doc[year]}`` would result in refs like ``MeshKiAngN.2020``.

### DIFF
--- a/doc/source/default-settings.rst
+++ b/doc/source/default-settings.rst
@@ -249,6 +249,21 @@ Bibtex options
     ``mesh-ki-ang-nuna`` then the built reference will be
     ``StudiesAboutEMeshKi``.
 
+    .. note::
+        Special characters will be replaced when generating the ``ref`` entry
+        (e.g.  ``Ö → O``, ``.`` and other symbols will be striped from the
+        string). 
+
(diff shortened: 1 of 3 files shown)


alejandrogallo:
Ok, I think it makes sense, could you add two test next to the existing in

https://github.com/papis/papis/blob/9c9490d1ee744e248caa95d29334e0cee65fcec1/tests/test_bibtex.py#L73

Thanks!

alejandrogallo:
and just a line documenting that you can escape `.` and `_` in
https://github.com/papis/papis/blame/master/doc/source/default-settings.rst#L229

alejandrogallo:
Wonderful thanks !

PR #301, opened 2020-12-05
Add a py.typed file

alexfikl:
Allows plugins that use `papis` to also be type checked. From https://www.python.org/dev/peps/pep-0561/#packaging-type-information.


alejandrogallo:
I wasn't aware! Thanks!

PR #302, opened 2020-12-07
Improve isbn_to_papis conversion

alexfikl:
Extracts a papis friendly `author_list` from isbnlib's list of authors.

The added test data has an example of before/after.


alejandrogallo:
Is this ready to go? It looks fine

alexfikl:
Yes, it should be good to go from my side. 

Thanks for taking a look!

PR #312, opened 2020-12-27
Stricter file download checks in crossref

alexfikl:
When running 
```
papis add --from doi '10.1007/s00607-002-1450-4'
```
the importer would get a `doc_url` of http://link.springer.com/article/10.1007/s00607-002-1450-4/fulltext.html, which would get redirected to the default paper page because of lack of access, and then download that as an html document.

I'm not sure disallowing redirects is the best way to handle that, since there are legitimate uses for them. What do you think?

### DIFF
--- a/papis/crossref.py
+++ b/papis/crossref.py
@@ -344,24 +344,41 @@ class Importer(papis.importer.Importer):
         return None
 
     def fetch(self) -> None:
-        self.logger.info("using doi {0}".format(self.uri))
+        self.logger.info("Using DOI '%s'", self.uri)
         doidata = papis.crossref.get_data(dois=[self.uri])
         if doidata:
             self.ctx.data = doidata[0]
-            _doc_url_keyname = papis.config.getstring('doc-url-key-name')
-            if _doc_url_keyname in self.ctx.data.keys():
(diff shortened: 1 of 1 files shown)


alejandrogallo:
Hmm so what happens if we disallow redirects, what do I get in this case?

alexfikl:
> Hmm so what happens if we disallow redirects, what do I get in this case?

With this PR, it basically won't save any files resulting from a redirect, which doesn't sound like a great idea. 

The issue I was trying to fix is just to not have papis download html webpages when I don't have access to the actual paper. Maybe checking the file type of the downloaded document against a whitelist is better? What do you think?

alexfikl:
Changed the importer to only save the files if
* the HTTP status code is `OK (200)`, but still allowing redirects
* the downloaded contents are recognizable by `filetype`, which excludes plaintext like html.

Does that sound more reasonable?

alejandrogallo:
I think this is reasonable, let's roll with that for the time being.

alexfikl:
> I think this is reasonable, let's roll with that for the time being.

Thanks! Feel free to ping me for a fix if it brakes people's workflow!

PR #314, opened 2020-12-30
Fix deadlock when filtering with the list widget

alexfikl:
Before trying to open a document with `papis open` and then starting to type would freeze the tui completely.

### DIFF
--- a/papis/tui/widgets/list.py
+++ b/papis/tui/widgets/list.py
@@ -240,6 +240,7 @@ class OptionsList(ConditionalContainer, Generic[Option]):  # type: ignore
                     for i, opt in enumerate(self.options_matchers)
                     if i in search_indices
                 ]
+                results = [d.get() for d in results]
         else:
             results = [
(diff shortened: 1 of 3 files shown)


papis/tui/widgets/list.py
```diff
@@ -240,15 +240,15 @@ def filter_options(self) -> None:
                     for i, opt in enumerate(self.options_matchers)
                     if i in search_indices
                 ]
+                results = [d.get() for d in results]
```
alexfikl, line 243: Moving the `get()` inside the `with Pool..` is the actual fix to the problem I was having. It seems to have been incorrectly indented in #308.

Is the `get()` still necessary if the pool gets closed/joined after exiting the `with` block? (personal question as I'm not that familiar with the multiprocessing module)

alejandrogallo:
Oh yes, thanks for that, I think we should define a `utils.map` function that takes care of this multiprocessing stuff automatically there, I'll do this at some point.

Issue #316, opened 2021-01-12
[Enhancement] Allow opening/creating a note from picker with ctrl+foo


jghauser:
It would be great if it were possible to open/create an attached note from the default picker with `c+foo`. This would be analogous to opening an attached file (`c-o`) or editing the yaml (`c+e`).

My use case for this is that I like to have the picker open with all documents that relate to a certain project. From there I then open the files as I need them without having to start the picker again every time I open a document. It would be great if this worked also for notes.

I'm not sure what the best shortcut would be, seeing that c-n is already taken.

If that's something you'd like to see, I'd be happy to give a shot at implementing the feature. I'm not much of a programmer, but with a pointer or two in the right direction, I think I should able to do it.

alejandrogallo:
This is doable, you can basically
play with the file
`papis/tui/app.py`

There you find the function
`get_commands` which basically gets the commands
and assigns them to keybindings for the picker.

This part of papis is yet to be documented, but you
can add there a new command following the example of
the edit command there, i.e.:

```python
    @kb.add(keys_info["edit_document_key"]["key"],  # type: ignore
            filter=has_focus(app.options_list.search_buffer))
    def edit(cmd: Command) -> None:
        from papis.commands.edit import run
        docs = cmd.app.get_selection()
        for doc in docs:
            run(doc)
        cmd.app.renderer.clear()
```

I hope this helps you to get started experimenting.

jghauser:
Thanks for the help! I'll see how I can make this work and report back once I got something.

jghauser:
This has been implemented (see merged pull request above)!

PR #319, opened 2021-02-07
[WiP] Add open note functionality to picker (issue 316)

jghauser:
This commit adds possibility to edit notes from the default picker (see #316). 

Currently, I set the default keybinding to `c-m` but that's mostly just to have anything (both `c-n` and `c-e` which would seem like the obvious choices are already in use).

With this commit, I want to get some feedback on how to best implement the feature. What I've done is to separate out, in `edit.py` from the function`cli` everything that relates to editing notes and created a new function (`edit_notes`) with it. This is so i can call that function in `app.py`. I'm not sure this is the best way of doing this; it breaks a bit with how the other command files are structured. I've also commented out a bunch of logging stuff as maybe we want to call these only when the function is called as a cli command?

What do you think is the best way of implementing this?

I ran the tests and a bunch of them failed, but I'm not sure this is because of my changes. I'm rather new to github and only slightly more experienced with python -- so, happy about any pointers on what to do!

### DIFF
--- a/papis/commands/edit.py
+++ b/papis/commands/edit.py
@@ -42,6 +42,31 @@ def run(document: papis.document.Document,
                 papis.document.describe(document)))
 
 
+def edit_notes(document: papis.document.Document,
+               git: bool = False) -> None:
+    logger = logging.getLogger('edit:notes')
+    logger.debug("Editing notes")
+    if not document.has("notes"):
+        document["notes"] = papis.config.getstring("notes-name")
+        document.save()
+    notes_path = os.path.join(
+        str(document.get_main_folder()),
+        document["notes"]
+    )
+
+    if not os.path.exists(notes_path):
+        logger.debug("Creating {0}".format(notes_path))
+        open(notes_path, "w+").close()
+
+    papis.api.edit_file(notes_path)
+    if git:
+        papis.git.add_and_commit_resource(
+            str(document.get_main_folder()),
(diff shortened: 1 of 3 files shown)


alejandrogallo:
Hi Julian,

first of all great work! And I'm sorry for the delay in my answer.

Actually I don't mind at all the way you implemented this,
for me it looks more or less the same as what I'd have done.

About the logging, we can just put the `logging.info` that
was in the notes section as `logging.debug` logging, this way
it will only be shown if we really want it to be shown.

What I can do is test your code and see where it breaks, and then I can
fix it if there is something to be fixed.

Are you planning to continue on this? Are you using this already in your fork?

Cheers!

alejandrogallo:
I have nothing to complain about in your **PR** as-is now, I just
changed the keybinding to `c-q` since `c-m` is traditionally reserved for enter,
and some people like me don't really use the enter key (sometimes we are on keyboards
without enter key.. ;)), but yeah, you can always change it to `c-m` in your config file,
as far as I know `c-q` is not really used for anything

jghauser:
Looks great, I'm totally fine with the shortcut being `c-q`, my pick was really mostly random. I've used this in my fork and it seems to work great. If there's anything to fix/change, let me know, otherwise this is done on my end :)

alejandrogallo:
I squashed the commits and merged it, thanks @jghauser, great work!

PR #331, opened 2021-06-04
Automatically retrieve DOI from PDF

stuartthomas25:
I think this feature is fairly critical and easily implemented. This change may belong in `get_matching_importer_or_downloader' but for some reason that wasn't working.

Also, there may be a way that does not rely on pdf2doi (MicheleCotrufo/pdf2doi).

### DIFF
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -98,6 +98,7 @@ import shutil
 
 import click
 import colorama
+import pdf2doi
 
 import time
 import papis.api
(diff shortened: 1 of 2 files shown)


alejandrogallo:
Hello stuart,

thank you for your interest and your effort.
I don't quite understand this pull request, as papis
already supports this
through [python-doi](https://github.com/papis/python-doi/blob/master/src/doi/__init__.py#L12).

Maybe this is probably not clear enough documented, but if you do

```
papis add ~/path/to/doc.pdf
```

and it is possible to retrieve a doi from the pdf, then the linked function
here retrieves it, mostly correctly (also arxiv identifiers).

Is this the behavior you were looking for?
Should I note this explicitly in the documentation?

stuartthomas25:
Hi Alejandro,

Thanks for the clarification. That wasn't working for a specific PDF and I assumed the feature didn't exist. I think I got it working. 

Thank you!

alejandrogallo:
don't worry, I hope I did not discourage you to get involved further, anyways I should document this feature, I'm opening an issue for this

PR #344, opened 2021-11-26
Add a downloader for Project Euclid

alexfikl:
Some papers on [Project Euclid](https://projecteuclid.org/) don't have DOIs and it was a bit annoying to get extended info on them, so this adds a downloader. 

It's pretty simple because it has all the standard `meta` headers and `papis.downloaders.base` does the heavy lifting.

EDIT: tests are passing at https://github.com/alexfikl/papis/actions/runs/1508129865.


alejandrogallo:
Sounds wonderful! Thank you very muchq

PR #345, opened 2021-11-27
Clean up logging

alexfikl:
The main change here is to use the `logger.log(fmt, *args)` way as it should be a bit faster according to the [logging module docs](https://docs.python.org/3/howto/logging.html#optimization) (basically it avoids any unnecessary string interpolation).

Besides that, a few random changes along the way:
* use `logger` instead of `LOGGER` in all modules
* capitalize logging messages (probably didn't get them all, but at least a few more than before)
* reworded a few messages as well.
* added a `ColoramaFormatter` to automagically pick up any `{c.Fore.Red}` sort of things without having to pass it to the log calls. Not sure about this one though. What do you think?

### DIFF
--- a/papis/commands/default.py
+++ b/papis/commands/default.py
@@ -48,6 +48,12 @@ import click
 import click.core
 
 
+class ColoramaFormatter(logging.Formatter):
+    def format(self, record: logging.LogRecord) -> str:
+        record.msg = record.msg.format(c=colorama)
+        return super().format(record)
+
+
 class MultiCommand(click.core.MultiCommand):
 
     scripts = papis.commands.get_scripts()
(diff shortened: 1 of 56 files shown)


papis/commands/default.py
```diff
@@ -48,6 +48,12 @@
 import click.core
 
 
+class ColoramaFormatter(logging.Formatter):
+    def format(self, record: logging.LogRecord) -> str:
+        record.msg = record.msg.format(c=colorama)
+        return super().format(record)
```
alexfikl, line 54: Not sure about adding this formatter. What do you think?

It's basically meant to allow using
```
logger.info("{c.Fore.RED}My message: %s{c.Style.RESET_ALL}", msg)
```
without having to also pass in the `colorama` module. One reason I don't like it is because it's going to try to format every message, instead of just those that try to use `colorama`. Maybe adding a very short function like `_c` and then use
```
logger.info(_c("{c.Fore.RED}My message: %s{c.Style.RESET_ALL}"), msg)
```
would be better?

Either way, this is contained only in the last commit and I can revert it without worries.

alejandrogallo: I was not aware of that, I don't think it is such aproblem
to have the formatter always format, I guess there is a way also
of disabling it.

So I'm quite ok with having the formatter like that because
in the end all messages come out formated, so we can afford the luxury
of the ergonomics.

I think it's good.

alexfikl: Yeah, there are ways to disable it, so it should be ok.

Thanks for taking a look and the feedback!

alexfikl:
The tests are also passing at
https://github.com/alexfikl/papis/actions/runs/1509681469
and I've used it a bit today and it didn't break anything obvious.

alejandrogallo:
Alright, sound and looks, I'll go ahead and merge! Thanks Alex!

PR #346, opened 2021-11-28
Do not require typing-extensions on Python 3.8+

paper42:


alejandrogallo:
As far as I've read this make sense, thank you for this!

some ref for future self: https://github.com/nabla-c0d3/sslyze/issues/421

PR #348, opened 2021-11-30
Use CSafeLoader for yaml for faster loading

alexfikl:
See also https://stackoverflow.com/a/50685946.

### DIFF
--- a/papis/yaml.py
+++ b/papis/yaml.py
@@ -1,4 +1,4 @@
-import yaml
+import yaml                                 # lgtm [py/import-and-import-from]
 import logging
 import click
 import os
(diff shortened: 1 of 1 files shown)


alexfikl:
Hm, codacy seems to be complaining, since it probably can't pick up that we're still using the safe loaders. On the bright side, the tests are still passing
https://github.com/alexfikl/papis/actions/runs/1519404098

alexfikl:
For a silly benchmark in papis, before it was
```
papis -l papers list -a shu  0.58s user 0.04s system 99% cpu 0.619 total
```
and now it is
```
papis -l papers list -a shu  0.33s user 0.06s system 99% cpu 0.394 total
```
(my library has around 1000 entries and about 20 match the search)

alejandrogallo:
This looks great, and reasonable. Thanks for putting also the fallback option for the python loader. I think this should be ready 
to go, I'll go ahead and merge

alexfikl:
Thanks!

PR #349, opened 2021-12-03
Remove most third-party libs from top level imports

alexfikl:
This is supposed to make running quick commands (`papis add`, `papis list`, etc) a bit faster by delaying importing bigger modules. 

From a quick profile the main culprits seemed to be `requests` and `bs4`, but I went ahead and cleaned all of them up. The only remaining top level imports in a file now should be
* `click`
* `os`, `re`, `sys`, `logging` and `typing` from python
* whatever `papis` modules are needed.

From a quick unscientific timing of `papis list`, before it was
```
papis -l papers list -a "ar"  0.55s user 0.19s system 143% cpu 0.513 total
```
and now it is
```
papis -l papers list -a "ar"  0.35s user 0.22s system 164% cpu 0.342 total
```  
(for a library of about 1000 entries with about 800 matching the search)

### DIFF
--- a/papis/arxiv.py
+++ b/papis/arxiv.py
@@ -17,14 +17,10 @@
 """
 import os
 import re
-import bs4
 import logging
-import urllib.request  # urlopen, Request
-import urllib.parse  # import urlencode
 from typing import Optional, List, Dict, Any
 
 import click
-import arxiv2bib
 
 import papis.filetype
 import papis.downloaders.base
@@ -64,6 +60,8 @@ def get_data(
         [key+':'+str(clean_params[key]) for key in clean_params]
     )
     logger.debug("query = '%s'", search_query)
+
+    import urllib.parse
     params = urllib.parse.urlencode(
         {
             'search_query': search_query,
@@ -74,6 +72,8 @@ def get_data(
     main_url = "http://arxiv.org/api/query?"
     req_url = main_url + params
     logger.debug("url = '%s'", req_url)
+
+    import urllib.request
     url = urllib.request.Request(
         req_url,
         headers={
@@ -81,6 +81,8 @@ def get_data(
         }
     )
     xmldoc = urllib.request.urlopen(url).read()
+
+    import bs4
     soup = bs4.BeautifulSoup(xmldoc, 'html.parser')
 
     entries = soup.find_all("entry")
@@ -107,10 +109,11 @@ def get_data(
 
 
 def validate_arxivid(arxivid: str) -> None:
-    from urllib.error import HTTPError, URLError
+    import urllib.request
     url = "https://arxiv.org/abs/{0}".format(arxivid)
     request = urllib.request.Request(url)
 
+    from urllib.error import HTTPError, URLError
     try:
         urllib.request.urlopen(request)
     except HTTPError:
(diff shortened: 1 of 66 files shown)


alexfikl:
Screenshot_20211202_204929

For the `papis list` command I tried, it seems like most of the remaining time is spent importing required things (`stevedore` and `pyparsing`), so not much to do there :cry: 

This is the full profile [profile.txt](https://github.com/papis/papis/files/7646486/profile.txt) (you can visualize it on [speedscope](https://www.speedscope.app/)).

alexfikl:
Also, tests are still passing :rocket: 
https://github.com/alexfikl/papis/actions/runs/1533379484

alejandrogallo:
Yes, stevedore is really a problem but at the time I did not know really how one is to solve it or if we should kick out stevedore altogether. It seems to me that we are not really profiting so much from stevedore for what it costs us, but well.

About changing all imports to where they are used, in principle some years ago I did that too, but then in principle sometimes in python is good to have them on top for testing that actually all modules are importable. Anyways I don't know right now what is the good style in python when importing modules and so on, do you have there some kind of resource about best practices there?

I'll look around and see what I find, but in principle I'm ok sacrificing some cleanliness for startup timing optimisation

alejandrogallo:
I made a bench (library size ~1100 and filtering to ar is 600)

```sh
$ hyperfine --warmup 5 "papis list -a ar"

Benchmark #1: papis list -a ar
  Time (mean ± σ):      3.066 s ±  0.050 s    [User: 3.577 s, System: 0.406 s]
  Range (min … max):    3.010 s …  3.154 s    10 runs
```

and with your commit I got

```sh
$ hyperfine --warmup 5 "papis list -a ar"

Benchmark #1: papis list -a ar
  Time (mean ± σ):      2.022 s ±  0.016 s    [User: 2.593 s, System: 0.340 s]
  Range (min … max):    1.999 s …  2.058 s    10 runs
```

so this means that this commit made the whole thing on my machine =33.33333333333333=% faster. I would be inclined to take it.

alexfikl:
> About changing all imports to where they are used, in principle some years ago I did that too, but then in principle sometimes in python is good to have them on top for testing that actually all modules are importable. Anyways I don't know right now what is the good style in python when importing modules and so on, do you have there some kind of resource about best practices there?
> 
> I'll look around and see what I find, but in principle I'm ok sacrificing some cleanliness for startup timing optimisation

The only recommendation I know here comes from [pep8 Imports section](https://www.python.org/dev/peps/pep-0008/#imports), which says that they should be at the top of the file. To my knowledge that's actually only about not mixing in imports and code at the top level, e.g. what this [flake8 rule](https://www.flake8rules.com/rules/E402.html) is checking, so not really applicable.

Another option would be to do some sort of lazy import and leave the packages at the top level (e.g. [this](https://snarky.ca/lazy-importing-in-python-3-7/) for a recent implementation). However, that won't do much for the testing aspect you mentioned, since it would also crash only on first call..

So yeah, TLDR, I don't recall seeing a nicer way to handle this sort of issue with the import times. On the bright side, there [some work](https://github.com/faster-cpython/ideas/discussions/153) on the python side to make this faster!

alexfikl:
> $ hyperfine --warmup 5 "papis list -a ar"

TIL about `hyperfine`! Thanks, that seems like a very useful little tool!

> so this means that this commit made the whole thing on my machine =33.33333333333333=% faster. I would be inclined to take it.

Nice! I might have gone a bit too much in the other direction with some things (e.g. pushing all the `urllib` imports down), which probably don't have a big effect. Let me know if you would prefer leaving some of them at the top level (I can do some benchmarking to check).

alejandrogallo:
No that's ok, at this point there is not so much we can do, I think we have to live with it in order to have this startup time better.
For me it is also quite bothering, but ok, I thnk as soon as there is some alternative to stevedore that would be good, but at the end of the day stevedore can't do it also faster, because it's looking also for the moduels, and I already discussed with someone that probably caching would not be a good idea in general. Anyways, for me this is good to go then.

Thank you also for clarifying the points about the standard, it's been awhile since I took a look at the PEP stuff.
And thank you for your great work!

PR #353, opened 2021-12-10
Fix logger message in MultiCommand class

alexfikl:
Another fix after #345 :(. I'm starting to think that `ColoramaFormatter` thing is more trouble than it's worth!

The issue here was that logger in the `MultiCommand` class gets created and stored before the `default` command gets a chance to execute and set up all the logging infrastructure. This means that `ColoramaFormatter` does not get set and called, so the logged message contained the `colorama` codes.


alejandrogallo:
ah yes, well this is fair enough ;) thanks!

Issue #323, opened 2021-02-23
Slow response during papis open


aborzunov:
I was following quick start guide, added two papers and stumbled over big latency inside `papis open`.

It takes ~3 sec to update search results after typing any letter. Im using `papis, version 0.11.1`. This problems occurs only on macOS 11.2.1, iterm2.

I tried `papis open` on ubuntu 20.04 with pretty same environment and didn't face this problem, it responded instantly.

alejandrogallo:
I have no obvious answer to give you, I'm sorry, this might be very well related to #321 , which would be interesting.
Of course it can not depend on `iterm2` in your case.

j-c-w:
Hi,

Could you do a run with a `--profile papis_profile` flag and attach the output?  (e.g. `papis --profile papis_profile open`?) 

Ideally you'd be trying to spend as much time in the run making it lag as possible so the profile doesn't just show the time the GUI spends waiting for inputs.

Jackson

haykh:
Hi! I'm having the same trouble on mac OS 11.3.1, the requests are super slow. Both using `iterm2` and `terminal`. Here's the output of `--profile` for my case (`PATH` is where my conda env is located):

[papis_profile.txt](https://github.com/papis/papis/files/6446457/papis_profile.txt)

If it helps, `papis` is install via `pip` under one of my `conda` envs.

alejandrogallo:
Thank you for the profile, this is indeed useful to see what is going on. Now I think this is related to the parallel processing that papis is doing, apparently the python version and the OS that you're using is having issues with it, I'd say that
running then everything on a single core would solve the issue,

if you're using master try running thus
```
PAPIS_NP=1 papis open
```
to run only using 1 thread. Maybe this will help.

haykh:
Thanks for the response! It actually helps but only slightly. The query is still far from being responsive. Still have to type and wait for a second for filter results to show up. And, yes, this is with the `master` version. Funny thing is I also use `papis` on my linux machine. There I don't have any problem and it's super responsive. 

Here's what it looks like on my mac:
https://user-images.githubusercontent.com/7251306/117589371-03acd080-b0f7-11eb-812f-094d1b4b33bf.mov

And the profile for `np` set to `1`. 
[profile_master_np1.txt](https://github.com/papis/papis/files/6448566/profile_master_np1.txt)

[...]

PR #355, opened 2022-01-25
Disable multiprocessing on darwin systems

haykh:
Temporarily solves #323

### DIFF
--- a/papis/utils.py
+++ b/papis/utils.py
@@ -1,4 +1,5 @@
 import os
+import sys
 import re
 import logging
 from itertools import count, product
@@ -32,7 +33,7 @@ def has_multiprocessing() -> bool:
 def parmap(f: Callable[[A], B],
            xs: List[A],
            np: Optional[int] = None) -> List[B]:
-    if has_multiprocessing():
+    if has_multiprocessing() and sys.platform != 'darwin':
         np = np or os.cpu_count()
         np = int(os.environ.get("PAPIS_NP", str(np)))
         with Pool(np) as pool:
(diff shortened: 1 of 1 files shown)


alejandrogallo:
Hy @haykh, 
thank you for your pull request. I am not sure this is the way to go that some macos user will not be able to use
parmap, but it is a little bit hard to decide as I do not use macOS. Maybe @avonmoll can comment on this request.
According to your last comment on the #323 thread you said that the overhead for the thread creation
was almost unnoticeable, at least as far as I can see, is this fixing this minimal overhead?

Anyways I have no problem merging this for the time being, but I'd like to be sure that it is for this minimal overhead.
In this case I'd ask from you to add a TODO comment with the link to the #323 request for future reference.
thanks!

avonmoll:
It's a big plus for me because I'm essentially already using it this way and can now update safely without overwriting the 'patch' from #323. For general mac users it's hard to say since I don't have access to a machine for which the multi-processing stuff worked fine on (for pythons >3.6). The only thing I can say is that papis _feels_ snappy for me now (with nearly 500 docs) and I imagine the user experience will still be decent for the average mac user.

alejandrogallo:
Ok, that sounds reasonable. I'll merge it, @haykh could you add the `TODO` notice on the line after the `if`?

haykh:
@alejandrogallo the overhead is minor but noticeable. Since the whole purpose of multiprocessing is to boost performance I thought it sort of defies that idea. I added a todo comment for the future, but, yes, like @avonmoll says, it never worked for me on python past v3.6/3.8. I'll give this a more proper look when I have a bit more time.

alejandrogallo:
Brilliant, thank you very much!

PR #361, opened 2022-02-22
DataTables

wavefrontshaping:
Add javascript support to dynamically filter the table result


alejandrogallo:
For context: this is how this looks on `elinks` (no javascript)

image

and when javascript is disabled (for instance in qutebrowser)

alejandrogallo:
I find you just got the best of both worlds, I agree this should be the default behaviour,
amazing work @wavefrontshaping in respecting also non-javascript workflows!

PR #368, opened 2022-03-18
fix export example in the readme

manuelhaussmann:
Unless I am missing something, this example is missing the `--format` part of the export.


alejandrogallo:
thank you manuel!

PR #374, opened 2022-04-16
Use configured default-library instead of default value

### LINKED ISSUE #249: The "papers" library has to be defined in configuration
When there is no `papers` library defined in the configuration, papis won't work and will raise the following exception.
```
File "/usr/lib/python3.8/site-packages/papis/config.py", line 477, in get_lib_from_name
    raise Exception(
Exception: Path or library 'papers' does not seem to exist
```
Another library is set as default and it works when the `papers` is defined in the configuration.

lennonhill:
Fix #249.

Can you think of any circumstances where this approach would fail?


alejandrogallo:
it seems ok to me!, thank you!

Issue #364, opened 2022-03-07
Adding Hash to document for namming folder strategy


reyman:
Hi,

I take some time to reread the code to manage folder name when adding document.

The `get_hash_folder(data: Dict[str, Any], document_paths: List[str]) -> str: )` seems called only when we are into [default strategy (code here)](https://github.com/papis/papis/blob/00baeeb891c6ff400a9bff4cc0c7754dd9b3e723/papis/commands/add.py#L306 )/

So it could be cool to add the hash directly as an attribute of doc to define something like that: 

`add-folder-name = {doc[author_list][0][family]} - {doc[year]} - {doc[hash]}`

alejandrogallo:
sure you could do this, but only for the building of the folder name or keep the hash forever.?

Related to this, one of the things that is left to be done in papis is creating a unique identifier for
every paper,this is kind of missing right now, that hash would be a good candidate probably,
but then everytime that a paper is added to the database has to be checked for having a hash and if not add one,
probably under the name `_papis_id` in the yaml or something like that.

I'm not yet sure, but this would certainly ease the work of things like implementing new database backends or
apps like the webapp and so on.

reyman:
Actually this is only to build the folder name but ...  you're right, a forever hash, accessible and stored into the yaml will probably simplify lot of things ! 

Into org-roam for example they use something like that, generating an unique ID PROPERTY (ex `:ID:       c4c543a1-fa0d-458a-80b1-ed87fdbe86d7`) for each note.

alexfikl:
We now have a `papis_id` in each document that acts as a sort of unique identifier. This seems to address the issue you had (even though it won't be the same as the hash generated by `get_hash_folder`).

Feel free to reopen if there are additional issues!

PR #378, opened 2022-05-04
Defauthash

reyman:
Hi,

A first proposal to think about hash by default. 
This is linked to #364 

Behavior i expect : 
- [X] random uuid is the default option 
- [X] empty `add-folder-name` is not authorized
- [X] doc[hash] is also available for `add-folder-name` option
- [ ] push hash for each files into yaml, stored into list
- [ ] test this hash to found duplication when adding document
- [ ] migration script

I'm searching the way to add a test the duplicate behavior, do you have an idea ?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 papis/deps/*
 .vim-run
+.idea/
 
 doc/build/
 #Vim swap files
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -213,17 +213,11 @@ def get_file_name(
 
     return filename
 
-
-def get_hash_folder(data: Dict[str, Any], document_paths: List[str]) -> str:
-    """Folder name where the document will be stored.
-
-    :data: Data parsed for the actual document
-    :document_paths: Path of the document
-
-    """
+def get_hash_for_file(data:Dict[str, Any], document_paths: List[str]) -> str:
+    """Unique Hash for document that will be stored.
+        :data: Data parsed for the actual document
+        """
     import random
-    author = "-{:.20}".format(data["author"])\
-             if "author" in data.keys() else ""
 
     document_strings = b''
     for docpath in document_paths:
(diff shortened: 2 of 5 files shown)


alejandrogallo:
I'll revisit this, we need a identifier for a document that never changes, that would be also useful for the nvim papis plugin and in general @jghauser @alexfikl what do you think?

alexfikl:
> I'll revisit this, we need a identifier for a document that never changes, that would be also useful for the nvim papis plugin and in general @jghauser @alexfikl what do you think?

Yeah, sure, sounds like a nice thing to have one, since we basically have a database. I'm not sure how it would look like though, hm..

jghauser:
Actually, this might indeed be useful for papis.nvim. I currently use the refs as unique identifiers, but there's a bug that I've been wanting to look at that happens when you change the ref of an existing entry. Luckily that's not something that people do a lot :sweat_smile:

alexfikl:
> The bad thing about this approach is that potentially just scanning by a database modifies your documents, which is a kind of behaviour that databases are not supposed to do, but that's ok in my book.

Maybe we can make that an error and just go like "Run `papis doctor --fix ids`" (or what the syntax actually is :)? So that it explicitly modifies the database.
 
> I propose to have the key to be called `_papis_id` and to have it in the `info.yaml` file as opposed as a separate file like `id.yaml` or something like this.

Yep, that sounds good! I don't feel strongly about it, but if we suffix it with `papis`, it could drop the leading underscore? Maybe we can document somewhere that keys starting with `papis` are reserved..

alejandrogallo:
yes to both points ;) Right now is `papis doctor -t papis-ids`, I have to document it before merging

alejandrogallo:
I close this since it is already implemented.

reyman:
Great to see this feature finished, sorry for the delay on that PR @alejandrogallo , 2022 was really busy ...

alejandrogallo:
don't mention it, we thank you anyways! we just needed it now for the plugins :)

PR #382, opened 2022-05-20
Template notes

gouderm:
I thought it would be more consistent to use the formater on the ``notes-name`` as well. This could for example be useful if one would like to name the notes similar to the title of the reference. Static notes-names can of course still be used (at least with the default ``python``-formater or ``jinja``-formater).

Also, I added a little more documentation for ``notes-name`` and ``notes-template``.

Finally, I forgot to import ``papis.format`` and fixed it herewith. However, I must admit that I don't understand why calling ``papis.format.format()`` works in the first place, without importing ``papis.format``. Is ``papis.format`` imported globally?

### DIFF
--- a/doc/source/default-settings.rst
+++ b/doc/source/default-settings.rst
@@ -333,14 +333,17 @@ Bibtex options
 
     In ``papis edit`` you can edit notes about the document. ``notes-name``
     is the default name of the notes file, which by default is supposed
-    to be a TeX file.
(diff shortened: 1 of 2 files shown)


alejandrogallo:
thank you, that looks nice and is a good idea.

About the `papis.format`, I am not sure, it is so used in so many places that it is possible that it stucked in the local python environment. flake8 should have complained about this, but oh well, it is not the end of the world ;)

Thanks so much!

PR #383, opened 2022-05-22
added option to delete notes-file of a document

gouderm:
Similar to `papis rm --file`, I added the option `-n`/`--notes` to delete the notes-file. Thereby, the info file of the document is updated accordingly. Deleting the notes-file works also with the options `--all` and `--force`. Furthermore, it's also possible to use both options `--file` and `--notes` simultaneously.

Besides that, I discovered and fixed a bug that occurred, when creating new notes for a document. Before the fix, the database would not update the document after adding a new notes-file. Therefore, one had to run `papis --cc` after creating new notes. I didn't find any open issue about that.

### DIFF
--- a/papis/commands/edit.py
+++ b/papis/commands/edit.py
@@ -74,10 +74,13 @@ def edit_notes(document: papis.document.Document,
     logger = logging.getLogger('edit:notes')
     logger.debug("Editing notes")
 
+    db = papis.database.get()
     if not document.has("notes"):
         notes_name = papis.config.getstring("notes-name")
         document["notes"] = papis.format.format(notes_name, document)
         document.save()
+        db.update(document)
+
     notes_path = os.path.join(
         str(document.get_main_folder()),
         document["notes"]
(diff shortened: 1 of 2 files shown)


gouderm:
Seeing that the Static Code Analysis failed, is there a way to run these tests locally or at least before opening a pull request? Unfortunately, `make test` already fails with the official master-branch, before making any changes.

alejandrogallo:
You can basically use
`./tests/ci-run-tests.sh`
to test your pull request, if that passes then it should be good to go. The Codacy static code analysis often complains for very minor things, I wouldn't be too much worried about this.

It is probably a good moment to update the doucmentation about testing and maybe to even remove that old makefile ... ;)

Your code make sense, I think I'll go ahead and merge it

alejandrogallo:
I updated the documentation for testing and fixed a small flake8 mistage in your changes

gouderm:
Perfect, that helps a lot. Thanks!

PR #384, opened 2022-05-26
adds jinja2 support for query string

gouderm:
Up until now, it was not possible to use query-strings with the jinja2-formater in the papis-database (= cache-based database). This should fix the `DocMatcher` which is being used in the papis-database. Honestly, I think this issue should be fixed in a more elegant way, since it will not work with formaters other than `python` and `jinja2` ([since custom formaters can also be imported as plugins I think](https://papis.readthedocs.io/en/latest/configuration.html#config-settings-formater)). Maybe you can come up with a better idea :)

I haven't tested the whoosh-database yet and am not sure if this is there also a problem.

### DIFF
--- a/papis/docmatcher.py
+++ b/papis/docmatcher.py
@@ -28,8 +28,21 @@ class DocMatcher(object):
     """
     search = ""  # type: str
     parsed_search = []  # type: List[Any]
-    doc_format = '{%s[DOC_KEY]}' % (papis.config.getstring('format-doc-name'))
     logger = logging.getLogger('DocMatcher')
+
+    formater = papis.config.getstring("formater")
+    if formater == "python":
+        doc_format = '{%s[DOC_KEY]}' % (
+            papis.config.getstring('format-doc-name'))
+
+    elif formater == "jinja2":
+        doc_format = '{{%s["DOC_KEY"]}}' % (
+            papis.config.getstring('format-doc-name'))
+
+    else:
+        doc_format = ''
+        logger.error(f"doc_format not implemented for {formater=}")
+
     matcher = None  # type: Optional[MATCHER_TYPE]
 
     @classmethod
--- a/papis/docmatcher.py
+++ b/papis/docmatcher.py
@@ -41,7 +41,7 @@ class DocMatcher(object):
 
     else:
         doc_format = ''
-        logger.error(f"doc_format not implemented for {formater=}")
+        logger.error("doc_format not implemented for formater='%s'", formater)
 
     matcher = None  # type: Optional[MATCHER_TYPE]
 


papis/docmatcher.py
```diff
@@ -28,8 +28,21 @@ class DocMatcher(object):
...
+        doc_format = '{%s[DOC_KEY]}' % (
+            papis.config.getstring('format-doc-name'))
+
+    elif formater == "jinja2":
+        doc_format = '{{%s["DOC_KEY"]}}' % (
+            papis.config.getstring('format-doc-name'))
+
+    else:
+        doc_format = ''
+        logger.error(f"doc_format not implemented for {formater=}")
```
alexfikl, line 44:
```suggestion
        logger.error(f"doc_format not implemented for '%s'", formater)
```
since the `{var=}` syntax is only available in python 3.8 (and papis supports >=3.5).

papis/docmatcher.py
```diff
@@ -28,8 +28,21 @@ class DocMatcher(object):
...
     logger = logging.getLogger('DocMatcher')
+
+    formater = papis.config.getstring("formater")
+    if formater == "python":
+        doc_format = '{%s[DOC_KEY]}' % (
+            papis.config.getstring('format-doc-name'))
+
+    elif formater == "jinja2":
+        doc_format = '{{%s["DOC_KEY"]}}' % (
+            papis.config.getstring('format-doc-name'))
```
alexfikl, line 40: Could this use `papis.format.format`? Ideally, we don't want to hardcode the available formatters and just use the plugin system somehow.

papis/docmatcher.py
gouderm: Thank you for the feedback, I wasn't aware of that. Fixed it in the new commit.

papis/docmatcher.py
gouderm: Indeed, it is probably a good idea to implement the `doc_format` variable directly in the `papis.format.Formater` themselves, rather than hard-coding it in the `DocMatcher`.

alexfikl (commented):
I'm afraid I don't have any clean suggestions to make, so we should probably wait for @alejandrogallo for a better view of how this should work. 

For this use case, it seems sufficient to set `sformat = doc[parsed[0]]` at
https://github.com/papis/papis/blob/87350624513fb9aef393bd70c49fd95076946816/papis/docmatcher.py#L69
and sidestep the formatters completely. But maybe a more general fix would be to add a method `Formater.make_document_format(key)` and let the formatter wrap the `doc` as it needs to.

alejandrogallo:
sorry guys it has been a while since I last did this.

I have a suggestion, which probably simplifies everything. I think the docmatcher is not super cleanly written, I guess I would do it differently now. Right now I do not see the need of having the `doc_format` at all

As far as I understand and remember the `doc_format` is there just to parse

```
author:einstein ⇒ {doc[author]}
```

and to be fed to the matcher which uses under the hood `papis.format` to complete the following diagram

```
                v-------- docmatcher
author:einstein ⇒ {doc[author]} ⇒ A. Einstein
                                ^------ papis.format
```

what I suggest is that we get rid of the middleman altogether, which would mean just to input to the matcher then `A. Einstein`, i.e.

```python
    @classmethod
    def return_if_match(
            cls,
            doc: papis.document.Document) -> Optional[papis.document.Document]:
        match = None
        is_keyword = False
        for parsed in cls.parsed_search:
            if len(parsed) == 1:
                search = parsed[0]
                is_keyword = False
            elif len(parsed) == 3:
                search = parsed[2]
                is_keyword = True

            if cls.matcher is not None:
                match = doc if cls.matcher(doc,
                                           search,
                                           str(doc[parsed[0]])
                                           if is_keyword else None) else None
            if not match:
                break
        return match
```

Sorry I don't have time right now to test it thoroughly with jinja2 maybe some of you can test it with jinja? I tested it with the normal formatter and it worked.

What do you think?

alexfikl:
> What do you think?

I don't use the `jinja` formatter either, but the suggestion looks good to me and the simplification is definitely welcome!

gouderm:
That solution makes a lot more sense. I wasn't sure if `doc_format` would be used in other parts too. Though if it is only used in the `return_if_match` method, then getting rid of `doc_format` should be the best option.

I have broadly tested it with `jinja` and from what I can tell, it is working perfectly well after that change. Thanks!

alexfikl:
Thanks for working on this! #499 fixed this issue slightly differently, so I'll go ahead and close this.
Feel free to open another PR if there's some missing functionality!

Issue #386, opened 2022-05-29
Do not prompt to add files from importer if user specifies file in `add` command


maxmahlke:
* papis version (`$ papis --version` or commit number):
papis, version 0.12

# how to reproduce the issue

I frequently download article PDFs behind paywalls. I then add them to my library
with `papis add /path/to/pdf` and get the metadata using `--from doi doi.of.article`.

In all cases, the metadata importer then retrieves an html page (the paywall) from the publisher, opens it in my browser, and asks to confirm whether I want to add this file as the article. I never want this.

IMHO the importer should not trigger the file retrieval (only add metadata) if the user passes the file to add explicitly via the command line:

    papis add /path/to/pdf --from doi doi.of.article  -> only get metadata
    papis add --from doi doi.of.article  -> get metadata and file

This change could be implemented by changing

https://github.com/papis/papis/blob/87350624513fb9aef393bd70c49fd95076946816/papis/commands/add.py#L557

to

    if importer.ctx.files and not files:

Instead of a PR, I'm creating an issue as this might not be a change you'd like to make to the importer.

maxmahlke:
Sounds good to me! Sane behaviour by default but not losing any functionality.

I opened PR #387 to add the flag.

PR #387, opened 2022-05-31
Add `force_download` flag to `add` command

maxmahlke:
Addresses #386 

As discussed, the importer is only used to download files if no local file is passed or the new `--force-download` flag is set.

It seems that the docs are automatically generated so I did not add anything there. Let me know if there is something missing.


alejandrogallo:
Looks good ! Thanks!

PR #395, opened 2022-08-01
More flexibility to put newly added entries in configurable subfolders.

f0rki:
* Added a ini config option named `add-subfolder` to set a default for `--subfolder` on `papis add`
* Made the config option `add-folder-name` s.t., it can create
  subfolders in the library based on document properties, e.g., grouping all papers from a year into subfolder with `{doc[year]}/{doc[title]}`.
* I switched to using `pathlib` from the python standard library for handling the paths instead of using the `os.path` functions.

### DIFF
--- a/config.py
+++ b/config.py
@@ -54,6 +54,7 @@ general_settings = {
 
     "add-confirm": False,
     "add-folder-name": "",
+    "add-subfolder": "",
     "add-file-name": None,
     "add-interactive": False,
     "add-edit": False,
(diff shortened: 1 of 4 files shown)


alejandrogallo:
I'm sorry for the delay, I'll review this today

alejandrogallo:
It looks really great, thank you for condensing everything tidy into one commit and taking the time to concisely document it, I think this is useful for many people.

Cheers!

PR #396, opened 2022-08-14
adds -p/--pick-subfolder option to `papis add`

gouderm:
This pull-request adds the option `-p`/`--pick-subfolder` to the `papis add` command. When using this flag, then the configured picker is called to choose a parent-subfolder. This option can be used together with the `-d`/`--subfolder` flag, to create a new subfolder inside the picked subfolder.

The list of subfolders to pick from is defined by iterating over all documents in the library, setting the parent folder of each document as a subfolder. This leads to the problem of not finding empty subfolders (folders that do not containing any documents) in the library, though this might be also the expected/logical behavior?

### DIFF
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -250,6 +250,7 @@ def run(paths: List[str],
         folder_name: Optional[str] = None,
         file_name: Optional[str] = None,
         subfolder: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


alejandrogallo:
This looks good, the empty folder behaviour seems right to me, let's go for now with it!
Thanks!f

PR #402, opened 2022-08-31
add: Remove unused out_folder_name variable

alexfikl:
It seems like the `folder_name is None` case was broken when `out_folder_path` was introduced.

This should hopefully fix that, but I haven't tested it :grimacing:

### DIFF
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -286,8 +286,6 @@ def run(paths: List[str],
     import tempfile
     logger = logging.getLogger('add:run')
 
-    # The folder name of the new document that will be created
-    out_folder_name = None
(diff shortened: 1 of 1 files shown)


papis/commands/add.py
```diff
@@ -315,6 +313,7 @@ def run(paths: List[str],
 
     if not folder_name:
         out_folder_name = get_hash_folder(data, in_documents_paths)
+        out_folder_path = out_folder_path / out_folder_name
```
alexfikl, line 316: `out_folder_name` isn't used anywhere after this point anymore.

alejandrogallo: Good one!

PR #405, opened 2022-10-07
docs: add papis.nvim plugin

jghauser:
I've recently published the first iteration of a [neovim plugin for papis](https://github.com/jghauser/papis.nvim). While definitely still work-in-progress, it's probably quite a bit more advanced than the existing vim plugin and should be stable enough for most. I'm using it myself and am committed to further updating and improving it.


alejandrogallo:
This looks awesome, thank you very much as always.
I would consider deprecating the `papis-vim` stuff in favour of your approach, which just seems much better.

alejandrogallo:
This looks awesome, thank you very much as always.
I would consider deprecating the `papis-vim` stuff in favour of your approach, which just seems much better.

jghauser:
I do think my approach is superior, and it doesn't really seem like the vim plugin is actively maintained. One disadvantage of my plugin is that it won't work on vim, and you might want to keep the vim plugin for users who don't (want to) use neovim.

PR #410, opened 2022-10-22
Fix doc generation

alexfikl:
Running with
```
make html SPHINXOPTS="-W --keep-going -n"
```
showed quite a few errors, which this should fix.

### DIFF
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -1,7 +1,5 @@
-
 API
 ===
-
 
 .. automodule:: papis.api
   :members:
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -45,9 +45,14 @@ extensions = [
     'sphinx.ext.coverage',
     'sphinx.ext.ifconfig',
     'sphinx.ext.viewcode',
+    'sphinx.ext.intersphinx',
     'sphinx.ext.inheritance_diagram',
     'sphinx_click.ext',
 ]
+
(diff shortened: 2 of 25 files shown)


alexfikl:
The online docs are current not quite right either. e.g. 
* commands: https://papis.readthedocs.io/en/latest/commands.html
* the default value for `status_line_format`: https://papis.readthedocs.io/en/latest/configuration.html#terminal-user-interface-picker

and probably others. Haven't checked everything, but this should improve it a bit.

jghauser:
I was actually just starting to write down some thoughts on improving the docs. I think a more general overhaul would be useful (rewriting sections, re-organising sections, etc.). I'll start a discussion thread in the next days once I've thought about it a little. Just wanted to give you a heads up (not that it clashes with this PR in any way!).

alexfikl:
> I was actually just starting to write down some thoughts on improving the docs. I think a more general overhaul would be useful (rewriting sections, re-organising sections, etc.). I'll start a discussion thread in the next days once I've thought about it a little. Just wanted to give you a heads up (not that it clashes with this PR in any way!).

That sounds great! I was having some similar thoughts while jumping around all the sections and fixing these small issues, so I'm happy to help out :rocket:

alejandrogallo:
That would be great, there are some things that are also outdated, which would be super annoying for new users.
Please Alex feel free to merge this whenever you see fit.

Issue #407, opened 2022-10-20
Faulty bibtex export


jghauser:
This is on the latest git commit using the arch `papis-git` package.

# how to reproduce the issue
`papis export --all --format bibtex --out ~/Documents/library/main.bib '*' `

When I export my library as a bibtex file, I get a bunch of entries like this:
```
@article{,
}
```
And these in turn make pandoc fail when generating pdfs. I know they probably stem from erronous library entries. Is there some way of finding out what entries those would be? Also, I think even in the case of such erronous entries, it would be better to simply suppress them (or warn the user) rather than silently creating a faulty `.bib` file.

alexfikl:
Hm, there doesn't seem to be a clean way to get the entries that look empty. One thing you can try with current `master` is running
```
papis --log DEBUG --logfile main.log export [other arguments]
```
and look for instances of `'ref = '` (with quotes because of [this](https://github.com/papis/papis/blob/5ec7d44ea6f9ec472fcd4e0e925eb7e16b00643d/papis/bibtex.py#L240)), i.e. where no `ref` is found. The next lines should tell you a bit about the document, at least the folder it's in.

Agreed that this should fail a bit more gracefully though!

jghauser:
That was very helpful, thanks! I realised there were some double entries in the log of which some had `'ref = '` and doing a `papis --clear-cache` got rid of them and fixed the export. I'm gonna leave the issue up and maybe come up with something to warn users about missing `ref`s.

alejandrogallo:
Hmm this is important, it could be a workflow breaker for fsome people,
maybe something like

```sh
papis doctor
```

would be in order for this kind of situations. A command where the user
can just generally asses the state of the library, look for missing references,
look for missing `files` etc, if we want to do this maybe we can make a list of
what does it mean to have a sane library in terms of `yaml` files and so on

PR #412, opened 2022-10-23
Better error handling in to_bibtex

alexfikl:
Improves the handling by
* skipping the document if no `bibtex_type` can be determined
* skipping the document if no `ref` can be determined
* adding some error messages to see which documents are causing problems

@jghauser Does this help with #407? It should at least skip documents with no `ref` like the one in your example. I tried modifying some files in my library, but don't have anything substantial to test with.

### DIFF
--- a/papis/bibtex.py
+++ b/papis/bibtex.py
@@ -1,7 +1,7 @@
 import os
 import string
 import logging
-from typing import Optional, List, Dict, Any
+from typing import Optional, List, Dict, Any, Iterator
 
 import click
 
@@ -52,7 +52,7 @@ bibtex_key_converter = {
 
 
 def exporter(documents: List[papis.document.Document]) -> str:
-    return "\n".join(to_bibtex(document) for document in documents)
+    return "\n".join(to_bibtex_multiple(documents))
 
 
 class Importer(papis.importer.Importer):
@@ -210,6 +210,17 @@ def create_reference(doc: Dict[str, Any]) -> str:
     return ref_cleanup(ref)
 
 
+def to_bibtex_multiple(documents: List[papis.document.Document]) -> Iterator[str]:
+    for doc in documents:
+        bib = to_bibtex(doc)
+        if not bib:
+            logger.warning("Skipping document export: '%s'",
+                           doc.get_info_file())
+            continue
+
+        yield bib
+
+
 def to_bibtex(document: papis.document.Document) -> str:
     """Create a bibtex string from document's information
 
@@ -220,16 +231,29 @@ def to_bibtex(document: papis.document.Document) -> str:
     bibtex_string = ""
     bibtex_type = ""
 
-    # First the type, article ....
+    # determine bibtex type
     if "type" in document:
         if document["type"] in bibtex_types:
             bibtex_type = document["type"]
         elif document["type"] in bibtex_type_converter:
             bibtex_type = bibtex_type_converter[document["type"]]
+        else:
+            logger.error("BibTeX type '%s' not valid: '%s'",
+                         document["type"],
+                         document.get_info_file())
+            return ""
+
     if not bibtex_type:
         bibtex_type = "article"
 
+    # determine ref value
     ref = create_reference(document)
+    if not ref:
+        logger.error("No valid ref found for document: '%s'",
+                     document.get_info_file())
+
+        return ""
+
     logger.debug("Used 'ref = %s'", ref)
 
     bibtex_string += "@{type}{{{ref},\n".format(type=bibtex_type, ref=ref)
--- a/tests/test_bibtex.py
+++ b/tests/test_bibtex.py
@@ -80,3 +80,60 @@ def test_clean_ref() -> None:
             (r"Albert_Ein\_stein\.1923.b", "AlbertEin_stein.1923B"),
             ]:
         assert rc == papis.bibtex.ref_cleanup(r)
+
+
+def test_to_bibtex_wrong_type() -> None:
+    """Test no BibTeX entry is constructed for incorrect types."""
+
+    doc = papis.document.from_data({
+        "type": "fictional",
+        "ref": "MyDocument",
+        "author": "Albert Einstein",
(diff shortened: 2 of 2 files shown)


jghauser:
Looks great, I'll check it out! Though [as I said in the other thread](https://github.com/papis/papis/issues/407#issuecomment-1285713537) I don't think the main cause was missing ref keys but a messed up database (leading to missing ref keys in a roundabout way). And I don't know how to replicate that :sweat_smile:. I can however test it by messing with some entries like you have.

alexfikl:
> Looks great, I'll check it out! Though [as I said in the other thread](https://github.com/papis/papis/issues/407#issuecomment-1285713537) I don't think the main cause was missing ref keys but a messed up database (leading to missing ref keys in a roundabout way). And I don't know how to replicate that. I can however test it by messing with some entries like you have.

Ah, right! Well, at least this should give some better error message to point to the problematic documents. 

No pressure on testing it then! I'll leave this as a draft until #403 is in and add some unit tests in there for simple cases.

alexfikl:
@alejandrogallo @jghauser This should be ready for a look now. Added some test cases and some more explicit messages for when `to_bibtex` fails somehow, so hopefully it improves on the status quo. Let me know if you spot any improvements!

jghauser:
Did a quick test of this and the output is very useful. Apparently there are a few entries that are faulty in my library that I never knew about!
However, I think because I use jinja2 formatting, I can't actually see what the entries are. Here's how my output looks:
```
ERROR:document:bibtex:BibTeX type 'SSRN Working Paper' not supported: '{doc[title]} - {doc[author]}'
WARNING:bibtex:Could not export doc to bibtex: '{doc[title]} - {doc[author]}'
ERROR:document:bibtex:BibTeX type 'SSRN Scholarly Paper' not supported: '{doc[title]} - {doc[author]}'
WARNING:bibtex:Could not export doc to bibtex: '{doc[title]} - {doc[author]}'
```
I also think giving the user the paths to the `info.yaml` might be more effective. Then fixing it just involves copying the path rather than querying the library. Though maybe that's just my particular workflow...

alexfikl:
> ERROR:document:bibtex:BibTeX type 'SSRN Working Paper' not supported: '{doc[title]} - {doc[author]}'
> WARNING:bibtex:Could not export doc to bibtex: '{doc[title]} - {doc[author]}'
> ERROR:document:bibtex:BibTeX type 'SSRN Scholarly Paper' not supported: '{doc[title]} - {doc[author]}'
> WARNING:bibtex:Could not export doc to bibtex: '{doc[title]} - {doc[author]}'
> ```

Hm, that's quite unfortunate. This uses `papis.document.describe` should use the formatter. Did you set `document-description-format` to something `jinja2` friendly?

> I also think giving the user the paths to the `info.yaml` might be more effective. Then fixing it just involves copying the path rather than querying the library. Though maybe that's just my particular workflow...

Yeah, that's a good point! The whole thing can be controlled by the user through `document-description-format`, so not sure if it's worth hardcoding it to the `info.yml` file. I'll give it a try and see!

jghauser:
Ok, my bad, never even heard of that config option (I'll add it to things to be documented :wink:)

alexfikl:
> Ok, my bad, never even heard of that config option (I'll add it to things to be documented wink)

Sounds good! We should also do a bit better with the defaults in https://github.com/papis/papis/blob/dd57720c09b480b0fef08a4c13c3d4e0a5e2cc8d/papis/config.py. They're all geared towards the python formatter, so if you use another formatter you need to set all of them.. which is not ideal :cry:

alexfikl:
> I also think giving the user the paths to the `info.yaml` might be more effective. Then fixing it just involves copying the path rather than querying the library. Though maybe that's just my particular workflow...

Should be printing out the full path to the `info.yml` file now.

jghauser:
> Sounds good! We should also do a bit better with the defaults in https://github.com/papis/papis/blob/dd57720c09b480b0fef08a4c13c3d4e0a5e2cc8d/papis/config.py. They're all geared towards the python formatter, so if you use another formatter you need to set all of them.. which is not ideal cry

Yeah, I agree. I think it would be good to have some defaults for jinja2 as well. I have a pretty complete (but apparently not quite :sweat_smile:) config with jinja2, so I can provide some defaults when we get to that.

The error handling now works well, I like how it shows the path (and thus also works fine with jinja2 even without the `document-description-format` option). Only thing that might be a bit confusing is that each failure generates a `ERROR` and a `WARNING`. First, this might seem to imply that there's two failures where there are really just one, and second, it's a bit odd that two messages with different degrees are generated by one and the same failure.

alexfikl:
> Only thing that might be a bit confusing is that each failure generates a `ERROR` and a `WARNING`. First, this might seem to imply that there's two failures where there are really just one, and second, it's a bit odd that two messages with different degrees are generated by one and the same failure.

Hm, not quite sure on what side to err on this. On one hand, you have a good point that it's odd, but it doesn't bother me much because one message tells you what's wrong and the other one just informs you the entry was skipped in the export, so they're not too repetitive (could be phrased better?). We could make the warning a `logger.debug` instead so it only shows if someone really wants to see it?

jghauser:
I think you're right. Maybe best to leave it as is. They are technically speaking two different kinds of error, and it's definitely clear from the context what is going on. 

Maybe the text of the warning could be "Skipping export of document: ..." just to make clear what's going on. Also the `ERROR` message might be "BibTeX type 'X' not valid" (just to make it clear that this isn't something that Papis happens to not have implemented).

alexfikl:
> Maybe the text of the warning could be "Skipping export of document: ..." just to make clear what's going on. Also the `ERROR` message might be "BibTeX type 'X' not valid" (just to make it clear that this isn't something that Papis happens to not have implemented).

Done! https://github.com/papis/papis/compare/8da353249d0916d5c408c7cd51e222c3e952b115..e8f3cb2c8e1bb732c7d0ccf6c1328c1603e5e124

PR #414, opened 2022-10-25
Update readthedocs config to version 2

alexfikl:
The documentation for the new config format is at
https://docs.readthedocs.io/en/stable/config-file/v2.html#python-install

As far as I can tell, this is correct, but I obviously can't test it :(
I haven't managed to get these configs right from the first try, so it's likely it'll need some fiddling after merging..

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,5 +1,17 @@
+version: 2
+
+sphinx:
+  configuration: docs/source/conf.py
+
+build:
+  os: ubuntu-20.04
+  tools:
+    python: "3.10"
+
 python:
-  version: 3.5
-  pip_install: true
-  extra_requirements:
-    - develop
+  install:
+    - method: pip
+      path: .
+      extra_requirements:
+        - develop
+  system_packages: true


.readthedocs.yml
```diff
@@ -1,5 +1,17 @@
+version: 2
+
+sphinx:
+  configuration: docs/source/conf.py
+
+build:
+  os: ubuntu-20.04
+  tools:
+    python: "3.10"
```
alexfikl, line 9: This was previously set to python 3.5 (i.e. oldest version supported) and this sets it to 3.10 (i.e. newest version supported). 

Any preference here? I would say latest version is better, since it has newer versions of packages, especially `sphinx` itself.

alejandrogallo: I agree, for documentation let's stick with what it works!

alejandrogallo:
Yeah, since we have the CI failing anyways let's merge it and see if this works.
If not then maybe we can host the docs anyways ourselves through gh-pages, I normally do that.

alexfikl:
> Yeah, since we have the CI failing anyways let's merge it and see if this works. If not then maybe we can host the docs anyways ourselves through gh-pages, I normally do that.

Found a little typo in #418 but it all seems to be working now, i.e. all the commands show up and everything on
https://papis.readthedocs.io/en/latest/commands.html

Thanks for taking a look!

PR #421, opened 2022-11-20
Add first implementation of papis-doctor

alejandrogallo:
As we discussed some time ago, that would be nice to have a `papis doctor`
functionality.o

I have prepared a first implementation of a candidate.
In principle one implement different tests
and then queries the library like on all other commands.

Right now there are two tests implemented,
one that looks for missing keys and another one that
checks if the `files` of the documents actually exists.
Here is an output from my library.

```sh
papis doctor --all
WARNING:doctor:12 errors found
keys author /home/gallo/Documents/papers/3d7c4e4778f26e1c66684a11f604b9f4
keys author /home/gallo/Documents/papers/b30815699c8c34bf2fdd472cdc1f56c9
keys author /home/gallo/Documents/papers/5296700afbfb46da3e9361774ad9440a
keys author /home/gallo/Documents/papers/c666475c4b79a6216aa04e1914e72f15
keys author /home/gallo/Documents/papers/c1d2e472a175482dfa30670b30eb583f
keys author /home/gallo/Documents/papers/56c9ea78eb34406e61a52631c09f992f
files /home/gallo/Documents/papers/e31d853be4f5f11162f3c8b2abb13a2c-bassani-f.-and-inch/position-of-the-bassani-f.-and-inchauspe-n..pdf /home/gallo/Documents/papers/e31d853be4f5f11162f3c8b2abb13a2c-bassani-f.-and-inch
keys ref /home/gallo/Documents/papers/5df2864bdb8eb62891a8dc229eec1032-kaldor-uzi
keys title /home/gallo/Documents/papers/05e50e4ea461ea527851f947086c9d5b
keys author /home/gallo/Documents/papers/05e50e4ea461ea527851f947086c9d5b
keys ref /home/gallo/Documents/papers/05e50e4ea461ea527851f947086c9d5b
files /home/gallo/Documents/papers/1db30a88d9ad08ebbf76ee84bcc295f6-pederson-mark-r.-an/improved-theoretical-methods-for-studies-of-defects-in-insulators-application-to-the-pederson-mark-r.-and-klein-barry-m.pdf /home/gallo/Documents/papers/1db30a88d9ad08ebbf76ee84bcc295f6-pederson-mark-r.-an
```

For library writers I've also implemented a `--json` exporter

```sh
>> papis doctor --all --json | jq
```

```json
[
  {
    "msg": "author",
    "path": "/home/gallo/Documents/papers/3d7c4e4778f26e1c66684a11f604b9f4",
    "name": "keys",
    "suggestion": "\n                 papis update --doc-folder /home/gallo/Documents/papers/3d7c4e4778f26e1c66684a11f604b9f4\n                 "
  },
  {
    "msg": "author",
    "path": "/home/gallo/Documents/papers/b30815699c8c34bf2fdd472cdc1f56c9",
    "name": "keys",
    "suggestion": "\n                 papis update --doc-folder /home/gallo/Documents/papers/b30815699c8c34bf2fdd472cdc1f56c9\n                 "
  },
  {
    "msg": "author",
    "path": "/home/gallo/Documents/papers/5296700afbfb46da3e9361774ad9440a",
    "name": "keys",
    "suggestion": "\n                 papis update --doc-folder /home/gallo/Documents/papers/5296700afbfb46da3e9361774ad9440a\n                 "
  },
  {
    "msg": "author",
    "path": "/home/gallo/Documents/papers/c666475c4b79a6216aa04e1914e72f15",
    "name": "keys",
    "suggestion": "\n                 papis update --doc-folder /home/gallo/Documents/papers/c666475c4b79a6216aa04e1914e72f15\n                 "
  },
  {
    "msg": "author",
    "path": "/home/gallo/Documents/papers/c1d2e472a175482dfa30670b30eb583f",
    "name": "keys",
    "suggestion": "\n                 papis update --doc-folder /home/gallo/Documents/papers/c1d2e472a175482dfa30670b30eb583f\n                 "
  },
  {
    "msg": "author",
    "path": "/home/gallo/Documents/papers/56c9ea78eb34406e61a52631c09f992f",
    "name": "keys",
    "suggestion": "\n                 papis update --doc-folder /home/gallo/Documents/papers/56c9ea78eb34406e61a52631c09f992f\n                 "
  },
  {
    "msg": "/home/gallo/Documents/papers/e31d853be4f5f11162f3c8b2abb13a2c-bassani-f.-and-inch/position-of-the-bassani-f.-and-inchauspe-n..pdf",
    "path": "/home/gallo/Documents/papers/e31d853be4f5f11162f3c8b2abb13a2c-bassani-f.-and-inch",
    "name": "files",
    "suggestion": "\n                  papis edit --doc-folder /home/gallo/Documents/papers/e31d853be4f5f11162f3c8b2abb13a2c-bassani-f.-and-inch\n                  "
  },
  {
    "msg": "ref",
    "path": "/home/gallo/Documents/papers/5df2864bdb8eb62891a8dc229eec1032-kaldor-uzi",
    "name": "keys",
    "suggestion": "\n                 papis update --doc-folder /home/gallo/Documents/papers/5df2864bdb8eb62891a8dc229eec1032-kaldor-uzi\n                 "
  },
  {
    "msg": "title",
    "path": "/home/gallo/Documents/papers/05e50e4ea461ea527851f947086c9d5b",
    "name": "keys",
    "suggestion": "\n                 papis update --doc-folder /home/gallo/Documents/papers/05e50e4ea461ea527851f947086c9d5b\n                 "
  },
  {
    "msg": "author",
    "path": "/home/gallo/Documents/papers/05e50e4ea461ea527851f947086c9d5b",
    "name": "keys",
    "suggestion": "\n                 papis update --doc-folder /home/gallo/Documents/papers/05e50e4ea461ea527851f947086c9d5b\n                 "
  },
  {
    "msg": "ref",
    "path": "/home/gallo/Documents/papers/05e50e4ea461ea527851f947086c9d5b",
    "name": "keys",
    "suggestion": "\n                 papis update --doc-folder /home/gallo/Documents/papers/05e50e4ea461ea527851f947086c9d5b\n                 "
  },
  {
    "msg": "/home/gallo/Documents/papers/1db30a88d9ad08ebbf76ee84bcc295f6-pederson-mark-r.-an/improved-theoretical-methods-for-studies-of-defects-in-insulators-application-to-the-pederson-mark-r.-and-klein-barry-m.pdf",
    "path": "/home/gallo/Documents/papers/1db30a88d9ad08ebbf76ee84bcc295f6-pederson-mark-r.-an",
    "name": "files",
    "suggestion": "\n                  papis edit --doc-folder /home/gallo/Documents/papers/1db30a88d9ad08ebbf76ee84bcc295f6-pederson-mark-r.-an\n                  "
  }
]
```

Also as you see in the json output, I thought maybe it's good
to have a suggestion function per test checked,
in text output there is the `--suggest` option

```sh
> papis doctor --suggest --all

...

keys ref /home/gallo/Documents/papers/05e50e4ea461ea527851f947086c9d5b
Suggestion:

                 papis update --doc-folder /home/gallo/Documents/papers/05e50e4ea461ea527851f947086c9d5b
                 

files /home/gallo/Documents/papers/1db30a88d9ad08ebbf76ee84bcc295f6-pederson-mark-r.-an/improved-theoretical-methods-for-studies-of-defects-in-insulators-application-to-the-pederson-mark-r.-and-klein-barry-m.pdf /home/gallo/Documents/papers/1db30a88d9ad08ebbf76ee84bcc295f6-pederson-mark-r.-an
Suggestion:

                  papis edit --doc-folder /home/gallo/Documents/papers/1db30a88d9ad08ebbf76ee84bcc295f6-pederson-mark-r.-an
```

I would say there is now the time to discuss what would be the
intended usage and the ergonomic use of the command.

Also, the code is quite straightforwardly implemented I think.


papis/commands/doctor.py
```diff
@@ -0,0 +1,169 @@
...
+
+def run(doc: papis.document.Document, tests: List[str]) -> List[Error]:
+    """
+    Runner for doctor. It runs all the tests given by the test
+    argument, and it gets the test from the global REGISTERED_TESTS
+    dictionary.
+    """
+    results = []  # type: List[Error]
+    for test in tests:
+        results += REGISTERED_TESTS[test].operate(doc)
```
alexfikl, line 103:
```suggestion
        results.extend(REGISTERED_TESTS[test].operate(doc))
```
? Not sure if there's any actual difference? (Ignore if there isn't)

papis/commands/doctor.py
```diff
@@ -0,0 +1,169 @@
...
+
+@click.command("doctor")
+@click.help_option("--help", "-h")
+@papis.cli.query_option()
+@papis.cli.sort_option()
+@click.option("-t", "--tests", "_tests",
+              default=lambda: papis.config.getlist("doctor-default-tests"),
+              multiple=True,
+              help=("Tests to run on every document, possible values: {}"
+                    .format(REGISTERED_TESTS)))
```
alexfikl, line 115:
```suggestion
                    .format(", ".join(t.name for t in REGISTERED_TESTS)))
```
? The helps looks very heavy with the whole object.

alexfikl (approved):
Left a couple of nitpicks, but overall this works for me as advertised! Even found a document without a title :tada: 

The only higher level suggestion I would have is to call these things `checks` instead of `tests`? i.e. `REGISTERED_CHECKS`, etc. Sounds a bit less CS-y maybe..

papis/commands/doctor.py
alejandrogallo: it's the same, but I'll take it because it is semantically clearer ;)

alejandrogallo:
Yes I agree, i changed the names to check in all related parts,
test is a very loaded word.
So you think that the workflow is acceptable? I'm not sure this is the most ergonomic.
I would also like to ask if someone has a couple of more ideas for useful checks,
like making a check about maybe repeated `ref`s in the library and so on?

jghauser (commented):
This is definitely useful. I only have some general remarks:
- the author check would be more useful if it weren't triggered when an editor is defined instead (I get around 100 files simply because they are collections with just an editor defined)
- maybe the message could be "keys X missing" just to be more explicit
- might be cool to offer going through the list of errors one by one and offering to e.g. ignore/edit info.yaml
- might also be cool to somehow do a check automatically when adding new entries to the library (not quite sure how or at what stage though)
- i think checking for non-unique refs would be useful!

alejandrogallo:
- I added a duplicated-keys check, which does the duplication testing, I'm still not really 
  convinced about the erogonomicity of the whole, but I think it's a good start.
- About the going one by one, I agree, maybe I can add an `-i`, `--interactive` flag
  to then do an interactive action in every relevant case.
- About the checking automatically everytime one adds, I kind of agree, however I'm also not
  sure how.
- About the editor, yes, you're right, but I don't know how to do it generally,
  I could create a test that does this ad-hoc, and maybe write it in the documentation
  as an example of how people could write this in theyr `config.py` file to register a check.

thank you for the suggestions!

alejandrogallo:
that's a great idea, I'll implement the check

alejandrogallo:
- I can create an all-caps check where the user gives the keys where
  she wants to detect the caps. However it's difficult to have a fix
  action in general for any kind of string, because you want the
  Author Names capitalized like in `Bin Mahmoud, Hasan` and so on.
- A check that checks if there are html codes in the titles, like
  &amp; and so on.  In this case the auto fix would be to replace the
  html code by their unicode equivalents.

jghauser:
> I can create an all-caps check where the user gives the keys where
she wants to detect the caps. However it's difficult to have a fix
action in general for any kind of string, because you want the
Author Names capitalized like in Bin Mahmoud, Hasan and so on.

Yeah, that is true! I guess we could offer three options for converting strings: 1) title case, 2) sentence case, 3) custom/manual. I'm thinking that the first 2 options would cover more than 90% of the cases. Though I'm not sure if that would complicate things more than it would help.

> A check that checks if there are html codes in the titles, like
& and so on. In this case the auto fix would be to replace the
html code by their unicode equivalents.

Perfect! <3

alejandrogallo:
I'd like to merge this soonish because I'm starting to
implement small things in the webapp out of lazyness to do another branch,
but I think that's ok

image

papis/commands/doctor.py
```diff
@@ -0,0 +1,374 @@
...
+
+
+def refs_check(doc: papis.document.Document) -> List[Error]:
+    """
+    It checks that a ref exists and if not it
+    tries to create one according to the ref-format configuration
+    if the user chooses to fix it.
+    """
+    folder = doc.get_main_folder()
+    bad_symbols = re.compile(r"[ ,{}\[\]@#`']")
```
alexfikl, line 137: Is there some documentation for what characters are allowed (by bibtex?)? Didn't find anything from a short search..

papis/commands/doctor.py
```diff
@@ -0,0 +1,374 @@
...
+                                      .format(folder)),
+                      fix_action=_fix,
+                      payload="",
+                      doc=doc)]
+    m = bad_symbols.findall(str(doc["ref"]))
+    if m:
+        return [Error(name=REFS_CHECK_NAME,
+                      path=folder or "",
+                      msg=("Bad characters ({}) found in reference."
+                           .format(set(list(m)))),
```
alexfikl, line 163:
```suggestion
                           .format(set(m))),
```
? Does that not work?

papis/commands/doctor.py
```diff
@@ -0,0 +1,374 @@
...
+
+def duplicated_keys_check(doc: papis.document.Document) -> List[Error]:
+    """
+    Check for duplicated keys in `doctor-duplicated-keys-check`
+    """
+    keys = papis.config.getlist("doctor-duplicated-keys-keys")
+    folder = doc.get_main_folder()
+    results = []  # type: List[Error]
+    for key in keys:
+        if doc[key] in DUPLICATED_KEYS_SEEN[key]:
```
alexfikl, line 185:
```suggestion
        if str(doc[key]) in DUPLICATED_KEYS_SEEN[key]:
```
? Seeing as the append below only uses strings?

papis/commands/doctor.py
```diff
@@ -0,0 +1,374 @@
...
+        return __fix
+
+    for key in papis.config.getlist("doctor-html-codes-keys"):
+        if doc[key]:
+            m = HTML_CODE_REGEX.findall(str(doc[key]))
+            if m:
+                results.append(Error(name=HTML_CODES_CHECK_NAME,
+                                     path=folder or "",
+                                     msg=("Field '{}' contains html codes '{}'"
+                                          .format(key, m)),
```
alexfikl, line 256:
```suggestion
                                     msg=("Field '{}' contains html codes {}"
                                          .format(key, m)),
```
? `m` should already be a list of quoted strings, so the extra quote may look weird..

papis/commands/doctor.py
```diff
@@ -0,0 +1,374 @@
...
+
+def run(doc: papis.document.Document, checks: List[str]) -> List[Error]:
+    """
+    Runner for doctor. It runs all the checks given by the check
+    argument, and it gets the check from the global REGISTERED_CHECKS
+    dictionary.
+    """
+    results = []  # type: List[Error]
+    for check in checks:
+        results.extend(REGISTERED_CHECKS[check].operate(doc))
```
alexfikl, line 283: Ha.. doctor.. operate.. nice! :tada:

papis/commands/doctor.py
```diff
@@ -0,0 +1,374 @@
...
+
+@click.command("doctor")
+@click.help_option("--help", "-h")
+@papis.cli.query_option()
+@papis.cli.sort_option()
+@click.option("-t", "--checks", "_checks",
+              default=lambda: papis.config.getlist("doctor-default-checks"),
+              multiple=True,
+              help=("Checks to run on every document, possible values: {}"
+                    .format(", ".join(registered_checks_names()))))
```
alexfikl, line 295: Do a `click.Choice` for the check names?

papis/commands/doctor.py
```diff
@@ -0,0 +1,374 @@
...
+                  .format(error.msg))
+        if suggest:
+            print("\tSuggestion: {}"
+                  .format(error.suggestion_cmd))
+        if fix:
+            logger.warning("Fixing...")
+            error.fix_action()
+        if edit:
+            input("Press any key to edit...")
+            if error.doc:
```
alexfikl, line 373:
```suggestion
            if error.doc:
            	input("Press any key to edit...")
```
? Or add a message that it got skipped? Seems weird to press a key to edit and nothing happens

alexfikl (approved):
Left a few more nitpicky comments, but overall this looks super useful to me! Definitely eager to see it merged as well!

papis/commands/doctor.py
alejandrogallo: that is a good question, I also don't know for sure, but these are not kosher

papis/commands/doctor.py
alejandrogallo: yes this works wonderfully thank you!

papis/commands/doctor.py
alejandrogallo: agreed!

papis/commands/doctor.py
alejandrogallo: I only made doc to be optional to be kind of future-proof maybe with the explorer of documents, so practically there must be always a document, but you're right, like this is better.

papis/commands/doctor.py
jghauser: I remember trying to find out about those characters for my neovim plugin, and the best i could find is [this stackexchange thread](https://tex.stackexchange.com/questions/581901/what-is-the-safe-character-set-for-a-bibtex-label?noredirect=1&lq=1). It seems there isn't a clear set of characters, and what works depends on what software the `.bib` files are processed with.

alexfikl: Yeah, that seems to be the best explanation. From a bit of digging
* `bibtexparser` (which we're using) seems to accept anything besides a space at https://github.com/sciunto-org/python-bibtexparser/blob/19e8dfe05490f2fb8ea1a9cc593090400eaea42a/bibtexparser/bibtexexpression.py#L160
* `btparse` (which is used by `biber` and `biblatex` and probably others) has an allowed list of characters at https://metacpan.org/dist/Text-BibTeX/view/btparse/doc/btparse.pod (under *name*)

But like that stackexchange post says, different languages may expand some non-alphanumeric characters and lead to interesting results when actually putting things into a `tex` file..

alexfikl: Either way, the current choice is reasonable, so I'm perfectly fine with leaving this as is!

papis/commands/doctor.py
alejandrogallo: Yes that is a good idea

alejandrogallo:
I added katex, it's becoming the very thing I tried to avoid with so much javascript lying around, but I think katex is really unavoidable
image
image

alejandrogallo:
add optional timeline 
image

alejandrogallo:
- IDEA: Implement a check for fields that should be an integer, like month, year and so on.

alejandrogallo:
@alexfikl do you know if pytest is broken with python3.5? I think I messed something up in the setup.py since I added the pytest to the development requirements, and sicne then py3.5 does not work because pytest is not getting installed by pip

alexfikl:
> @alexfikl do you know if pytest is broken with python3.5? I think I messed something up in the setup.py since I added the pytest to the development requirements, and sicne then py3.5 does not work because pytest is not getting installed by pip,

Hm, it should still work since other recent PRs like #438 worked just fine. I looked at your changes to `setup.py` and it looks like `pytest` should be pulled in by `pytest-cov` like before.. so not really sure what it doesn't like :( but it is [adeb452](https://github.com/papis/papis/pull/421/commits/adeb452666ebec36b48328faa72d3d6d7b8b64c6) that started it somehow

Weird that `pip` doesn't throw any sort of errors or anything.. it just doesn't install that dependency. Maybe a bug in `pip` somehow? Can you add `pytest` explicitly to `setup.py`?

alejandrogallo:
I'll merge this behemoth because I want to start with the `papis_id`, there are still a couple of things to do here &ut everything is usable and I will clean them up for `v0.13`

PR #424, opened 2022-11-23
Feature/serve clean

alejandrogallo:
Serve was very string heavy, it would discourage people
from contributing.
I wrote an html dsl for it except I realised that dominate
https://github.com/Knio/dominate/
is actually well-designed and it has no dependencies.

The code is very much simplified right now and it opens the door for easier extensibility
of features for the webapp.

I'll leave this here for a while before merging for comments.

### DIFF
--- a/papis/commands/serve.py
+++ b/papis/commands/serve.py
@@ -1,4 +1,3 @@
-import click
 import re
 import os
 import json
@@ -8,6 +7,9 @@ import urllib.parse
 from typing import Any, List, Optional, Union, Tuple, Callable
 import functools
 import cgi
+
+import click
+import dominate.tags as t
 
 import papis.api
 import papis.cli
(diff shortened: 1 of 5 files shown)


papis/commands/serve.py
```diff
@@ -455,33 +413,33 @@ def clear_cache(self, libname: str) -> None:
...
         self.wfile.flush()
 
     @ok_html
     def page_document(self, libname: str, ref: str) -> None:
         docs = papis.api.get_documents_in_lib(libname, ref)
         if len(docs) > 1:
             raise Exception("More than one document match %s", ref)
-        elif len(docs) == 0:
+        if len(docs) == 0:
             raise Exception("No document found with ref %s", ref)
```
alexfikl, line 439:
```suggestion
        if len(docs) == 0:
            raise Exception("No document found with ref '{}'".format(ref))
```
? And in the raise above. Before it should just print the tuple?

alexfikl (commented):
I'm afraid I don't have much to add, but I did run this and pushed all the buttons and it all seems to work! 

Left one nitpick, but the code also looks good.

alejandrogallo:
Great, I'll merge later, I hope it's more approachable the code like this,
before with all the strings it was daunting wanting to change something,
now future webapp features will hopefully be more of a breeze to implement

alejandrogallo:
The errors don't come from me, apparently flake8 just introduced some kind of regression. I squashed the commits and removed the doctor part, I did not realise that the doctor was also inside. Now merging.

alexfikl:
> The errors don't come from me, apparently flake8 just introduced some kind of regression. I squashed the commits and removed the doctor part, I did not realise that the doctor was also inside. Now merging.

Hm, yeah, `pyflakes` 3.0.0 seems to be out (previously 2.5.0), so maybe that's causing the issues :\ Not sure if that needs an update in `flake8-quotes` or where..

I'll take a look during the weekend if you don't get there first!

alejandrogallo:
My feeling is that we can prob wait, I don't think we are the only ones having this issue ;)

PR #426, opened 2022-11-25
Pin flake8<6 until flake8-quotes is fixed

alexfikl:
Turns out `flake8-quotes` is incompatible with `flake8` v6, so this just pins it to older versions for now.

I'll go ahead and merge once the CI passes again.


alejandrogallo:
Thank you for taking the time! That looks quite annoying from flake8

PR #427, opened 2022-11-26
Small style and performance fixes (refurb)

alexfikl:
Was looking at https://github.com/dosisod/refurb and implemented some of the fixes they recommend. Shouldn't be anything controversial (?).

### DIFF
--- a/papis/arxiv.py
+++ b/papis/arxiv.py
@@ -170,13 +170,13 @@ def find_arxivid_in_text(text: str) -> Optional[str]:
         ), re.I
     )
     miter = regex.finditer(text)
-    try:
+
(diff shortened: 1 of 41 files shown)


alejandrogallo (commented):
Ah, I have never used or knew about contextlib, great thanks, you taught me something, it looks much more ergonomic!

alexfikl:
> Ah, I have never used or knew about contextlib, great thanks, you taught me something, it looks much more ergonomic!

You're welcome! :tada: Yeah, it has a couple of nifty utilities, but maybe not very commonly needed. I probably used `contextlib.chdir` the most.

PR #429, opened 2022-11-27
CodeQL fixes

alexfikl:
This fixes two issues:
* do not use `tempfile.mktemp` (e.g. https://github.com/papis/papis/security/code-scanning/329) and use `tempfile.TemporaryFile` and friends in a context manager instead. This should also have the benefit of cleaning up the tests files after the test is run.
* use `pytest.monkeypatch` to change `os.environ` (see https://docs.pytest.org/en/7.1.x/how-to/monkeypatch.html)

The changes are just in the tests, so I'll merge when this passes :rocket:

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -22,30 +22,33 @@ class MockDownloader(Downloader):
 
 
 def create_random_pdf(suffix: str = "", prefix: str = "") -> str:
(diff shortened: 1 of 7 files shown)


alexfikl:
Hm, apparently windows did not appreciate that. I'll investigate a bit later!

PR #431, opened 2022-11-27
More CodeQL fixes

alexfikl:
Some more reported issues:
* inexact asserts from `unittest`, e.g. https://github.com/papis/papis/security/code-scanning/504
* overwriting attributes in `Downloader` subclasses, e.g. https://github.com/papis/papis/security/code-scanning/463
* some other misc fixes here and there


alejandrogallo:
Thank you for this, this is very important and not so rewarding task, it makes papis much better and future-proof!

PR #432, opened 2022-12-02
Fix click command heading

alexfikl:
This is meant to help address a note from https://github.com/papis/papis/discussions/413#discussion-4496598 (in the Commands section).

Before, the docs looked like https://papis.readthedocs.io/en/latest/commands.html#cli. This removes a level in the sections and looks like this (i.e. remove the extra `papis add` heading)

Screenshot_20221202_213746

@jghauser Can you take a look and see if this is more like what you had in mind? The code is quite hacky, but it seems to get the job done..

### DIFF
--- a/doc/source/commands/add.rst
+++ b/doc/source/commands/add.rst
@@ -1,4 +1,5 @@
 Add
 ---
+
 .. automodule:: papis.commands.add
 
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -1,8 +1,3 @@
-# -*- coding: utf-8 -*-
-#
-# papis documentation build configuration file, created by
-# sphinx-quickstart on Fri May 12 00:56:29 2017.
-#
 # This file is execfile()d with the current directory set to its
 # containing dir.
 #
@@ -12,23 +7,17 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import re
-import sys
 import os
-
-# Find the directory with papis installed
-sys.path.append(
-    os.path.join(
-        os.path.dirname(__file__),
-        "..", ".."
-    )
-)
+import sys
 import papis
+import docutils
+
+from docutils.parsers.rst import Directive
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-# sys.path.insert(0, os.path.abspath('.'))
+sys.path.append(os.path.join(os.path.dirname(__file__), "..", ".."))
 
 # -- General configuration ------------------------------------------------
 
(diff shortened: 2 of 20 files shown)


jghauser:
Thanks a lot for starting the work on this! I've been beginning with reorganising things as well, I'll open a PR soon. One question: how do I generate the html files locally so I can look at them? I installed sphinx with pip and then tried running `make html` but I get:
```
python /home/julian/.local/share/pyenv/shims/sphinx-build -b html -d build/doctrees   source build/html
  File "/home/julian/.local/share/pyenv/shims/sphinx-build", line 3
    [ -n "$PYENV_DEBUG" ] && set -x
      ^^^^^^^^^^^^^^^^^
SyntaxError: invalid syntax. Perhaps you forgot a comma?
make: *** [sphinx.mk:55: html] Error 1
```
Do you know what might be up here? Thanks!

alexfikl:
> Do you know what might be up here? Thanks!

I use 
```
cd doc
make html SPHINXOPTS="-W --keep-going -n"
```
to generate the html, which should then be in `build/html/index.html`. As for the error you're getting, that's probably my fault. In the Makefile, it sets
```
SPHINXBUILD   ?= python $(shell which sphinx-build)
```
but your `sphinx-build` is a shell script installed by `pyenv`, so that doesn't work. We should probably fix that. Until then, you can also just call sphinx directly using
```
cd doc
sphinx-build -b html -W --keep-going -n source build/html
```

jghauser:
Great, that worked, thanks! Checking it out now and will get back in a bit

jghauser:
This is much better. I have some other ideas for improvement. I'll use the `add` command to illustrate.

1. Multiple descriptions of what commands do.

There are sort of two short descriptions of what the commands do. One is at the very top of the help "The add command..." and the second is after "Command line interface", where it says "Add a document...". I think the two should be integrated. I think we might want to delete the description in the comment at the top of `add.py` and add the description to the `click` stuff at the bottom of the file. In that way the more complete description will appear also when using `--help`.

2. Order of information

I think we can delete the "command-line interface" heading and move all information under that heading to the very beginning. This documention is about the CLI commands and I think we don't need a subheading with that title.

Maybe the order could be something like:
- info coming from click help
- examples
- a ref to relevant configuration options (in the second paragraph of the `add` documentation, there's a reference to such options, and I think that might be some info that shouldn't be shown with `--help` (as would happen if we do (1)). Or maybe we just delete this entirely, but make it easily findable in the config section of the docs.

3. Information that shouldn't be there

Stuff in `Examples in python` shouldn't be here, but in some "Advanced config" section or so.

You can also just merge this and I can implement my suggestions myself in a follow-up commit (if others think this makes sense).

alexfikl:
Yeah, that organization order makes sense to me too. That's sort of how a lot of command-line tools are organized too (at least `git` ones seem to follow that), so it should be fairly intuitive to others.

> You can also just merge this and I can implement my suggestions myself in a follow-up commit (if others think this makes sense).

Yep, yep! This was just meant to see if we can make `sphinx-click` look a bit more like what we'd like, so extensive doc changes are definitely better left for another PR :rocket: I would suggest copy-pasting most of your comment to https://github.com/papis/papis/discussions/413 so it stays all in one place!

I'll wait to see if @alejandrogallo has any thoughts, but otherwise this is good to go from my end!

jghauser:
Ok perfect, I'll copy over that info into the discussion!

alejandrogallo:
It looks good for me! It is much better now :)

PR #438, opened 2022-12-13
Reorder config imports during initial setup

### LINKED ISSUE #436: Changing loading order for config setting
Hey,

I think the current order of loading and overwriting the settings (suchs as `picktool` or `editor`) is a bit strange compared to other command line programs. The current read and set order is:

1. Take default config
2. Read global config in papis config folder
3. Read arguments from `--set`
4. Read library specific configs

I propose to change 3. and 4., so that passing arguments to the command line directly (e.g. `papis --set opentool rofi open`) has priority over the library specific configuration.

This should be an easy fix, just moving the block https://github.com/papis/papis/blob/a8aaddf50033fa09b64a303e195db569e8bcd438/papis/commands/default.py#L259-L261

behind

https://github.com/papis/papis/blob/a8aaddf50033fa09b64a303e195db569e8bcd438/papis/commands/default.py#L276-L285

What do you think?

alexfikl:
This reorders the configuration loading like mentioned in #436. I've played around with it a bit and it doesn't seem to break anything for me, but I'm not a heavy user of either `--set` or library-local config files. Mostly tried variations of
```
papis -l papers config sort-field
papis -l papers --set sort-field doi config sort-field
``` 
@Sebastian-Jung If you could take a look / test this and see if it does what you want (it just moves the chunks of code around like you suggested), that would be awesome!

Fixes #436


papis/commands/default.py
```diff
@@ -252,37 +258,52 @@ def run(verbose: bool,
...
+    # merge its contents
+    local_config_file = papis.config.getstring("local-config-file")
     for path in library.paths:
-        # Now the library should be set, let us check if there is a
-        # local configuration file there, and if there is one, then
-        # merge its contents
-        local_config_file = os.path.expanduser(
-            os.path.join(path,
-                         papis.config.getstring("local-config-file")))
+        local_config_file = os.path.expanduser(os.path.join(path, local_config_file))
```
Sebastian-Jung, line 296: This only works for the first library but will fail for the others because you overwrite the variable.

Sebastian-Jung (changes requested):
Looks good except for the local_config_file variable being overwritten in the for loop. Haven't tested it though.

papis/commands/default.py
alexfikl: Nice catch! It's renamed in
https://github.com/papis/papis/compare/5827098cc978e5de394e168a7e60446eb6effaba..e2d5d5b65b3754300d1ce50206ea4c72ccedf747

Sebastian-Jung:
Looks good to me. Also tested it and it seems to work just fine 👌🏽

Sebastian-Jung (approved):
Looks good to me, tested it and it works. Not sure though if it has any complex and weird side effects somewhere else 😃

alexfikl:
> Looks good to me, tested it and it works. Not sure though if it has any complex and weird side effects somewhere else smiley

Thanks a lot for taking a look and testing! It was very helpful!

I'm not 100% sure of any side effects, so we'll wait to see if @alejandrogallo spots anything weird, but a priori it should be good :rocket:

alejandrogallo:
That looks good to me, thank you two!
(I'll ignore the vim folding.. haha, that's ok)

alexfikl:
> (I'll ignore the vim folding.. haha, that's ok)

Haha, old habits! I've removed the fold markers :-1:

PR #439, opened 2022-12-15
Move all dependencies from ci-install to setup.py

alexfikl:
This also adds the additional devel dependencies added in #421 

@alejandrogallo picked out the `setup.py` changes in here to see if the python 3.5 failures still happen.


setup.py
```diff
@@ -80,46 +103,33 @@
         "dominate",
     ],
     python_requires=">=3.5",
-    classifiers=[
```
alexfikl, line 83: Moved these a bit higher up so that all the dependencies (install and extras) are closer together.

alexfikl:
Ok, I think I figured out what was going on there. There were two issues:
* `python-lsp-server` does not support python 3.5 at all. The first version released already was `>=3.6`.
* the `ci-install` script wasn't failing if one of the commands failed, so we didn't see the error at that step.

I think this should fix it! It will cause some merge conflicts with #421 (sorry!) but hopefully nothing too bad!

alejandrogallo:
That's good, should we merge it and then I'll merge it to my papis-doctor branch and see if that's good

PR #440, opened 2022-12-15
Update how downloader tests handle external resources

alexfikl:
At the moment, the tests for the downloaders work something like this:
* get a "resource" from a website, i.e. the HTML page for a paper
* stick it in `tests/downloaders/resources/downloader_resource_n.html`
* run the test against it and see what the downloader can scrape
* put the result in a `tests/downloaders/resources/downloader_resource_n_out.json` (or some other format)
* Once all the data is there and works as expected, use the resources as cached known values to test against.

This is all a bit manual and tedious, so this PR introduces some helpers to make it easier to cache and update the resources (many resources are quite old). For this, it adds two functions to `tests.downloaders.__init__`
* `get_remote_resource(infile, url, **kwargs)` that gets the resource the downloader should scrape
* `get_local_resource(outfile, data, **kwargs)` that gets the expected result to check against `downloader.ctx.data`.

To control how they work, a `PAPIS_UPDATE_RESOURCES` environment variable is used. Basically, if `PAPIS_UPDATE_RESOURCES=remote` it will re-download the file at the given `url` on the call to `get_remote_resource`.

The motivation for this is that the downloaders can be quite useful, but a few our out of date and a lot of the tests are disabled in `setup.cfg` so we can't keep track of what's working :(

As an example, the `test_springer` test is enabled and updated to use this machinery. The idea being that  it should be easy to copy paste that, find some links for another downloader and run
```
PAPIS_UPDATE_RESOURCES=both python -m pytest test/downloaders/test_<downloader>.py
```
to have all the resources downloaded and all the files there already.

What do you think?


alejandrogallo:
This makes a lot of sense, it was something that had to be done, it's a great idea!

alexfikl:
> This makes a lot of sense, it was something that had to be done, it's a great idea!

Awesome! Thanks for taking a look! :heart: 

I'll go ahead and merge this and make other PRs to update the remaining downloader tests.

PR #443, opened 2022-12-17
Improve to_bibtex performance

alexfikl:
Some updates while doing #412. The main changes are:

* only do one sweep through the document keys (i.e. do the `bibtex_key_converter` directly)
* construct a `List[str]` with the bibtex lines and join at the end. Using `str += extra` is known to be quadratic.

For a few simple benchmarks (using hyperfine) trying to export ~20 and ~200 papers (library has about ~1500). Before it was
```
Benchmark 1: papis -l papers export -f bibtex -a greengard
  Time (mean ± σ):     420.5 ms ±   6.8 ms    [User: 440.8 ms, System: 194.0 ms]
  Range (min … max):   407.9 ms … 432.8 ms    10 runs

Benchmark 2: papis -l papers export -f bibtex -a optimization
  Time (mean ± σ):     525.7 ms ±   4.9 ms    [User: 562.8 ms, System: 184.5 ms]
  Range (min … max):   518.9 ms … 536.8 ms    10 runs
```
and now it is
```
Benchmark 1: papis -l papers export -f bibtex -a greengard
  Time (mean ± σ):     403.0 ms ±   3.9 ms    [User: 430.5 ms, System: 181.5 ms]
  Range (min … max):   397.6 ms … 408.9 ms    10 runs

Benchmark 1: papis -l papers export -f bibtex -a optimization
  Time (mean ± σ):     421.3 ms ±   4.5 ms    [User: 447.5 ms, System: 188.0 ms]
  Range (min … max):   414.8 ms … 426.8 ms    10 runs
```

Not much, but the new version doesn't seem to take noticeably more time when exporting more files!

### DIFF
--- a/papis/bibtex.py
+++ b/papis/bibtex.py
@@ -117,7 +117,7 @@ bibtex_key_converter = {
 
 
 def exporter(documents: List[papis.document.Document]) -> str:
-    return "\n".join(to_bibtex_multiple(documents))
+    return "\n\n".join(to_bibtex_multiple(documents))
 
 
 class Importer(papis.importer.Importer):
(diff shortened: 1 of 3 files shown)


alejandrogallo:
The changes look good! It's not much increase in performance but it can be as you say for bigger libraries one would notice a difference, which I might notice, that's good!

alexfikl:
> The changes look good! It's not much increase in performance but it can be as you say for bigger libraries one would notice a difference, which I might notice, that's good!

Thanks for taking a look!

I ran `papis --profile` and the remaining time seems to be spent in loading the db and doing the filtering (in this case), so left for another time.

PR #444, opened 2022-12-17
Update bibtex_types and bibtex_keys from biblatex manual

alexfikl:
The manual can be found at https://ctan.org/pkg/biblatex?lang=en. Is there some more authoritative source?

This PR updates the existing mappings and list of types and keys from the latest version of the document. Hopefully without any typos..

### DIFF
--- a/papis/bibtex.py
+++ b/papis/bibtex.py
@@ -1,7 +1,7 @@
 import os
 import string
 import logging
-from typing import Optional, List, Dict, Any, Iterator
+from typing import Optional, List, FrozenSet, Dict, Any, Iterator
 
 import click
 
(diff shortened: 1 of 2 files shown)


alejandrogallo:
I think the source is authoritative enough, that's great.i don't remember where I got the list from in the first place

alexfikl:
> I think the source is authoritative enough, that's great.i don't remember where I got the list from in the first place

Yep, I didn't find anything that looked better either, since that should be valid for `biblatex` and `biber`. Thanks for taking a look! :rocket:

papis/bibtex.py
```diff
@@ -13,35 +13,109 @@
...
+    "letter",
+    "movie",
+    "music",
+    "performance",
+    "review",
+    "standard",
+    "video",
+]) | frozenset(papis.config.getlist("extra-bibtex-types"))  # type: FrozenSet[str]
 
 bibtex_type_converter = {
```
alexfikl, line 57: I used these `converter` dicts for the aliases defined in the manual, but I'm not quite sure what the entries that were already there were.

@alejandrogallo  Do you happen to recall where does come from?

alejandrogallo: no, sadly I don't remember quite, this was some years ago, sorry

alexfikl: Did a bit of `git` blaming and it turns out this was introduced in 
https://github.com/papis/papis/pull/121
to handle some Zotero fields. Some of these do still appear in the Zotero schema
https://api.zotero.org/schema

I'll add a note for the future :D

alejandrogallo: that's wise ;)

PR #445, opened 2022-12-20
Unpin flake8 now that new flake8-quotes is released

alexfikl:


alexfikl:
Nevermind, we can't unpin because flake8 v6 only supports python >=3.6 and doesn't recognize type annotations in comments, so it emits a lot of "imported but unused" warnings..

PR #447, opened 2022-12-20
Update more downloader tests (cont.)

alexfikl:
This updates the remaining disabled downloader tests from `setup.cfg`.
* `acs`: updated, but does not work due to Cloudflare disallowing web scraping. There doesn't seem to be a good way around this.
* `hal`: seems to be in the progress of changing their domains [1] and this updates the downloader to also understand the new ones.
* `iopscience`: updated and seems to be working.
* `tandfonline`: updated and seems to be working.
* `fallback`: updated and working.

[1] https://www.ccsd.cnrs.fr/2022/06/hal-archives-ouvertes-fr-va-devenir-hal-science/


alexfikl:
The tests seem to be passing, so I'll go ahead and merge this!

PR #449, opened 2022-12-20
Initial working implementation of papis_id

alejandrogallo:
This is already working, I will probably update the webapplication to use this
as a test, and implement a couple of tests.

I think the changes were quite minimal and they are not noticeable in the runtime.
Also the unique id is the same for both databases (namely `papis_id`) and
the only part of papis that knows about it (`papis.id` module) (and probably will know at all in teh future)
is database, so that @jghauser in principle can now use

`papis open papis_id:lalalaalalaala`

in both databases interchangeably.
In python, there is the handy

```
db.find_by_id("lalalaalalala")
```
that throws if two documents match this id.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,26 @@
+VERSION v0.13
+=============
+
+## NEW `papis_id` special key in `info.yaml`
+
+In order to make plugin writing easier we have decided to introduce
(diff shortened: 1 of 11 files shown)


alejandrogallo:
This seems really not to cause any problems I will merge this.

Issue #448, opened 2022-12-20
`papis add` stuck in while loop on linux


prataffel:
* papis version (`$ papis --version` or commit number): a81e575006b630a3558063cfc6878338ce3bf683
* operating system: linux (result of `uname -r`: `5.15.81-1-MANJARO` 
* python version: `Python 3.10.8`

# how to reproduce the issue
Command and output of `papis --log INFO add --from doi 10.1090/noti963`

```
INFO:importer:doi:Using DOI '10.1090/noti963'
INFO:cli:add:There are 1 possible matchings
INFO:cli:add:Merging data from importer doi
- author:
- author_list:
- doi:
+ author: Gratzer, G.
+ author_list: [{'given': 'G.', 'family': 'Gratzer', 'affiliation': []}]
+ doi: 10.1090/noti963
- issue:
+ issue: 04
?        ++
- journal:
+ journal: Notices of the American Mathematical Society
- language:
+ language: en
?           ++
- month:
+ month: 4
?        +
- pages:
+ pages: 1
?        +
- publisher:
- title:
- type:
- url:
+ publisher: American Mathematical Society (AMS)
+ title: What Is New in LaTeX? VI. LaTeX on an iPad. Empire
+ type: article
+ url: http://dx.doi.org/10.1090/noti963
- volume:
+ volume: 60
?         ++
- year:
+ year: 2013
?       ++++
^^^^^^^^^
diff from
-----  -
+++++ Importer(doi, uri=10.1090/noti963)

Add all[a] Split[s] Reject[n] Quit[q] Cancel[c]
GENERAL DIFFERENCE

```
*papis is stuck forever*

---

As it appears to me, this happens because I am stuck in [this `while loop`](https://github.com/papis/papis/blob/a81e575006b630a3558063cfc6878338ce3bf683/papis/commands/add.py#L308). 

For me, this can be fixed by swapping to `pathlib.Path.is_relative_to` in `papis.utils.is_relative_to` (according to [the documentation of `os.path.relpath`](https://docs.python.org/3.10/library/os.path.html#os.path.relpath), a `ValueError` is only raised under Winows).

alexfikl:
Thanks for the report!

Your analysis seems completely correct, so hopefully it's an easy fix. I'll take a look today or tomorrow, but PRs are always very welcome!

prataffel:
[...]

PR #450, opened 2022-12-20
Fix being stuck in `while` loop when adding an entry

prataffel:
Should be a fix for #448.

Tests are not performed, as adding a trailing slash to `folder` when [creating the test library](https://github.com/papis/papis/blob/a81e575006b630a3558063cfc6878338ce3bf683/tests/__init__.py#L118) causes the tests to get stuck in the while loop without the fix, which I don't think is helpful.

### DIFF
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -305,7 +305,7 @@ def run(paths: List[str],
         temp_doc = papis.document.Document(data=data)
         temp_path = os.path.join(out_folder_path, folder_name)
         components = []     # type: List[str]
-        
+
         temp_path = os.path.normpath(temp_path)
         out_folder_path = os.path.normpath(out_folder_path)
 
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -306,7 +306,7 @@ def run(paths: List[str],
         temp_path = os.path.join(out_folder_path, folder_name)
         components = []     # type: List[str]
         while (
-                temp_path != out_folder_path
+                not papis.utils.paths_are_identical(temp_path, out_folder_path)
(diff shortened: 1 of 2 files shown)


papis/commands/add.py
```diff
@@ -306,7 +306,7 @@ def run(paths: List[str],
         temp_path = os.path.join(out_folder_path, folder_name)
         components = []     # type: List[str]
         while (
-                temp_path != out_folder_path
+                not papis.utils.paths_are_identical(temp_path, out_folder_path)
```
alexfikl, line 309: Maybe you can stick a [`os.path.normpath`](https://docs.python.org/3/library/os.path.html#os.path.normpath) in front of `temp_path` and `out_folder_path` before starting this loop instead of using `paths_are_identical`? That would avoid them getting cast to a `Path` over and over again and handle the "trailing backslash" issue.

alexfikl (approved):
Left a little nitpick, but otherwise this looks good to me! Thanks for the help!

> Tests are not performed, as adding a trailing slash to folder when creating the test library causes the tests to get stuck in the while loop without the fix, which I don't think is helpful.

If this fixes it for you, it's completely fine, but for future reference, you can always do something like
```
from multiprocessing import Process
p = Process(target=my_actual_test)
p.start()
p.join(timeout=1)
```
and then fail if the process is still alive after a second. That should catch if some things take waaaay too long (like an infinite loop)! :D

papis/commands/add.py
prataffel: This should work, too and is easier as you pointed out. Thank you very much.

alexfikl: Perfect, thank you!

PR #454, opened 2022-12-22
Improvements to commands.config

alexfikl:
Was trying to figure out what configs were used in my venv and fell down a little rabbit hole!

This make some updates to the ``config`` command:
* Improve the docs a bit.
* Do not show scary tracebacks on errors, but some hopefully helpful error messages.
* Allow more positional arguments to query multiple settings at once.
* Added `--default` to show all the defaults papis knows about for a given section (could be helpful to explore the settings without going into the code)
* Added a `--json` flag to print multiple options in the json format instead of the default INI-ish format.

@jghauser if you have the time to take a look at the documentation parts (phrasing and such), it would be greatly appreciated, as usual!

### DIFF
--- a/papis/commands/config.py
+++ b/papis/commands/config.py
@@ -63,7 +63,7 @@ import click
 import papis.config
 import papis.commands
 
-logger = logging.getLogger("config:run")
+logger = logging.getLogger(__name__)
 
 
 def run(option_string: str) -> Optional[str]:
@@ -75,15 +75,19 @@ def run(option_string: str) -> Optional[str]:
         section = option[0]
         key = option[1]
     else:
+        raise ValueError("unsupported option format: '{}'".format(option_string))
+
+    logger.debug("key = %s, sec = %s", key, section)
+    return papis.config.get(key, section=section)
+
+
+def _run_with_log(option_string: str) -> Optional[str]:
+    try:
+        return run(option_string)
+    except ValueError:
         logger.error(
             "options should be in a <section>.<key> or <key> format: got '%s'",
             option_string)
-        return None
-
-    logger.debug("key = %s, sec = %s", key, section)
-
-    try:
-        return papis.config.get(key, section=section)
     except papis.exceptions.DefaultSettingValueMissing as exc:
         logger.error("\n%s", str(exc).strip("\n"))
 
@@ -127,9 +131,9 @@ def cli(options: List[str],
         if len(options) == 1:
             # NOTE: a single option is printed directly for a bit of backwards
             # compatiblity and easier use in shell scripts, so remove with care!
-            click.echo(run(options[0]))
+            click.echo(_run_with_log(options[0]))
         else:
             for option in options:
-                value = run(option)
+                value = _run_with_log(option)
                 if value is not None:
                     click.echo(format(option, value))
--- a/papis/commands/config.py
+++ b/papis/commands/config.py
@@ -154,8 +154,8 @@ def run(
 
     if len(options) == 0:
         # NOTE: no options given -> just get all the settings
+        defaults = papis.config.get_default_settings()
         if default:
-            defaults = papis.config.get_default_settings()
             if section is None:
                 result = defaults
             elif section in defaults:
@@ -166,11 +166,23 @@ def run(
                              section, "', '".join(defaults))
         else:
             if section is None:
-                result = {
-                    sec: {k: config[sec][k] for k in config.options(sec)}
-                    for sec in config if sec != "DEFAULT"
-                    }
+                for sec in defaults:
+                    result[sec] = defaults[sec].copy()
+                    if sec in config:
+                        result.update({
+                            key: papis.config.get(key, section=sec)
+                            for key in config.options(sec) if key in result
+                            })
+            elif section in defaults:
+                result = defaults[section].copy()
+                if section in config:
+                    result.update({
+                        key: papis.config.get(key, section=section)
+                        for key in config.options(section) if key in result
+                        })
             elif section in config:
+                # NOTE: libraries are in the config but not in the defaults,
+                # so we just print whatever settings are in there.
                 result = {key: config[section][key] for key in config.options(section)}
             else:
                 logger.error("Section '%s' not found in configuration file. "
--- a/papis/commands/config.py
+++ b/papis/commands/config.py
@@ -19,7 +19,7 @@ If you wanted to see which ``dir`` the ``books`` library uses, you would do:
 
     papis -l books config dir
 
-With ``-l``, Papis selects a specific library (here, the "books" library). The 
+With ``-l``, Papis selects a specific library (here, the "books" library). The
 rest works just as above.
 
 Settings from a specific section in the configuration file can also be
@@ -32,18 +32,20 @@ example, if your ``books`` library is configured as a section, you can do:
 
 This is equivalent to the above ``papis -l books config dir`` command.
 
-For a more complex example, the :ref:`Bibtex` command has its own
-configuration settings. These can be accessed through
+The same notation can also be used to access configuration settings specific
+to Papis commands. For example, the :ref:`Bibtex` command's settings can be
+accessed with:
 
 .. code::
 
     papis config bibtex.default-read-bibfile
-    > main.bib
 
 You can find a list of all available settings in the configuration section
 at :ref:`general-settings`. Commands and other plugins can define their own
-settings, which are documented separately. To list the default values in a
-given section, use
+settings, which are documented separately.
+
+You can also use ``config`` to find out about some section's default
+settings:
 
 .. code::
 
--- a/papis/commands/config.py
+++ b/papis/commands/config.py
@@ -56,7 +56,7 @@ Command-line Interface
 """
 
 import logging
-from typing import List, Optional
+from typing import Any, List, Optional
 
 import click
 
@@ -93,11 +93,37 @@ def run(option_string: str) -> Optional[str]:
 @click.command("config")
 @click.help_option("--help", "-h")
 @click.argument("options", nargs=-1)
-def cli(options: List[str]) -> None:
+@click.option(
+    "--list-defaults", "list_defaults",
+    help="List all default configuration settings in the given section",
+    default=False, is_flag=True)
+def cli(options: List[str],
+        list_defaults: bool) -> None:
     """Print configuration values"""
-    logger.debug("Config options: %s", options)
+    import colorama
 
-    for option in options:
-        value = run(option)
-        if value is not None:
-            click.echo("{} = {}".format(option, value))
+    logger.debug("config options: %s", options)
+    logger.info("Configuration file: '%s'", papis.config.get_config_file())
+
+    def format(key: str, value: Any) -> str:
+        return (
+            "{c.Style.BRIGHT}{key}{c.Style.NORMAL} "
+            "= {c.Fore.GREEN}{value!r}{c.Style.RESET_ALL}"
+            .format(c=colorama, key=key, value=value))
+
+    if list_defaults:
+        defaults = papis.config.get_default_settings()
+        for option in options:
+            if option not in defaults:
+                logger.error("Section '%s' has no defaults", option)
+                continue
+
+            click.echo("")
+            click.echo("[{}]".format(option))
+            for key, value in defaults[option].items():
+                click.echo(format(key, value))
+    else:
+        for option in options:
+            value = run(option)
+            if value is not None:
+                click.echo(format(option, value))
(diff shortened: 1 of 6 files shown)


jghauser:
This is great, love the ability to query multiple settings at once. I use this in papis.nvim and hopefully it will speed up things a bit! I'll have a look at the docs!

alexfikl:
> This is great, love the ability to query multiple settings at once. I use this in papis.nvim and hopefully it will speed up things a bit! I'll have a look at the docs!

@jghauser Ah! This changed the output to show `<key> = <value>` instead of just `<value>` for each option. Will that break stuff for you? 

What would you be a useful format to list more options? Add a `--json` flag or just list them in order?

jghauser:
Not a big deal if things break here, I can easy fix things! Though maybe nicer if the output remains the same when querying single settings. I only use this when the user hasn't copied over the settings explicitly to neovim (which I recommend people do as querying papis is rather slow and slows down neovim startup).

alexfikl:
> Not a big deal if things break here, I can easy fix things! Though maybe nicer if the output remains the same when querying single settings. I only use this when the user hasn't copied over the settings explicitly to neovim (which I recommend people do as querying papis is rather slow and slows down neovim startup).

The output for a single option should be restored now!

If you want a specific format for showing multiple options, let me know! Although that can probably be extended in the future with some flags, so not a particularly urgent issue.

jghauser:
Yeah, I think for now this is good. I do have some ideas for betting settings management in papis.nvim and might get back to this when I get around it (I think json might come in handy!).

alejandrogallo:
this is great !

jghauser:
What's the best way for me to submit changes to the docs here @alexfikl ? should I just edit the relevant files and commit directly to this fork (that seems to be possible)? I feel like adding suggestions is a bit too laborious an approach for prose...

jghauser:
Oh oh, missed a trailing whitespace. Lemme fix this quickly!

jghauser:
Ok, hopefully tests will pass now now. While updating the docs, two things confused me a little.

The `config` command usually queries setting's values, but with the `--list-defaults` argument, it's a section's values. Would it maybe make more sense to allow querying both sections and individual settings in some consistent way?

Also, when querying a setting that is set to an empty string, it might look to a user as if nothing was returned. I'm thinking maybe your earlier choice to return "key = value" pairs is better after all.

alexfikl:
Thanks a lot for the help with the docs! It looks a lot clearer and easier to follow now :heart: 

> The `config` command usually queries setting's values, but with the `--list-defaults` argument, it's a section's values. Would it maybe make more sense to allow querying both sections and individual settings in some consistent way?

Yeah, that cross my mind too, but I wasn't sure how to change it. Some things that come to mind
1. When calling with `--list-defaults`, just print the defaults in all the sections.
2. Add a flag to explicitly mention the section, which could also benefit querying more options in the same section, e.g.
```
papis config --section <section> <key>
``` 
as an alternative to
```
papis config <section>.<key>
```
? And then we can call
```
papis config --section <section> --list-defaults
```

That looks a bit more consistent to me, but suggestions are very welcome! I'm inclined to choose the first for simplicity. Would you have a use case for the more granular queries?

> Also, when querying a setting that is set to an empty string, it might look to a user as if nothing was returned. I'm thinking maybe your earlier choice to return "key = value" pairs is better after all.

Hm, we could add a warning? I'll have to check if logger warnings go to `stderr`.
Otherwise, I'm not too bothered with just leaving it be, since most cli tools tend to go by "no news is good news!".

jghauser:
I like your second proposal, it's very consistent. Would that then also allow querying individual defaults? As in, would `papis config <key> --list-defaults` be admissible? Also, now that I'm thinking about it, it would also be neat to have `papis config` (without arguments) spit out a list of all (and with `--list-defaults` all default) settings.

And you're right about the "no news is good news!" thing. It's how most CLI tools do it, so we should probably follow that.

alexfikl:
> I like your second proposal, it's very consistent. Would that then also allow querying individual defaults? As in, would `papis config <key> --list-defaults` be admissible? Also, now that I'm thinking about it, it would also be neat to have `papis config` (without arguments) spit out a list of all (and with `--list-defaults` all default) settings.

I like that as a workflow! I'll get on it :D

alexfikl:
Ok, I reworked it a bit to be more in line with what you proposed. It should be fairly complete now!

To summarize a bit:
* Querying specific settings still works with `papis config some-option` and adding `--default` to that will instead print the default value.
* Querying sections also works with `papis config -s some-section`, which prints all the values in that section + adding `--default` prints all the defaults instead + adding options prints specific options from the section.
* Querying multiple values + multiple defaults also works just the same.
* Just calling `papis config` will print all the settings in all the sections + with support for `--default` as well.

@jghauser I updated the docs to describe these new options, so another look is probably warranted :D

jghauser:
This is great! I've been playing around with it for a bit and have some comments/questions:

- What is shown when querying without `--default`? I feel like as a user my assumption would be that it shows whatever Papis is using, so either the explicitly set setting or, if that doesn't exist, the default (whereas with `--default`, always the default is shown). But I think what's happening is that a setting is only shown if it's explicitly set? But I might be wrong with my expectations here, what do you think? In either case, I'll make sure to document it.
- I think the settings I set in my user `config.py` are counted as default. Is that necessitated by how settings are set by Papis or could we change that? If it's not (easily) changeable, I'll make a note of it in the docs.

alexfikl:
>     * What is shown when querying without `--default`? I feel like as a user my assumption would be that it shows whatever Papis is using, so either the explicitly set setting or, if that doesn't exist, the default (whereas with `--default`, always the default is shown). But I think what's happening is that a setting is only shown if it's explicitly set? But I might be wrong with my expectations here, what do you think? In either case, I'll make sure to document it.

If querying without `--section`, it should work the way you described, i.e.
```
papis config <key> 				-> either config file value or default
papis config --default <key>	-> always default
```

However, when querying with a section, it just shows stuff from the configuration file, i.e.
```
papis config --section <sec>		  	-> all keys in the configuration file (no defaults)
papis config --default --section <sec>	-> all defaults (whether in the config file or not)
``` 

If you add a key and a section, i.e. `papis config --section <sec> <key>`, it should still pick up the default if there is one. Is that what you're seeing as well?

>     * I think the settings I set in my user `config.py` are counted as default. Is that necessitated by how settings are set by Papis or could we change that? If it's not (easily) changeable, I'll make a note of it in the docs.

Hm, I only use the static INI file for the configuration, so haven't had any interaction with the `config.py` thing. I'll take a look if that does something different and get back to you!

jghauser:
Ok, thanks for the explanation!

> If you add a key and a section, i.e. papis config --section <sec> <key>, it should still pick up the default if there is one. Is that what you're seeing as well?

Yes, I get the same behaviour!

I think it would be better if adding --section didn't change the behaviour re showing only explicitly set settings without `--defaults`.  To keep behaviour consistent across versions of the command, I would prefer the command without `--defaults` to always show whatever settings Papis is using (explicitly set or default). Or maybe I'm missing something -- do you prefer it the way it's set up now? Though I don't think this is super crucial and if tedious to implement, I'll simply make a note in the docs :).

alexfikl:
> I would prefer the command without --defaults to always show whatever settings Papis is using (explicitly set or default).

Agreed that that's what we want to go for! When is it not doing that? Can you given an example?

jghauser:
> Agreed that that's what we want to go for! When is it not doing that? Can you given an example?

What I mean is the difference between these two:
- `papis config <key>`
- `papis config --section <sec>`

Where the first gives either explicitly set settings or defaults and the second only gives the explicitly set stuff. I would intuitively interpret the second command as implying the not returned settings aren't set at all (or to nil). In that instance, if I understand things correctly, `papis config` would not give me the values used by Papis.

> EDIT: Also, I looked a bit at using a config.py file and it seems like you can just put anything in there and it'll get exec-ed. I don't think we can do much about that if users modify the default settings in there..

Fair enough, that makes sense. Given that the `config.py` is an advanced thing anyway, I don't think this is an issue. I'll make a note about it in the docs.

alexfikl:
> What I mean is the difference between these two:
> 
>     `papis config <key>`
>     `papis config --section <sec>`
> 
> 
> Where the first gives either explicitly set settings or defaults and the second only gives the explicitly set stuff. I would intuitively interpret the second command as implying the not returned settings aren't set at all (or to nil). In that instance, if I understand things correctly, `papis config` would not give me the values used by Papis.

Ah, I see! Thank you for explaining and sorry for being a bit dense about it!

That shouldn't be too hard to implement :rocket:

jghauser:
Nothing to be sorry about, you're doing all the hard work here :)

alexfikl:
Ok, I think this does what you suggested now! For an example, using the [bibtex section settings](https://github.com/papis/papis/blob/287f2d8199e6ad85e4a5c31cfde1e809a15dd7a1/papis/commands/bibtex.py#L152-L156), I get
```
> papis config --default --section bibtex
[bibtex]
default-read-bibfile = ''
auto-read = ''
default-save-bibfile = ''
```
and without the defaults
```
> papis config --section bibtex
[bibtex]
default-read-bibfile = 'references.bib'
auto-read = 'true'
default-save-bibfile = ''
```
which is updated with the settings I have in my configuration file. Does it work as expected now?

jghauser:
:tada: I checked it out and it's now really working exactly like I suggested. Thanks! :)

alexfikl:
> I checked it out and it's now really working exactly like I suggested. Thanks! :)

Awesome! Thanks a lot for all the help with this!

papis/commands/config.py
```diff
@@ -1,34 +1,60 @@
...
 
-This works as well for default settings, i.e., settings that you have not
-customized, for example the setting ``match-format``, you would check
-it with
+With ``-l``, Papis selects a specific library (here, the "books" library). The
+rest works just as above.
+
+Settings from a specific section in the configuration file can also be
+accessed. To take an example, the `Bibtex`_ command's settings can be
+accessed with
```
jghauser, line 37:
```suggestion
You can also query a specific setting within a section. For example like this:
```

papis/commands/config.py
```diff
@@ -1,34 +1,60 @@
...
+The ``config`` command returns the value used by Papis. Therefore, if you
+haven't customized some setting, it will return the default value. In contrast,
+if you have customized it, it will return the value set in the configuration
+file. To show the list of default and overwritten configuration values in a
+section use
+
+.. code::
+
+    papis config --default --section <name>
+    papis config --section <name>
```
jghauser, line 14:
```suggestion
    papis config --section bibtex
    papis config --default --section bibtex
```

papis/commands/config.py
```diff
@@ -1,34 +1,60 @@
...
+Settings from a specific section in the configuration file can also be
+accessed. To take an example, the `Bibtex`_ command's settings can be
+accessed with
 
 .. code::
 
-    papis config match-format
-    > {doc[tags]}{doc.subfolder}{doc[title]}{doc[author]}{doc[year]}
+    papis config --default --section bibtex default-read-bibfile
+    papis config --section bibtex default-read-bibfile
```
jghauser, line 42:
```suggestion
    papis config --section bibtex default-read-bibfile
    papis config --default --section bibtex default-read-bibfile
```

papis/commands/config.py
```diff
@@ -1,34 +1,60 @@
...
 
-You can find a list of all available settings in the configuration section.
+For some more advanced usage, we can also query multiple settings at once. Then
+it may be useful to overwrite the selected section for several of the settings.
+This can be achieved by
```
jghauser, line 46:
```suggestion
Alternatively, you can also you the ``<section.setting>`` format to query the value of
a setting in a specific `section`:

.. code::
    papis config bibtex.default-read-bibfile
    papis config --default bibtex.default-read-bibfile
    
For some more advanced usage, we can also query multiple settings at once.  Here,
sections specified with ``<section>.<setting>`` override the section specified by
``--setting <section>``. This can be achieved by:
```

jghauser (commented):
Finally had time to go through the docs and made some changes. I've been trying out the comments functionality, but it's really a bit clunky for more than typo fix in the case of prose. It's really quite hard to keep an overview of the whole doc. Hope I didn't mess up anything :sweat_smile:

papis/commands/config.py
```diff
@@ -1,34 +1,60 @@
...
 .. code::
 
     papis -l books config dir
 
-This works as well for default settings, i.e., settings that you have not
-customized, for example the setting ``match-format``, you would check
-it with
+With ``-l``, Papis selects a specific library (here, the "books" library). The
+rest works just as above.
+
```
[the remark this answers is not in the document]
alexfikl: Yep, you're right. Removed it in [8c81871](https://github.com/papis/papis/pull/454/commits/8c81871e5c115e5c79cdc79d410b8e34f77e2495).

alexfikl:
This looks good to me now, so I went ahead and merged it! Thanks a lot for all the help and suggestions :heart:

alejandrogallo:
wow this is really fantastic, great job you two!!

PR #455, opened 2022-12-22
commands.add: call Importer.match instead of constructor

alexfikl:
Not sure why this was calling the constructor before, but it seems incorrect. This makes it work more like `get_matching_importer_or_downloader` and other places.

@alejandrogallo Does this make sense? The motivation was to allow importers to pass more parameters to the constructor in `match`, e.g. the `arxivid` for the arXiv ones.

### DIFF
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -568,14 +568,15 @@ def cli(files: List[str],
             "Select matching importers (for instance 0, 1, 3-10, a, all...)")
         matching_importers = [matching_importers[i] for i in _range]
 
-    for importer_tuple in from_importer:
+    for name, resource in from_importer:
         try:
-            importer_name = importer_tuple[0]
-            resource = importer_tuple[1]
-            importer = import_mgr[importer_name].plugin(uri=resource)
-            importer.fetch()
-            if importer.ctx:
-                matching_importers.append(importer)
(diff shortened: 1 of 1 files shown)


alexfikl:
> then the `doi` is in `from_importer` inside. I think the logic is to blindly trust the user and not match since otherwise there is no way of forcing it. The matching is kind of done implicitly since if the user was completely wrong, the `importer.ctx` test will give false.

I think it keeps most of that logic (in spirit at least), i.e. if the match fails it shows and error and doesn't try to use the importer anymore. Before it would fail in `fetch` or accessing `ctx` like you said and be caught by the `try..except` block around it.

> This is what I take out of it. But it's quite a messy business, we should refactor `add` and `update`, I started writing a `papis.smart_importer` module to pack this functionality there and use it from the webapp. If you're in the mood of taking a look at this importer business I can prepare a pull request.

Yeah, I can definitely take a look! 

Cleaning up and refactoring the existing logic in `add` and `update` into something more robust and reusable should be pretty cool! From my experience, it's always been a bit hit and miss, so these days I tend to be explicit and ask for a specific importer / downloader.

alexfikl:
@alejandrogallo Wanted to close this, but just to confirm.. You're not a big fan of using `match` here and are making some better changes in #457?

PR #458, opened 2022-12-22
arxiv: support both urls and arxivids in importer

alexfikl:
I sometimes have pdfs from arXiv that are conveniently named `<arxivid>.pdf` so its easier to do a `papis add --from arxiv '<arxivid>' <arxivid>.pdf` and not fill out the whole url.

This adds support to the `Importer` to first check if it's an id.


alexfikl:
Added benefit.. just noticed that this makes the docstring correct too! :rocket:

PR #462, opened 2022-12-24
Ansi colors

jghauser:
I'm trying to get papis to default to using the colours defined by my terminal. Basically, I've replaced all mentions of fg/bg:COLOR to fg/bg:ANSICOLOR. I'm unsure still about two things

1. In line 127 in tui/utils.py there's an `example_style` thing that has all sorts of odd colours defined (various shades of green among others). Where are these used? I can't think of ever seeing them... 
2. How do I change the colours used for example in the "DUPLICATION WARNING" tui? These still don't adhere to the ansi styles. See this screenshot:
20221224_14h31m19s_grim

Finally, some questions about contributing:
- How do I run the tests again? `make test` as mentioned in `CONTRIBUTING.md` doesn't seem to work. It gives me `make: *** No rule to make target 'test'.  Stop.`
- Ignore the messy commits. Git is confusing and I mostly don't have a clue of what I'm doing :sweat_smile:


alexfikl:
Haven't look at the code, but just to answer a few questions!

> 2. How do I change the colours used for example in the "DUPLICATION WARNING" tui? These still don't adhere to the ansi styles. See this screenshot:

The colors for that warning in particular is hardcoded here
https://github.com/papis/papis/blob/2ebc7ee643cb8d7aa46d7a6ab644f5d8382bf321/papis/commands/add.py#L405

There's a few places around the code that hardcode the colors in the message string, so you might want to grep for `c.Fore` or something like that to find them all.

> * How do I run the tests again? `make test` as mentioned in `CONTRIBUTING.md` doesn't seem to work. It gives me `make: *** No rule to make target 'test'.  Stop.`

I run pytest directly, e.g.
```
python -m pytest --sw -v -s tests
python -m pytest --sw -v -s tests/subdir/test_something.py
```
where `--sw` will stop at the first test failure, `-v` makes it a bit more verbose and `-s` captures and shows stdout / stderr.

jghauser:
Thanks, that's helpful! I'll see if I can figure out why colorama doesn't use ansi colors.

jghauser:
Actually, I just realised that my earlier message was misleading. The problem isn't with the colorama bit, that uses the right colours. The problem is with the part below, where the yaml content of the file is dumped. There, both the green and red aren't using my terminal's colours. This seems to happen here: 

https://github.com/papis/papis/blob/2ebc7ee643cb8d7aa46d7a6ab644f5d8382bf321/papis/commands/add.py#L411-L416

`text_area()` in `tui\utils.py` uses prompt-toolkit if I understand things correctly. So I guess the question is then how to make prompt-toolkit use ansi colours?

alejandrogallo:
about the tests I updated the `HACKING.md` document, to tell users to run
`./tools/ci-run-tests.sh`, this should be the default script to run tests,
because I also often don't really remember how to run them.

About the colours, in order to be on the same page I explain.
There are 2 sources of colours, on the one hand `colorama` and on the other
`prompt_toolkit`. The `text_area` bit is coming from `prompt_toolkit`.
The text_are uses the colors from the class `PygmentsLexer` in https://github.com/papis/papis/blob/master/papis/tui/utils.py#L72

I think the problem is the styling of `prompt_toolkit`.
Namely the pygments classes with them has a styling like so

```python
# The default Pygments style, include this by default in case a Pygments lexer
# is used.
PYGMENTS_DEFAULT_STYLE = {
    "pygments.whitespace": "#bbbbbb",
    "pygments.comment": "italic #408080",
    "pygments.comment.preproc": "noitalic #bc7a00",
    "pygments.keyword": "bold #008000",
    "pygments.keyword.pseudo": "nobold",
    "pygments.keyword.type": "nobold #b00040",
    "pygments.operator": "#666666",
    "pygments.operator.word": "bold #aa22ff",
    "pygments.name.builtin": "#008000",
    "pygments.name.function": "#0000ff",
    "pygments.name.class": "bold #0000ff",
    "pygments.name.namespace": "bold #0000ff",
    "pygments.name.exception": "bold #d2413a",
    "pygments.name.variable": "#19177c",
    "pygments.name.constant": "#880000",
    "pygments.name.label": "#a0a000",
    "pygments.name.entity": "bold #999999",
    "pygments.name.attribute": "#7d9029",
    "pygments.name.tag": "bold #008000",
    "pygments.name.decorator": "#aa22ff",
    # Note: In Pygments, Token.String is an alias for Token.Literal.String,
    #       and Token.Number as an alias for Token.Literal.Number.
    "pygments.literal.string": "#ba2121",
    "pygments.literal.string.doc": "italic",
    "pygments.literal.string.interpol": "bold #bb6688",
    "pygments.literal.string.escape": "bold #bb6622",
    "pygments.literal.string.regex": "#bb6688",
    "pygments.literal.string.symbol": "#19177c",
    "pygments.literal.string.other": "#008000",
    "pygments.literal.number": "#666666",
    "pygments.generic.heading": "bold #000080",
    "pygments.generic.subheading": "bold #800080",
    "pygments.generic.deleted": "#a00000",
    "pygments.generic.inserted": "#00a000",
    "pygments.generic.error": "#ff0000",
    "pygments.generic.emph": "italic",
    "pygments.generic.strong": "bold",
    "pygments.generic.prompt": "bold #000080",
    "pygments.generic.output": "#888",
    "pygments.generic.traceback": "#04d",
    "pygments.error": "border:#ff0000",
}

which does not use ansi color codes, because in prompt_toolkit parlance ansi color codes are called
`ansired` ...

**Comment by jghauser** [comment]
Thanks a lot! I've had a look at `PygmentsLexer` trying to figure out where I could explicitly load a pygments style in order to override the defaults with something that uses ansi colors, but I'm not quite sure I understand how I would do that. Could you point me in the right general direction, so I can give this a stab?

**Comment by alejandrogallo** [comment]
I am not entirely sure, but I would say something like


```py
PAPIS_PYGMENTS_DEFAULT_STYLE = {
    "pygments.whitespace": "#bbbbbb",
    "pygments.comment": "italic #408080",
    "pygments.comment.preproc": "noitalic #bc7a00",
    "pygments.keyword": "bold #008000",
    "pygments.keyword.pseudo": "nobold",
    "pygments.keyword.type": "nobold #b00040",
    ...
}
```

and then one has to create a prompt_toolkit style that is understood

```py
from prompt_toolkit.styles import Style

papis_style = Style.from_dict(PAPIS_PYGMENTS_DEFAULT_STYLE)
```

and then I am not sure, maybe it's enough to put this style in

```py
    text_window = Window(height=text_height,
                         style=papis_style,
                         content=BufferControl(buffer=buffer1, lexer=lexer))
```

in https://github.com/papis/papis/blob/master/papis/tui/utils.py#L73, I would say so, but I'm not 100% sure. It's not clear to me how to tell
the lexer to use our style, but since they named it lexer I would suppose that it only
does lexing, and the styling can be overriden later in the pipeline.

jghauser:
Thanks @alejandrogallo , I will check this out!

jghauser:
I've done some digging and have figured out how to print nicely ansicolored text. However, I've not managed to do so for `BufferControl` content types (here's [a prompt_toolkit issue](https://github.com/prompt-toolkit/python-prompt-toolkit/issues/616) detailing why it's not as straightforward as @alejandrogallo's suggestion). Thinking about this a bit more, I've also realised that the two instances in which `text_area` is used (in `add.py` and in `rm.py`), the ability of editing the text is both unnecessary and (in my opinion) confusing. I'm wondering whether it wouldn't make sense to just convert text_area into a pager instead. I think if we want to edit multiline texts, simply opening the preferring editor would be the better option in my opinion. Let me know what you think!

alejandrogallo:
I totally agree, let's think what would be the best behaviour. I think as you say, that editing is irrelevant so that the editor would not be suitable here.

I'm thinking maybe the document will get described with a format that can be overriden by the user, and then there is a yes no question with y and n. Back then I did not want an editor because if for instance vim gets started, then the user does not know why vim has been fired, because the information before that the document is a duplicate and so on is hidden by the fact that vim just started. But if you think about this and think of a good ergonomic option please do tell, these things should be as ergonomic as possible

jghauser:
Yeah, I think that makes sense. The process would be the following:
1. You get a duplication warning and are shown some info (I actually don't mind how it's shown now, we could simply make it static non-editable text)
2. At the bottom, you are asked: "Do you want to add the new document? (Y/n)"

This would be simpler than now, there wouldn't be a first quit/save option before deciding whether or not to add. Does anyone have any workflow where giving the option to make changes to some file and re-attempt adding would be useful? Maybe we should also give the option of editing the `info.yaml` of the document to be added to sort out the warning manually (so then we would have add/not add/edit options)?

Simply turning `text_area` into a pager would also allow us to pretty much re-use it in `rm.py` where it's used to ask the user whether this is indeed the right document to delete. There, too, we want to just display some text and give a y/n choice.

jghauser:
I've updated this a bit. I've tried to do the following:
- remove all non ansi color codes
- make things look nice on both dark and light mode (I mostly use dark mode and before some texts were hard to read, for example in the diff modes)
- rephrased some messages (esp. diff mode)
- removed colors from the duplication warning (I think it's not needed now that we have nice colors for logging levels)
- made the `text_area` stupider (it's now a pager and can't be used to edit text) but prettier (it's got nice ansi color highlighting)

Let me know what you think :)

papis/commands/add.py
```diff
@@ -421,7 +419,7 @@ def run(paths: List[str],
         for d_path in tmp_document.get_files():
             papis.utils.open_file(d_path)
     if confirm:
-        if not papis.tui.utils.confirm("Really add?"):
+        if not papis.tui.utils.confirm("Do you want to add the new document anyway?"):
```
alexfikl, line 422:
```suggestion
        if not papis.tui.utils.confirm("Do you want to add the new document?"):
```
? Since this should show up during normal working conditions (not just if there's a duplicate or something)

papis/tui/widgets/diff.py
```diff
@@ -190,7 +192,7 @@ def oset(event: Event, option: str, value: bool) -> None:
 
     diffshow(
         texta=texta, textb=textb,
-        title="GENERAL DIFFERENCE",
+        title="all changes",
```
alexfikl, line 195:
```suggestion
        title="All changes",
```
? It would show up as `--- Diff view: All changes ---`. Not sure which way's better, but it popped out to me that it wasn't capitalized..

papis/tui/widgets/diff.py
```diff
@@ -217,7 +219,7 @@ def oset(event: Event, option: str, value: bool) -> None:
         diffshow(
             texta=str(dicta.get(key, "")) + "\n",
             textb=str(dictb.get(key, "")) + "\n",
-            title="Key: {0}".format(key),
+            title='changes for key "{0}"'.format(key),
```
alexfikl, line 222:
```suggestion
            title='Changes for key "{0}"'.format(key),
```
?

papis/commands/rm.py
```diff
@@ -146,15 +146,14 @@ def cli(query: str,
     if not (_file or _notes):
         for document in documents:
             if not force:
-                tbar = "The folder {0} would be removed".format(
-                    document.get_main_folder())
-                logger.warning("This document will be removed, check it")
+                logger.warning("Folder to be removed: {0}".format(
+                               document.get_main_folder()))
```
alexfikl, line 150:
```suggestion
                logger.warning("Folder to be removed: '%s'", 
                               document.get_main_folder())
```
? Logging generally uses the `%` formatting.

papis/tui/utils.py
```diff
@@ -1,6 +1,50 @@
 import re
 from typing import Optional, List, Callable, Any
 
+PAPIS_PYGMENTS_DEFAULT_STYLE = {
```
alexfikl, line 4: Are these taken from somewhere? Maybe add a little note explaining why it's like this if possible?

papis/tui/utils.py
```diff
@@ -1,6 +1,50 @@
...
+    "pygments.name.namespace": "bold ansicyan",
+    "pygments.name.exception": "bold ansired",
+    "pygments.name.variable": "ansiblue",
+    "pygments.name.constant": "ansired",
+    "pygments.name.label": "ansigreen",
+    "pygments.name.entity": "bold ansigray",
+    "pygments.name.attribute": "ansigreen",
+    "pygments.name.tag": "bold ansigreen",
+    "pygments.name.decorator": "ansimagenta",
+    # Note: In Pygments, Token.String is an alias for Token.Literal.String,
```
alexfikl, line 25:
```suggestion
    # NOTE: In Pygments, Token.String is an alias for Token.Literal.String,
```
? I think most editors recognize the uppercase NOTE as being special (I think `pylint` even yells about `FIXME` and `TODO`).

alexfikl (approved):
I left a couple of nitpicks, but overall this looks very nice to me! :rocket: 

I tested it in my terminal and it's definitely an improvement color-wise + didn't use the text editing functionality in there, so won't miss it!

papis/tui/widgets/diff.py
jghauser: The rule I go with is that after a colon we should capitalise the next word only when 2 or more full sentences follow (see [here](https://www.grammarbook.com/blog/commas/capitalization-with-colons/) for example), but I definitely don't have strong feelings about it haha

papis/commands/add.py
jghauser: Oh true, I hadn't considered that. Thanks!

papis/tui/utils.py
jghauser: I just used a copy of the default style and changed it to use ansi colors. I'll make a note of it, thanks!

papis/tui/widgets/diff.py
alexfikl: That's good to know, thanks for the link! Then yes, it's just fine as is.

For my curiosity, does that also work for title capitalization? (I had in mind something like [this](https://apastyle.apa.org/style-grammar-guidelines/capitalization/title-case), from a quick search)

jghauser: Good point, it could be considered a title. In that case I guess the question is whether we use (British style) sentence case for them (in which case my initial form would still be right) or we use (US style) title case, in which case we should capitalise even more (e.g. "Diff View: All Changes").

alexfikl: :exploding_head: Yeah, it's completely fine for now, haha! Thank you for explaining.

alexfikl:
@jghauser I've been using this patch for a few weeks now without any issues and the colors much nicer. Should we merge it?

@alejandrogallo Any remaining issues with removing the editing from the `text_area`?

alejandrogallo:
I am for merging too! I have no issues with this, that's great work you two!

alejandrogallo (approved):

jghauser:
Yes, let's do it!

alexfikl:
> Yes, let's do it! Should I squash the commits first or what's the norm here?

Woo! Went ahead and squashed it in!

I don't know if there's any particular norm.. my personal choice has been: if there are a few well-separated commits => merge it, otherwise squash it :confused:

jghauser:
Good choice, those were definitely not all well-separated commits ;)

PR #463, opened 2022-12-24
Add some more helpful targets to Makefile

alexfikl:
This adds 
* a `doc` target that should make the docs in HTML format.
* a `test` target that just runs pytest.
* a `ci-install` and `ci-test` target to install and run the tests like on the ci.

@jghauser hopefully these are helpful!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,9 +37,8 @@ jobs:
         python-version: ${{ matrix.python-version }}
 
     - name: Install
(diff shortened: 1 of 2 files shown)


alejandrogallo:
I missed that, so now the old hacking would have made sense ... ups! X)

alexfikl:
> I missed that, so now the old hacking would have made sense ... ups! X)

Haha, yeah, I missed your commit as well! Sorry for the overlap there :confounded:

PR #464, opened 2022-12-24
Remove LGTM badge

alexfikl:
LGTM went EOL on December 15th, so this removes the badge.

@alejandrogallo The `coveralls` badge also points to something that still uses the TravisCI and hasn't been updated in a long time. Can you update that or should I remove it as well?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -175,14 +175,11 @@ Papis isn't the only fish in the pond. You might also be interested in:
 - `Zotero <https://www.zotero.org/>`__ (opensource)
 
 
-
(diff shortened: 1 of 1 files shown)


alejandrogallo:
I think we can safely remove it and at some point when we work more on test coverage we can put it back in with github ci

alexfikl:
> I think we can safely remove it and at some point when we work more on test coverage we can put it back in with github ci

Removed it! Also added a CodeQL badge to replace the LGTM one, since that's what we're using now.

PR #465, opened 2022-12-25
Use a consistent naming scheme for loggers

alexfikl:
This makes a few changes to how the logging is setup. To summarize a bit:

* adds a new `papis.logging` module that should contain all the logging colors and formatting.
* names all the loggers `__name__`, i.e. based on the module they're in.
* changes the log format to `[LEVEL] submodule.submodule: Some message`. Note that `"papis."` is removed from `__name__` in the actual output.
* adds color coding to the log level in the format, i.e. ERROR is red, INFO is cyan, etc.

All the uses in the source and tests are updated to use this. I've checked a few of the commands in real-life usage and it seems to be working fine, but more testing is definitely welcome!

### DIFF
--- a/papis/__init__.py
+++ b/papis/__init__.py
@@ -1,20 +1,4 @@
-import os
-
-# Information
 __license__ = "GPLv3"
 __version__ = "0.12"
 __author__ = __maintainer__ = "Alejandro Gallo"
 __email__ = "aamsgallo@gmail.com"
-
-
-if os.environ.get("PAPIS_DEBUG"):
-    import logging
-    log_format = (
-        "%(relativeCreated)d-"
-        + "%(levelname)s"
-        + ":"
-        + "%(name)s"
-        + ":"
-        + "%(message)s"
-    )
-    logging.basicConfig(format=log_format, level=logging.DEBUG)
--- a/papis/commands/default.py
+++ b/papis/commands/default.py
@@ -66,6 +66,9 @@ class ColoramaFormatter(logging.Formatter):
                 record.levelname,
                 colorama.Style.RESET_ALL)
 
+        if record.name.startswith("papis."):
+            record.name = record.name[6:]
+
         return super().format(record)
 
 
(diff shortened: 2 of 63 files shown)


papis/commands/default.py
```diff
@@ -231,26 +245,25 @@ def run(verbose: bool,
...
                   + "%(levelname)s"
-                  + ":"
+                  + "] "
                   + colorama.Fore.GREEN
                   + "%(name)s"
                   + colorama.Style.RESET_ALL
-                  + ":"
+                  + " "
                   + "%(message)s"
                   )
```
alexfikl, line 262: I also changed the log format from
```
LEVEL:module:module:Some message over here
```
to
```
[LEVEL] module:module Some message over here
```
i.e. put the level in brackets and add another space before the message. 

What do you think? I'm not super sold on either version, but was looking at it and wanted to bring it up at least. I'll revert it back to the previous version if no one has any preferences for now.

jghauser: I'm wondering why not just keep the ".", rather than transforming it into ":", then we could have the message be `[LEVEL] module.module: Some message over here` which would retain the colon as a separator between the message and the stuff before.

alexfikl: Yeah, that looks good! Thanks for the suggestion!

For reference, default `dmesg` does
```
[timestamp] submodule: Some message
```
and `journalctl` does
```
timestamp host submodule: Some message
``` 
(without any other brackets) so this should be nice and in line with some other common tools.

alexfikl:
@alejandrogallo Any thoughts about this?

alejandrogallo:
is this good to go already ? I think with @jghauser suggestions it has converged into a nice format, it's a good idea! maybe we should wrap it in a papis.logging module in order to enforce in the future this behavior?

alejandrogallo:
but it's not necessary to wrap it, we can also do this in the future

alexfikl:
> I think with @jghauser suggestions it has converged into a nice format

Definitely! :+1: 

> maybe we should wrap it in a papis.logging module in order to enforce in the future this behavior?

Yeah, that sounds like a good idea! It should be pretty straightforward, so I'm definitely up for doing it as part of this PR.

papis/__init__.py
```diff
@@ -1,20 +1,4 @@
-import os
-
-# Information
 __license__ = "GPLv3"
 __version__ = "0.12"
 __author__ = __maintainer__ = "Alejandro Gallo"
 __email__ = "aamsgallo@gmail.com"
-
-
-if os.environ.get("PAPIS_DEBUG"):
```
alexfikl, line 10: Was this used for anything? I might have been a bit hasty in removing it..

My understanding was that the logging gets initialized in `papis.commands.default`, which has flags to enable more verbosity, so it seemed redundant.

alexfikl:
This looks good to go to me now!

papis/__init__.py
alejandrogallo: this was as far as I remember to see the timing and debugging information of importing modules in papis before the default command kicks in, in order to optimize then startup time. but I think this should rather be done better by a profiling tool, so I guess it's ok to remove?

alejandrogallo:
this is great work! it looks very clean and ok by me !

alexfikl:
> this is great work! it looks very clean and ok by me !

@alejandrogallo @jghauser Thanks a lot for taking a look!

PR #466, opened 2022-12-28
Fix a circular import cli->pick->api->pick

alexfikl:
Not quite sure why no other test or linter picks this up, so it's a bit annoying..

EDIT: `pylint` picks it up, but that's a bit noisy, so I don't know if it's worth enabling on the CI.

### DIFF
--- a/papis/pick.py
+++ b/papis/pick.py
@@ -7,7 +7,6 @@ from typing import Callable, TypeVar, Generic, Sequence, Type
 import papis.config
 import papis.document
 import papis.plugin
-import papis.api
 
 logger = logging.getLogger("pick")
 T = TypeVar("T")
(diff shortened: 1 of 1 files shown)


alexfikl:
I'll go ahead and merge this, since it shouldn't be too controversial. I ran `pylint` over everything and it didn't find any other circular imports anymore, so it should be ok!

alejandrogallo:
Thanks, that's a good one!
I try to use also since some time ago pylint in my workflow, but some pylint complaints are just too conservative, I even added a pylint configuration ignoring some stuff.

alexfikl:
> I try to use also since some time ago pylint in my workflow, but some pylint complaints are just too conservative, I even added a pylint configuration ignoring some stuff.

Yes, that's been my experience as well :disappointed: Plus it still seems to be quite active, so every new `pylint` release requires some work.

PR #467, opened 2022-12-28
Fix unicode handling of journals in bibtex exporter

alejandrogallo:
I think there was a small bug in the unicode exporter for journal types.
I also removed the function unicode_to_latex since apparently @alexfikl found the
same behaviour in the bibtexparser package.

Maybe @alexfikl can take a look since you're are the last one that worked on this?
Thanks!!


papis/bibtex.py
```diff
@@ -342,8 +342,6 @@ def to_bibtex(document: papis.document.Document, *, indent: int = 2) -> str:
                     "Key '%s' is not present for ref '%s'",
                     journal_key, document["ref"])
 
-            bib_value = string_to_latex(bib_value)
```
alexfikl, line 345: I added this because I had some journal named `"Something & Something"` and the resulting bib file would be invalid. Is there a better way to fix that?

alexfikl (commented):
Oops, sorry for breaking this! I'm not sure about the fix though, since it would break other documents. Are there some tests for this? Maybe we should add some more testing around the unicode parts.

On the other hand, big :+1: for removing `unicode_to_latex` in favor of the `biblatexparser` version!

papis/bibtex.py
alejandrogallo: I would say ah, I see, good question. I can think of two ways of going around it, since this might happen not only with journal names but also with titles and so on.

I have been thinking about setting an overridable field for an exporter. I.e,
for instance you might have

```yaml
title: Dr. Jêkyll & Mr. Hÿdë
title_bibtex: Dr. Jêkyll \& Mr. Hÿdë
```

and then if you have in your library unicode activated because you use
xelatex or whatever, then the bibtex exporter should prefer
`title_bibtex` over `title` checking if there is a
`(key_name + "_bibtex") in document`.

This is a way of overriding problematic cases, not ideal and potentially
a lot of hand work but at least it's quite special-cases-proof.

The other option is to identify which characters are problematic in the *tex
family when copied /as-is/ to the bibtex. I can think of &$\( ...
So it would be quite a mess to list all options I think.

alejandrogallo:
Yes I agree, good idea we should add some more tests for this.

papis/bibtex.py
alexfikl: Hm, yeah, I found another project trying to escape some basic latex strings
https://github.com/JelteF/PyLaTeX/blob/1a73261b771ae15afbb3ca5f06d4ba61328f1c62/pylatex/utils.py#L14-L30
but not sure that's complete in any way. Your idea seems more general and pretty straightforward, so it sounds good to me!

alejandrogallo: that's interesting, I'll go ahead an implement (and document) the overriding functinality, which is as you say a minimal change and can help some folks get things done quick and dirty ;)

alejandrogallo:
It hink I think this is mostly ready to go, I'll squash some commits and so on. But in summary:

- Add general bibtex documentation
- Add `_latex` prefix for overriding keys

I'm still not certain what happens when there is the & and so on in the unicode allowing fields.
It's certainly a problem if there is a user that has 20% of papers of a journal called
`Journal of Cats & Dogs` and replacing this in all journals, is still easily doable
but kind of not so ergonomic. hmm..

alexfikl (approved):
Left a couple of comments in the documentation section, but the rest looks good to go!

Thanks for adding tests!

alexfikl:
> I'm still not certain what happens when there is the & and so on in the unicode allowing fields. It's certainly a problem if there is a user that has 20% of papers of a journal called `Journal of Cats & Dogs` and replacing this in all journals, is still easily doable but kind of not so ergonomic. hmm..

I have a few dozen SIAM papers where the publisher field is written in full `Society for Industrial & Applied Mathematics` that would hit this. We'll see how annoying it is and make it a bit smarter if needed..

tests/__init__.py
```diff
@@ -109,6 +117,16 @@ def create_real_document(
...
 
+class with_fresh_config(contextlib.ContextDecorator):
+
+    def __enter__(self) -> papis.config.Configuration:
+        config = papis.config.reset_configuration()
+        config["settings"] = {}
+        return config
+
+    def __exit__(self, *exc: Any) -> None:
+        return
```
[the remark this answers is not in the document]
alejandrogallo: Ah yes!, I missed that, I merged both function using your approach of creating a temporary file for the configuration, I think this is cleaner than what I did, thanks a bunch!

alexfikl:
@alejandrogallo I think this has been ready for a while, but I merged in `main`, fixed conflicts and some other small nitpicks. What do you think about merging it in?

alexfikl:
> > I'm still not certain what happens when there is the & and so on in the unicode allowing fields. It's certainly a problem if there is a user that has 20% of papers of a journal called `Journal of Cats & Dogs` and replacing this in all journals, is still easily doable but kind of not so ergonomic. hmm..
> 
> I have a few dozen SIAM papers where the publisher field is written in full `Society for Industrial & Applied Mathematics` that would hit this. We'll see how annoying it is and make it a bit smarter if needed..

Also, since it's been a while here.. we can now add a small `papis doctor` check that automatically adds `<key>_latex` when needed in a future PR :grin:

maxmahlke:
+1 for getting this merged, I'm tired of getting `misplaced alignment tab character &` errors when citing `Journal of Cats & Dogs` :-)

Anything that needs to be done where I can contribute?

alexfikl:
> +1 for getting this merged, I'm tired of getting `misplaced alignment tab character &` errors when citing `Journal of Cats & Dogs` :-)
> 
> Anything that needs to be done where I can contribute?

Agreed! I think we can merge this (the big chunk of the changes just delete the latex-to-unicode stuff).

PR #468, opened 2022-12-28
Add working implementation of bibtex-ignore-keys

alejandrogallo:
I needed a setting to not export the abstract of my big library, so that other programs
have a better time parsing it.

I think this is a useful addition to the bibtex settings, what do you people think?

### DIFF
--- a/doc/source/default-settings.rst
+++ b/doc/source/default-settings.rst
@@ -183,10 +183,10 @@ Bibtex options
 .. papis-config:: bibtex-ignore-keys
    :default: []
 
-  When exporting document to bibtex format, do not export the keys
+  When exporting a document to the BibTeX format, do not export the keys
   appearing in this list. This might be useful if you have some keys
-  that might have a lot of content, such as ``abstract``, or maybe you
-  have used a valid bibtex key for some other purposes, like the ``note``
+  that have a lot of content, such as ``abstract``, or maybe you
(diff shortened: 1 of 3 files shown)


doc/source/default-settings.rst
```diff
@@ -180,6 +180,15 @@ Bibtex options
...
     extra-bibtex-keys = ["tags", "doc_url"]
 
+.. papis-config:: bibtex-ignore-keys
+   :default: []
+
+  When exporting document to bibtex format, do not export the keys
+  appearing in this list. This might be useful if you have some keys
+  that might have a lot of content, such as ``abstract``, or maybe you
+  have used a valid bibtex key for some other purposes, like the ``note``
+  key.
```
alexfikl, line 190:
```suggestion
  When exporting a document to the BibTeX format, do not export the keys
  appearing in this list. This might be useful if you have some keys
  that have a lot of content, such as ``abstract``, or maybe you
  have used a valid BibTeX key for some other purposes, like the ``note``
  key.
```

alexfikl (approved):
Looks good to me!

I keep deleting the abstract from my bib files too, so this is definitely useful.

jghauser:
This is neat and makes it much easier for me to get rid of `doi`s,  which I tend to have little use for in my bibliographies :)

alejandrogallo:
great !

PR #472, opened 2023-01-01
Update user-facing API docs

alexfikl:
Was looking at the functions in `api.py` and did some small drive-by formatting and phrasing changes to the docs.

The biggest change is probably adding proper wrappers around `pick` and `pick_doc` with full docs.


alejandrogallo:
that's useful, thanks!

PR #476, opened 2023-01-09
Tell `subprocess.Popen` to disown the process. Fixes #475.

### LINKED ISSUE #475: Detach the child process in `papis open`
* papis version (`$ papis --version` or commit number): papis, version 0.12

# how to reproduce the issue
After opening a document within the terminal using
```
papis open <document_identifier>
```
the terminal cannot be closed as it waits for the sub-process, e.g. `xdg-open`, to finish.

# resolve the issue

## On Linux
Using the `nohup` tool and redirecting the `stdout` stream to `/dev/null` should resolve the issue.

## On Windows
Adding the following parameter to the `subprocess.Popen` constructor should resolve the issue.
```
creationflags=subprocess.DETACHED_PROCESS | subprocess.CREATE_NEW_PROCESS_GROUP
```

cheinigk:
Fixes the issue #475.

### DIFF
--- a/papis/utils.py
+++ b/papis/utils.py
@@ -65,6 +65,11 @@ def general_open(file_name: str,
     cmd = shlex.split("{0} '{1}'".format(opener, file_name))
     logger.debug("cmd: %s", cmd)
 
+    # Check that opener can be called.
+    import shutil
+    if not shutil.which(cmd[0]):
+        raise FileNotFoundError("{} not found".format(opener))
+
     import subprocess
     if wait:
         logger.debug("Waiting for process to finish")
--- a/papis/utils.py
+++ b/papis/utils.py
@@ -77,7 +77,7 @@ def general_open(file_name: str,
             "stdout": None,
             "stderr": None,
             "close_fds": True
-        }
+        }	# type: Dict[str, Any]
         # Tell subprocess to detach the process.
         if sys.platform == "win32":
             popen_kwargs["creationflags"] = subprocess.DETACHED_PROCESS
--- a/papis/utils.py
+++ b/papis/utils.py
@@ -65,10 +65,12 @@ def general_open(file_name: str,
     cmd = shlex.split("{0} '{1}'".format(opener, file_name))
     logger.debug("cmd: %s", cmd)
 
-    # Check that opener can be called.
+    # NOTE: Detached processes do not fail properly when the command does not
+    # exist, so we check for it manually here
     import shutil
     if not shutil.which(cmd[0]):
-        raise FileNotFoundError("{} not found".format(opener))
+        raise FileNotFoundError(
+            "[Errno 2] No such file or directory: '{}'".format(opener))
 
     import subprocess
     if wait:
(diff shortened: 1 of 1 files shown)


papis/utils.py
```diff
@@ -71,9 +71,22 @@ def general_open(file_name: str,
...
+            "shell": False,
+            "stdin": None,
+            "stdout": None,
+            "stderr": None,
+            "close_fds": True
+        }
+        # Tell subprocess to detach the process.
+        if sys.platform == "win32":
+            popen_kwargs["creationflags"] = subprocess.DETACHED_PROCESS
+            popen_kwargs["creationflags"] |= subprocess.CREATE_NEW_PROCESS_GROUP
```
alexfikl, line 84: Why are these Windows only? Does it not apply to Unix shells?

papis/utils.py
```diff
@@ -71,9 +71,22 @@ def general_open(file_name: str,
...
+        else:
+            cmd.insert(0, "nohup")
+            popen_kwargs["stdout"] = subprocess.DEVNULL
+            popen_kwargs["stderr"] = subprocess.DEVNULL
```
alexfikl, line 88: Why set these to `DEVNULL` only here?

alexfikl (commented):
Thank you for the PR!
I had some questions on the special casing of the OS and then we can get it in.

papis/utils.py
alexfikl: Ah, just noticed that these are Windows-only constants. Nevermind!

papis/utils.py
```diff
@@ -71,9 +71,22 @@ def general_open(file_name: str,
...
-        subprocess.Popen(
-            cmd, shell=False,
-            stdin=None, stdout=None, stderr=None, close_fds=True)
+        popen_kwargs = {
+            "shell": False,
+            "stdin": None,
+            "stdout": None,
+            "stderr": None,
+            "close_fds": True
+        }
```
alexfikl, line 80:
```suggestion
        }	# type: Dict[str, Any]
```
You'll need to add that to make `mypy` happy.

papis/utils.py
cheinigk: Exactly, no worries :-)

papis/utils.py
cheinigk: I thought they are Unix only, but [the doc](https://docs.python.org/3/library/os.html#os.devnull) says, it's None on Windows. I'll put it to the default popen_kwargs.

alexfikl:
Any idea why the test is failing? For some reason this call doesn't return an error code even though the command does not exist:
```
        result = self.invoke([
            "Krishnamurti",
            "--tool", "nonexistingcommand"
        ])
>       self.assertNotEqual(result.exit_code, 0)
```
Is it because of the `nohup`? Does that not return a proper error code for some reason?

alexfikl:
> I think the problem is with the combination of `nohup` and running it in the background using `subprocess.Popen`. It's not possible to query it for a return code. The current situation is, that `Popen` raises an error if the command does not exist, but `nohup` exists on Linux. I've introduced an explicit check for the opener now.

Awesome! Thanks a lot for all your work here! I'll go ahead and merge this :rocket:

cheinigk:
Thanks for your help with the PR and for merging!

PR #478, opened 2023-01-12
Fix shell completion

### LINKED ISSUE #399: Tab Completion Not Working in 0.12
* papis version (`$ papis --version` or commit number): 0.12

# how to reproduce the issue
In both `zsh` and `bash` I cannot get tab completion to work properly with `papis`. Typing `papis` and tab completing provides absolutely nothing in `zsh` (similar to issue #303 but it happens with just `papis` as well as `papis add`), and doing the same in `bash` only completes with the files in that directory, which I think is the default behavior in `bash`.

The documentation for tab completion gives a couple of scripts to make completion work, but clicking on them gives a 400 error, so I am unable to test if there is a solution for this.

I am running Arch Linux and installed `papis` through the AUR.

alexfikl:
This is an attempt to fix the shell completion for papis. It's completely based off of the [support in click](https://click.palletsprojects.com/en/8.1.x/shell-completion/). It updates
* `zsh`: basically regenerates the completion script with the latest `click` version.
* `bash`: removes the previous handwritten code in favor of a `click` version. Haven't looked too carefully at whether the old version did more, but since it seems unmaintained, this is probably a better choice for now.
* `fish`: `click` has support for this too, so it's added!

I tested the `zsh` and `bash` completions and they seem to work as expected. I don't use `fish`, so not sure about that, but we can fix it when someone complains :grin:

If anyone makes use of this, some more testing would be highly appreciated!

Fixes #399.


alexfikl:
TODO: There are mentions of the completion stuff in the docs, which need updating before this goes in!

doc/source/shell_completion.rst
```diff
@@ -1,34 +1,41 @@
...
-::
+* ``bash``: the completion script is installed in ``$PREFIX/etc/bash_completion.d``
+  and can be sourced manually using (or added to your ``.bashrc``)
 
-  PREFIX/etc/bash_completion.d/papis
+.. code:: bash
 
-normally the ``PREFIX`` part is ``/usr/local/``, so you can add the
-following line to your ``~/.bashrc`` file
+    source $PREFIX/etc/bash_completion.d/papis.sh
```
marsam, line 16: I think, it could be better to install it in `/share/bash-completion/completions` because it's the path used by bash-completions, and it's supported by Linux distributions
```suggestion
    source $PREFIX/share/bash-completion/completions/papis
```

doc/source/shell_completion.rst
```diff
@@ -1,34 +1,41 @@
...
-::
+* ``fish`` the completion script is installed in ``$PREFIX/etc/fish/completions``
+  and can be sourced manually (or added to your ``config.fish``)
 
-  source /usr/local/etc/bash_completion.d/papis
+.. code:: bash
 
-or get the bash script from
-`here <https://raw.githubusercontent.com/alejandrogallo/papis/master/scripts/shell_completion/click/papis.sh/>`__.
+   source $PREFIX/etc/fish/completions/papis.fish
```
marsam, line 23:
```suggestion
   source $PREFIX/share/fish/vendor_completions.d/papis.fish
```

doc/source/shell_completion.rst
alexfikl: Updated both here and in `setup.py`.
https://github.com/papis/papis/compare/25da9438ce1d38605295247de831d09939f0db98..b52d04d886cbf604c73ecc5457061d1017c6b85b

doc/source/shell_completion.rst
alexfikl: Updated both here and in `setup.py`.
https://github.com/papis/papis/compare/25da9438ce1d38605295247de831d09939f0db98..b52d04d886cbf604c73ecc5457061d1017c6b85b

alexfikl:
@marsam I was hoping someone more knowledgeable would drop by and take a look, so your review is very appreciated! 

It should all be updated now and it seems to still work on my machine (still only tested `bash` and `zsh`). Let me know if you spot anything else!

marsam:
I think codacy is failing because `exclude_paths` has the prefix `/`, which is unnecessary 
https://github.com/papis/papis/blob/47fd47b9d630356cddf4e61e5dc181c72bb8997b/.codacy.yml#L9
i.e.
```diff
- - "/scripts/shell_completion/**"
+ - "scripts/shell_completion/**"
```

marsam (approved):

alexfikl:
> I think codacy is failing because `exclude_paths` has the prefix `/`, which is unnecessary

Yep, nice catch! Updated the `codacy` paths in there, which seemed to fix that. :rocket:

PR #479, opened 2023-01-13
Enable python 3.11 tests on windows

alexfikl:


alexfikl:
This may need the latest version of `lxml`, but it seems to work, so in it goes!

PR #485, opened 2023-01-31
Use singleton requests.Session object

alexfikl:
This adds a `papis.utils.get_session` that returns a `requests.Session` object to be used everywhere with the user-agent and proxies set by the user. This also replaced all the `urllib` usage.

Benefits: `requests` API is nicer; headers and proxies set consistently; some savings on multiple requests?
Downsides: ?

### DIFF
--- a/papis/arxiv.py
+++ b/papis/arxiv.py
@@ -22,10 +22,11 @@ from typing import Optional, List, Dict, Any
 
 import click
 
-import papis.filetype
-import papis.downloaders.base
 import papis.config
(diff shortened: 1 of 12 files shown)


alexfikl:
Initially made the session a singleton, but that doesn't seem like a good idea:
* a bit of a pain to close the session at the end.
* sessions hold on to cookies, which we don't necessarily want between services.

Since this is just now a refactor, I'll probably merge it when the tests pass! :rocket:

PR #489, opened 2023-02-10
Add DBLP support

### LINKED ISSUE #337: DBLP (computer science bibliography) Integration
Hi,

Would it be possible to support DBLP keys and links for adding papers? This seems like a straight forward integration.

alexfikl:
This adds support for the DBLP API with

* an explorer, which can be used with `papis explore dblp -q query pick`.
* an importer, which can be used with `papis add --from dblp 'https://dblp.org/rec/some/dblp/key'`.
* a helper `papis.dblp.search` function, which will probably stay internal

It has some tests and I've played with it a bit too, but I'll leave it here for a while in case anyone else wants to go it a go!

Fixes #337


alejandrogallo:
This is great thank you Alex, with it and the remaining explorers I'd like to implement also an explorer in the webapp, that would be quite cool to be able to have results from dblp and crossref, this looks by the way good!

alexfikl:
> This is great thank you Alex, with it and the remaining explorers I'd like to implement also an explorer in the webapp, that would be quite cool to be able to have results from dblp and crossref, this looks by the way good!

That would be really cool! Let me know if you get to it and something doesn't work as expected.

PR #490, opened 2023-02-12
Remove BASE and ISBNPlus support

alexfikl:
These services seem to be quite unmaintained, so we could remove them. @alejandrogallo What do you think?

* **BASE**: they seem to [require manually whitelisting](https://api.base-search.net/) individual IP addresses to gain access to the API, which doesn't seem particularly scalable for papis users. I contacted them to whitelist my IP so I can check if this still works, but haven't heard back yet.
* **ISBNPlus**: seems to be dead (not sure if this is the website: https://isbnplus.3scale.net/), but the API endpoints in the code don't work anymore (they return `403 Forbidden` or `503 Service Unavailable`). Tried signing up and creating a new key with no luck.

The currently working explorers seem to be: `arxiv`, `dissemin`, `dblp` and `crossref`.


alejandrogallo:
I agree we can erase both, isbnplus was ok when it lived, but it wasn't actually giving good results normally.
So we should rather have less and working than more and crappy. Thank you Alex

alexfikl:
> I agree we can erase both, isbnplus was ok when it lived, but it wasn't actually giving good results normally. So we should rather have less and working than more and crappy. Thank you Alex

Yep, I completely agree! Thanks for taking a look.

PR #493, opened 2023-02-13
Improve external script logging

alexfikl:
This improves the logging setup for scripts loaded through `papis.commands.external`, which before didn't pick up the settings passed to the command line.

This is done by:
* exporting some environment variables in `papis.commands.external`: `PAPIS_LOG_LEVEL`, `PAPIS_LOG_COLOR`, `PAPIS_LOG_FILE` and `PAPIS_DEBUG`.
* Having `papis.logging.setup()` take default values from the environ if they exist.
* To make everything consistent, `papis.commands.default` also takes the defaults from the environment if they exist.

Now a script just needs to add an empty `papis.logging.setup()` and it should just work™ :grin:

### DIFF
--- a/examples/scripts/papis-abbrev
+++ b/examples/scripts/papis-abbrev
@@ -49,7 +49,7 @@ import papis.api
 import papis.document
 import papis.logging
 
-papis.logging.setup("INFO")
+papis.logging.setup()
 logger = papis.logging.get_logger("papis.commands.abbrev")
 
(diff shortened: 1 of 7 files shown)


alejandrogallo:
No, I think this covers really all cases that anyone will encounter, awesome!

alexfikl:
> No, I think this covers really all cases that anyone will encounter, awesome!

Thanks!

PR #495, opened 2023-02-15
Doctor: add check for HTML tags

alexfikl:
Had a lot of abstracts that contained some weird HTML / XML tags in them, so this adds a little check similar to `html-codes` to clean those up.

### DIFF
--- a/papis/commands/doctor.py
+++ b/papis/commands/doctor.py
@@ -358,12 +358,65 @@ def html_codes_check(doc: papis.document.Document) -> List[Error]:
     return results
 
 
+HTML_TAGS_CHECK_NAME = "html-tags"
+HTML_TAGS_REGEX = re.compile(r"<.*?>")
+
+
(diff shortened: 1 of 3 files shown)


alexfikl:
@alejandrogallo @jghauser I was going to merge this since it's pretty straight forward, but wanted to ask. Did you have any restrictions in mind for what checks we should allow in here?

jghauser:
This looks useful! I've definitely had such tags polluting my abstracts at times. I'm guessing it could lead to some false positives at times, but given that `papis doctor` is a manual thing, that's not much of a problem. I think you can merge this :)

alexfikl:
> This looks useful! I've definitely had such tags polluting my abstracts at times. I'm guessing it could lead to some false positives at times, but given that `papis doctor` is a manual thing, that's not much of a problem. I think you can merge this :)

Thanks for taking a look! Yeah, there are probably some cases where the tags are desired, but this is manual and not turned on by default (in `default-doctor-checks`), so it should be fine.

PR #498, opened 2023-02-16
Add needed typing_extensions dependency

twrightsman:
Installing `papis` from master to a venv and running it raises an error due to a missing `typing_extensions` library. I've added it to `setup.py` so it gets installed with `papis` through `pip`. Open to feedback!


alexfikl:
Nice catch! In it goes! :rocket:

PR #499, opened 2023-02-16
Fix cache database query for jinja2

twrightsman:
Following the discussion and ideas in https://github.com/papis/papis/pull/384, I've created a new potential fix for querying with the cache database and the `jinja2` formatter. This should work regardless of the actual formatter used.

### DIFF
--- a/papis/database/cache.py
+++ b/papis/database/cache.py
@@ -100,13 +100,9 @@ def match_document(
         doc_key: Optional[str] = None) -> Optional[Match[str]]:
     """Main function to match document to a given search.
 
-    :param document: Papis document
-    :param search: A valid search string
-    :param match_format: Python-like format string.
-        (`see here
-        <https://docs.python.org/2/library/string.html#format-string-syntax>`__)
-    :param doc_key: Restrict the search to an optionally given document key
-    :returns: Non false if matches, true-ish if it does match.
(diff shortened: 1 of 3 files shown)


alexfikl:
This looks good to me, but it may need some updates in the tests.

How hard would it be to add a little test case to check that this works as expected for both formatters?

twrightsman:
I can definitely update the tests and possibly write a new case.

alexfikl:
> I can definitely update the tests and possibly write a new case.

Sounds good! Let me know if it's too much trouble and I can take a look.

twrightsman:
I made a couple of small changes to fix the tests. I noticed a lot of tests fail when using the `jinja2` formatter, so I think rather than writing a new test here, it's maybe better served to fix tests failing with `jinja2` in another PR?

alexfikl:
> I made a couple of small changes to fix the tests. I noticed a lot of tests fail when using the `jinja2` formatter, so I think rather than writing a new test here, it's maybe better to fix tests failing with `jinja2` in another PR?

Hopefully a good chunk of the `jinja2` tests are failing because the defaults in `papis/defaults.py` are geared towards the `python` formatter. You're right though, that would require a bigger sweep to check properly, so better left for later.

Thanks for your work on this!

PR #503, opened 2023-02-18
Minor DocMatcher optimizations

alexfikl:
Was profiling some more and found some places to improve in `DocMatcher`. 

The main changes are
* store `match-format` in the `DocMatcher` class and pass it to `matcher` by default, so that `papis.config.get` does not get called in `match_document` for every `document * search`.
* post-process results from `pyparsing` into a simpler list that can be used directly in `return_if_match`, to avoid the ifs.
* store the compiled regex when post-processing, instead of recompiling it in `match_document`.

All in all, this seems to have saved another whooping 10ms-ish on my machine, bringing a call to `papis list` with about 50 / 1200 documents down to about 170ms-ish. The rest of the time is spent loading up plugins / scripts / click.

### DIFF
--- a/papis/database/cache.py
+++ b/papis/database/cache.py
@@ -100,13 +100,7 @@ def match_document(
         doc_key: Optional[str] = None) -> Optional[Match[str]]:
     """Match a document's keys to a given search pattern.
 
-    The search pattern is matched against *doc_key*, if given, and *match_format*
-    otherwise.
-
-    :param search: A regex pattern to match the query against..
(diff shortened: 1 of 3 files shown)


alexfikl:
Tested this with `jinja2` it it still seems to work. I.e. these two give the same result
```
papis --set formater jinja2 --set match-format '{{doc.author}}' -l papers list --sort year -a tom
papis --set formater jinja2 -l papers list --sort year -a 'author:tom'
```

PR #505, opened 2023-02-19
Add flag to disable file downloads

alexfikl:
This is an improvement to #387 that tries to avoid downloading files all together (and all the additional traffic!) when the right conditions are met. For this, it adds:
* a `only_data` flag to `get_matching_importer_or_downloader`
* a `only_data` flag to `get_matching_importer_by_name` (new function in `papis.utils`)

so that they only call `Importer.fetch_files` if asked. It updates `papis add` and `papis update` to use these new functions.

It probably needs a bit more testing, but it seems pretty clean. @alejandrogallo What do you think?

### DIFF
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -580,57 +580,30 @@ def cli(files: List[str],
         confirm = False
         open_file = False
 
-    import_mgr = papis.importer.get_import_mgr()
-    matching_importers = []  # type: List[papis.importer.Importer]
+    only_data = bool(files) and not force_download
+    matching_importers = papis.utils.get_matching_importer_by_name(
+        from_importer, only_data=only_data)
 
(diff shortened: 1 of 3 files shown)


alexfikl:
Hm, this doesn't actually work as expected because a lot of importers just implement `fetch` in one go instead of `fetch_data` and `fetch_files` separately.

It's probably a good idea to clean that up, so maybe in some future PRs.

alexfikl:
This is pretty much just a refactor at this point, since it relies on the importers to properly split `fetch_data` and `fetch_files`, so going to merge it! :rocket:

Twix53791:
I did'nt find a flag to set this option when adding entries ? I missed something? So I added the flag myself. `papis add --only-data` will disable file download. --force-download takes priority over --only-data.

PR #510, opened 2023-02-21
commands.add: handle duplicate doc with existing folder

### LINKED ISSUE #509: Better handling of duplication warnings
As a follow up to the discussion in #506.

I've just looked at the problem and it's actually the following:
1) When trying to add a duplicate entry, the warning message is displayed *above* the printing of all sorts of information about the duplicate entry. This means that the duplication warning is often off screen. At the bottom it then simply says "Do you want to add the new document? (Y/n):" which doesn't make it obvious that we're dealing with a duplication. This meant I've sometimes tried to add the document anyway, which leads to:
2) An error :D.  There's a python trace that ends with "Exception: There is already a document in /home/julian/Documents/library/main/quinn-1998-emergence-of-object-representations, please check it," Given that this always fails, we might not want to offer a "Y/n" in the first place. Or maybe if the user selects "Y", we should ensure that paths don't clash. (Note also that the error should not end in a comma.)

alexfikl:
This improves the handling when a duplicate document is found like suggested in #509:
* The `'Duplicate Warning...'` log message is moved after the dump of the existing document, so that it sits right above the `'Add? Y/n'` prompt.
* If the user answered yes and the folder already exists, it tries `folder-a`, `folder-b`, etc. until a new one is found and uses that for the new (duplicate) document.

@jghauser What do you think of this?

Fixes #509.

### DIFF
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -244,6 +244,20 @@ def get_hash_folder(data: Dict[str, Any], document_paths: List[str]) -> str:
     return result
 
 
+def ensure_new_folder(path: str) -> str:
+    if not os.path.exists(path):
+        return path
+
+    from string import ascii_lowercase
+    suffix = papis.utils.create_identifier(ascii_lowercase)
+
+    new_path = path
+    while os.path.exists(new_path):
+        new_path = "{}-{}".format(path, next(suffix))
+
+    return new_path
+
+
 def run(paths: List[str],
         data: Optional[Dict[str, Any]] = None,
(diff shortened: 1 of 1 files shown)


jghauser:
> @jghauser What do you think of this?

I think this is perfect, exactly what a user would expect! :)

Thanks for the super quick fix!

alexfikl:
> I think this is perfect, exactly what a user would expect! :)
> 
> Thanks for the super quick fix!

Thanks for taking a look! In it goes then :rocket:

PR #515, opened 2023-02-25
Simplify mypy configuration

alexfikl:
This tries to simplify the `mypy` configuration a bit.

* removes some unnecessary ignores modules that have typing annotations now.
* use explicit `strict = True` in the config instead of listing the options. This may not be a great idea because the enabled options are not stable.


alexfikl:
There seems to be a `types-colorama` as well, but for some reason it generates errors when using format strings:
https://github.com/papis/papis/actions/runs/4270898779/jobs/7434996674#step:5:276
Not quite sure what that's about :(

PR #517, opened 2023-02-27
Improve logging and error messages

alexfikl:
Went through the logging messages an tried to improve the messages a bit. Mainly
* Made most messages into proper sentences.
* Made sentences sound more action-y, i.e. `Found 5 documents` vs `5 documents found`.
* Used better exceptions than a bare `Exception` in some places.
* Use the documented `exc_info` to add exception information to a logging message.
* The full traceback for an exception is only printed when `--log DEBUG`, otherwise a simpler message is added from the `exc_info`. It now prints
```
[ERROR] commands.list: An error has occurred! (Caught exception 'ValueError: OMG!'. Use `--log DEBUG` to see traceback)
```
and when using `--log DEBUG`
```
[ERROR] commands.list: An error has occurred!
┆ Traceback (most recent call last):
┆   File "/mnt/data/code/projects/papis/papis/papis/commands/list.py", line 228, in cli
┆     raise ValueError("OMG!")
┆ ValueError: OMG!
```
while before it was always 
```
[ERROR] commands.list: An error has occurred!
  File "/mnt/data/code/projects/papis/papis/papis/commands/list.py", line 228, in cli
    raise ValueError("OMG!")
ValueError: OMG!
```

All of this should be cosmetic changes, so hopefully no big deal. @jghauser Would appreciate your thoughts, if you have the time!

### DIFF
--- a/papis/arxiv.py
+++ b/papis/arxiv.py
@@ -64,7 +64,7 @@ def get_data(
     search_query = "+AND+".join(
         ["{}:{}".format(key, clean_params[key]) for key in clean_params]
     )
-    logger.debug("query = '%s'", search_query)
+    logger.debug("Performing query: '%s'.", search_query)
 
     with papis.utils.get_session() as session:
         response = session.get(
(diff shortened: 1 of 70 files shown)


alexfikl:
I'll go ahead an merge this, since it has very few behavior changes (just in the output), but it's hard to keep in sync because of frequent conflicts :cry:

PR #522, opened 2023-03-08
Improve support for DjVu files in filetype

alexfikl:
This extends the support from `filetype` to also include DjVu files. Unfortunately upstream seems to be pretty slow (by looking at their issues / pull requests), so this should be a quicker way to go.

Added some more functions there to wrap `filetype`:
* `guess_content_extension`
* `guess_document_extension`
that work the same, but return `None` if they can't figure it out.

### DIFF
--- a/papis/crossref.py
+++ b/papis/crossref.py
@@ -405,22 +405,21 @@ class Importer(papis.importer.Importer):
 
         self.logger.info("Trying to download document from '%s'", doc_url)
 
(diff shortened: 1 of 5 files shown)


alexfikl:
Some note for the future: python also has the `mimetypes` module
https://docs.python.org/3.11/library/mimetypes.html
but that doesn't seem to look at the magic numbers in the file.

PR #524, opened 2023-03-11
Use arxiv.py instead of arxiv2bib

alexfikl:
This switches the arxiv code to use [arxiv.py](https://github.com/lukasschwab/arxiv.py). Some benefits:
* it seems better maintained ([arxiv2bib](https://github.com/nathangrigg/arxiv2bib) was last updated in 2017),
* it allows removing our custom `bs4` based parsing,
* the API generally seems nicer.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -17,29 +17,52 @@
       pypkgs = pkgs.python310Packages;
       lib = pkgs.lib;
 
-      runtime_py_deps = with pypkgs; [
-        pyyaml
-        arxiv2bib
-        beautifulsoup4
-        bibtexparser
-        chardet
-        click
-        colorama
-        dominate
-        filetype
-        habanero
-        isbnlib
-        lxml
-        prompt_toolkit
-        pygments
-        pyparsing
-        python-doi
-        python-slugify
-        requests
-        stevedore
-        tqdm
-        whoosh
-      ];
+      runtime_py_deps = let
+        # manually package arxiv.py since it is not yet in upstream nixpkgs
+        arxivpy = python.pkgs.buildPythonPackage rec {
+          pname = "arxiv";
+          version = "1.4.8";
+
+          src = python.pkgs.fetchPypi {
+            inherit pname version;
+            sha256 = "sha256-KoGOp0nqpipuJPwx1Tt2m00z/1XPxd2nx7fTCaOyk3M=";
+          };
+
+          doCheck = false;
+          checkInputs = [ ];
+          propagatedBuildInputs = [pypkgs.feedparser];
+
+          meta = with lib; {
+            homepage = "https://github.com/lukasschwab/arxiv.py";
+            description = " Python wrapper for the arXiv API ";
+            license = licenses.mit;
+          };
+        };
+      in
+        with pypkgs; [
+          pyyaml
+          arxiv2bib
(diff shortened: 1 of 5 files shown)


alexfikl:
This needs python >=3.6 which was required with arxiv.py >=1.0.0 (the older versions have a different API).

alexfikl:
@alejandrogallo @jghauser I've been using this for a while and it seems to be working nicely, but I've been holding off on merging due to the changed dependency. What do you think?

jghauser:
I don't use arxiv and wouldn't know what to look for in testing it. So if you guys think this is working well, I think we can merge it :)

alexfikl:
> I don't use arxiv and wouldn't know what to look for in testing it. So if you guys think this is working well, I think we can merge it :)

Thanks for taking a look!

I only use it with `papis add --from arxiv 'URL'`, but this also updates the `papis explore` plugin. Otherwise, it doesn't really change that much :grin:

alexfikl:
@f0rki @jghauser  I was about to merge this, but noticed `flake.nix` still has `arxiv2bib` and `arxiv.py` doesn't seem to be packaged according to [their package listing](https://search.nixos.org/packages). What would be a good way to continue? Make a package request at NixOS/nixpkgs?

f0rki:
Ok so a bit of background: the current `flake.nix` uses a hacky workaround for setting up dev dependencies: it creates a local virtualenv and adds a shell hook to activate it. However, this only works for `nix develop`. So if we install the flake as regular nix package we miss all the pip-only dependencies. Installing from pip is quite painful in nix flakes and I haven't found a good solution. 

So, yes we should definitely get this packaged in upstream nixpkgs.

jghauser:
@f0rki, thanks for the input! Do you want to submit the package to nixpkgs? I can give it a try too. I'm pretty new to nix, but `arxiv.py` seems not too complex to package. I should be able to do it over the weekend.

Out of curiosity, why did you go with the hacky way of setting up a dev environment with `nix develop`? I've been playing around with the flake recently (mostly to add a dev container), and I'm wondering if things could be improved by doing it the non-hacky way.

f0rki:
Afaik there is currently no clean way to fetch dependencies from pip/pypi in a nix flake (I think dream2nix is what is currently developed to do so. There is also poetry2nix, but papis doesn't use poetry yet) Other than that you would have to package all python dependencies yourself and those were quite a lot. For example, the typing shims for mypy are all not packaged in nix. Also: packaging for nix requires to add hard coded hashes to ensure deterministic builds. I didn't want to do that for the dev environment, as it would add a bunch of maintenance overhead.

f0rki:
OK easy enough to package arixv.py for nix. We can package it inside of the `flake.nix` for now and I will try to get that upstream.

@alexfikl here are the commits https://github.com/alexfikl/papis/pull/1

alexfikl:
> OK easy enough to package arixv.py for nix. We can package it inside of the `flake.nix` for now and I will try to get that upstream.
> 
> @alexfikl here are the commits [alexfikl#1](https://github.com/alexfikl/papis/pull/1)

Thank you for taking care of this so quickly! (and @jghauser of course :heart:)
This should be good to go now, so once the CI passes, I'll probably merge it in.

PR #534, opened 2023-03-22
Add a Developer API section to the docs

alexfikl:
I was cleanup up some random docs lately (e.g. #532) so this adds them to the actual rendered html to make sure that everything is nice and pretty. It currently adds `papis.document`, `papis.importer`, `papis.utils` and `papis.yaml` with a big warning that these are dev docs. 

@jghauser I'm not sure where you left off with the doc re-organization, so let me know if this conflicts too much with what you had in mind. It looks something like this

Screenshot_20230322_210600

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -42,6 +42,9 @@ extensions = [
 
 intersphinx_mapping = {
     "https://docs.python.org/3/": None,
+    "https://requests.readthedocs.io/en/latest/": None,
+    "https://bibtexparser.readthedocs.io/en/master": None,
+    "https://docs.openstack.org/stevedore/latest/": None,
 }
 
 autodoc_member_order = "bysource"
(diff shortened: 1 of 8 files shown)


alexfikl:
Also, it's all the way down in the left-hand side TOC:

Screenshot_20230322_211527

jghauser (approved):

jghauser:
This looks super nice and I don't think it clashes in any major way with what I had in mind. And whatever changes might be needed later to make everything cohere can easily be done when we get there :).

alexfikl:
> This looks super nice and I don't think it clashes in any major way with what I had in mind. And whatever changes might be needed later to make everything cohere can easily be done when we get there :).

Yep, these should be pretty easy to move around.

Thanks for taking a look! :rocket:

PR #536, opened 2023-03-29
Improve testing infrastructure

alexfikl:
This does a major shakedown of the tests! :rocket: 

On the infrastructure side, it consolidates all the helpers into `tests/testlib.py`:
- `TemporaryConfiguration`: a context manager to create a clean empty configuration for a test.
- `TemporaryLibrary`: a context manager to create a new temporary library for a test.
- `ResourceCache`: a little helper (that used to be in `tests/downloads`) that can handle getting remote resources.
- `create_random_file`: same as the previous `create_random_pdf` and friends, but just one function that also randomizes the extension.

All of these are also exposed as [pytest fixtures](https://docs.pytest.org/en/stable/explanation/fixtures.html#about-fixtures) for ease of use. I'm thinking it might be a good idea to add these to a `papis.testing` module so that external scripts and plugins can also make use of them.

On the test side
* all tests use the pytest fixtures consistently.
* tests are ported away from `unittest` class type tests.
* any commented out tests are re-instated and updated.

On benefit of using `TemporaryConfiguration` and friends is that the user session is not modified due to running the tests. Before, the tests would read `~/.config/papis/config` and fail if settings were incompatible, they would add libs to `~/.cache/papis/database`, they would not clean up temporary files, etc. All of this is now a lot better sandboxed! :heart:


alexfikl:
I realize this is a pretty big PR with a lot of changes, so for a review I would recommend looking at
* what's in `tests/testlib.py`
* how the `pytest` fixtures look in `tests/conftest.py`
* how they're used in any of the tests, e.g. `tests/commands/test_add.py`.

Most of the tests are not modified, so likely not worth being too thorough there. Any review and suggestions are very welcome here! The main goal was to make tests easier to write and run for contributors, so any improvements in that direction would be awesome :heart:

alexfikl:
I'll go ahead and merge this, since it's generally an improvement. We can fix any issues in future PRs! :rocket:

PR #540, opened 2023-04-18
docs: Fix missing whitespaces in CLI help texts

leinher:
Found this typo in `papis add --help`:
```
--link / --no-link        Instead of copying the file to the library, create
                            a link toits original location
```

This missing whitespace is a result of code formatting to keep line width.

A quick regex search resulted in some more strings added the same way.


alexfikl:
Nice catch! This looks great! :rocket:

PR #544, opened 2023-04-20
cli: fix default value for --reverse

alexfikl:
Not super sure what's going on here, but this doesn't work as expected, e.g.
```
papis --set sort-reverse true list --sort year <query>
[descending]
papis --set sort-reverse true list --sort year --reverse <query>
[ascending]
papis --set sort-reverse false list --reverse <query>
[ascending]
```
They're all supposed to be ordered descending. Will need to investigate a bit why `click` doesn't like it.

### DIFF
--- a/papis/cli.py
+++ b/papis/cli.py
@@ -40,11 +40,12 @@ def sort_option(**attrs: Any) -> DecoratorCallable:
             help="Sort documents with respect to the FIELD",
             metavar="FIELD",
             **attrs)
+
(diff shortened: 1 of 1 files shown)


alexfikl:
Hm, actually never mind, The issue seems to be that this doesn't work with
```
papis --set sort-reverse true list --sort --reverse <query>
```
while it works fine from the config file. Maybe worth investigating, but not a biggie..

PR #541, opened 2023-04-18
Allow for http URLs where filepaths are required (`add --from bibtex` and `addto`)

f0rki:
Allows for two usecases:
* `papis addto -u https://example.com/paper.pdf paper`
* `papis add --from bibtex https://example.com/paper.bib`

PR #548, opened 2023-05-06
downloaders: improve extension handling

alexfikl:
This tries to add extensions to files downloaded by an importer or downloader.

The motivation is basically that they get opened using `open_file(filename)` at some point and the `opentool` does not necessarily know what type of file it is (not all tools look at the signature). To make this work:
* update the `Downloader.fetch_files` to add extensions to the temporary files.
* update the usage in `papis addto` that supports urls since #541.
* update the usage in `papis add --from bibtex` that also supports urls since #541.

For the downloaders, ran
```
papers --set opentool rifle add --from arxiv 'https://arxiv.org/abs/2305.00377' 
```
which correctly downloads a file with a `pdf` extension and opens it with my PDF viewer through `rifle`, which normally does not recognize the file.


alexfikl:
@ExpandingMan This should add extensions to the files supported by [`filetype`](https://github.com/h2non/filetype.py) and hopefully fix one of your issues. Can you give it a try and see if it works for your use case? I've only tried it with arXiv.

PR #553, opened 2023-05-09
Update typing annotations

alexfikl:
This removes all the comment style annotations `# type: List[str]` in favor of the direct annotations that are available in python 3.6+.

As a plus, we can enable to latest `flake8` too!


alexfikl:
> I agree that updating typing information will be probably better for the future for static analyzers, so let's go! Thank you so much!

Thanks for taking a look! :rocket:

PR #555, opened 2023-05-15
Fixes an error in bib_key conversion

avonmoll:
- outer iteration is done on `key`
- conversion is done to convert some papis fields to appropriate bibtex equivalents (e.g., `abstractNote` -> `abstract`)
- resulting variable was thereafter `bib_key`
- then, the `document` was being queried by `bib_key` (which, if it had been converted, would not be present in `documet`)
- thus anything that got converted would be skipped
- therefore, `bib_value` should be obtained from `document` using original `key` instead of `bib_key`


alexfikl:
Yeah, you're right! Great catch! (and sorry for the bug, since I touched that stuff last :cry:)

PR #558, opened 2023-05-19
Added bibtex type conversion for bookSection

avonmoll:
- there is no bibtex entry for bookSection
- closest option would be `inbook` (which apparently covers chapters and sections within a book


alexfikl:
Looks good to me! Thanks! :rocket: 

It also comes from Zotero, right?

PR #559, opened 2023-05-21
Add exporter for typst bibliography format

alexfikl:
This adds an exporter for the bibliography format supported by [`typst`](https://github.com/typst/typst) which is described [here](https://github.com/typst/hayagriva/blob/main/docs/file-format.md). It's a reasonably simple YAML format, so we just need to translate `papis` keys.

TODO:
* [x] Add some tests
* [x] Support for the `parent` key for chapters in books and proceedings.

### DIFF
--- a/papis/hayagriva.py
+++ b/papis/hayagriva.py
@@ -1,0 +1,131 @@
+from typing import Any, Dict, List
+
+import papis.document
+import papis.logging
+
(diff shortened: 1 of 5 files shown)


alexfikl:
I'll go ahead and merge this since it gives decent results. It's not fully supported, but we can tweak it as we go for better results :rocket:

PR #561, opened 2023-05-22
A few more conversions were needed

avonmoll:
- according to [bibtex](https://bibtex.com/e/entry-types/) `masterthesis` is actually `mastersthesis`
- added conversions for `monograph`, `presentation`, `blogPost`, and `webpage`

### DIFF
--- a/papis/bibtex.py
+++ b/papis/bibtex.py
@@ -55,7 +55,7 @@ bibtex_types: FrozenSet[str] = frozenset([
     "standard",
     "video",
     # type aliases (Section 2.1.2)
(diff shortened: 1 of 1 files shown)


papis/bibtex.py
```diff
@@ -55,7 +55,7 @@
     "standard",
     "video",
     # type aliases (Section 2.1.2)
-    "conference", "electronic", "masterthesis", "phdthesis", "techreport", "www",
+    "conference", "electronic", "mastersthesis", "phdthesis", "techreport", "www",
```
alexfikl, line 58: Yeah, that was likely a typo. Nice catch!

PR #562, opened 2023-05-22
bibtex: add all Zotero fields to converter

alexfikl:
This adds all the item types from https://github.com/zotero/zotero-schema to the bibtex converter.

@avonmoll Hopefully this is a bit more comprehensive now, at least as far as Zotero is concerned. Do you use other software with non-BibTeX fields that should be added?


alexfikl:
> Looks OK to me. In my case, I do not use anything other than papis. I mainly add docs via DOI and sometimes some non-bibtex types end up getting used. For example, this doi: `10.23943/9781400890088` got imported as `type: monograph`. Granted, that was added back in 2021, so I can't say exactly which version of papis I had installed at the time.

Ah, I see! Well, thanks a lot for taking a look and helping fix these things up!

Issue #417, opened 2022-11-10
[QUESTION]: Custom importer


ghost:
First of all thank you for this amazing tool!

As per title, this is not an issue. 

I find myself adding references from a specifc website ( see [here](https://aclanthology.org/2021.naacl-main.208/) ) , which forces me to use the `fallbck` importer and then I need to manually edit part of the `info.yaml`. 

If I were to write a custom importer for this website, is there a way to plug it, w/o having my own fork of papis?

Thannk you!

jghauser:
Not really an answer to your question as I'm not familiar with that part of the code, but couldn't you just use the `doi` importer? The article you linked to has a `doi` and when I tried it, it seemed to do a reasonably good job at getting the metadata. I would simply download the pdf and then import with: 

`papis add --from doi <DOI> <FILEPATH>`

mar4th3:
Thanks for the suggestion! I haven't thought of that. 

However, this does not completely handle my "ideal" setting. As you mentioned:

1. I need to download the file separately and 
2. I need to specify the `--from doi` option

Being able to get the pdf file directly is a must, as I will for sure forget to add it. Regarding 2, to give a bit more of context, I am using [qutebrowser](https://qutebrowser.org/) and have a key shortcut to call papis. This means I would have to add one just to call `papis add --from doi`, which of course I will forget to do when needed.

alexfikl:
Since that website seems to have a DOI, @jghauser's suggestion is probably the best! 

We should add this to the dev documentation, but it's pretty easy to add a new importer / downloader completely separate from the main `papis` codebase. Basically you write your own package with the downloader and then in `setup.py` (or equivalent) you add it to the `entry_points` like we do here:
https://github.com/papis/papis/blob/f47fbc60c01bd0f709eac19cabf5fa43b93e77ed/setup.py#L190-L211
All of these are simple plugins and could be shipped separately from `papis`, so you should be able to use them as a template. I would suggest something like this:
* Make a basic `Downloader` class like e.g. the one for [Springer](https://github.com/papis/papis/blob/f47fbc60c01bd0f709eac19cabf5fa43b93e77ed/papis/downloaders/springer.py).
* In `get_data`, scrape the website for the DOI.
* Once you have the DOI, call [crossref.get_data](https://github.com/papis/papis/blob/f47fbc60c01bd0f709eac19cabf5fa43b93e77ed/papis/crossref.py#L168) on it to get the info and return that!

[...]

PR #575, opened 2023-06-07
downloaders: add ACL Anthology downloader

mar4th3:
Hi! 

This is related to #417 

I finally managed to find some time to take a crack at writing a custom downloader.

Maybe you'd like to have a look?

The main questions that remain is: how do I test if this is working with `papis add`?

Thanks!


alexfikl:
> The main questions that remain is: how do I test if this is working with `papis add`?

I'm not quite sure how you set up your dev environment, but I usually have a virtual environment set up and then install `papis` in editable mode
```
python -m pip install -e '.[develop]'
```
which also installs the entry points for the plugins (like the downloaders). Then you can just do
```
papis add https://acl.org/some-id
```
and see if it correctly pics up the downloader you added and all the metadata.

HACKING.md
```diff
@@ -102,27 +98,100 @@ see the arXiv importers in `arxiv.py`. To add a new importer
...
-  from which a BibTeX file can be downloaded. This is then parsed and merged
-  automatically.
-- (Optional) Implement the `Downloader.get_document_url` method to return an
-  URL from which a document (e.g. PDF file) can be downloaded.
-
-The downloader can then be added to the `"papis.downloader"` key in `setup.py`,
-similarly to an importer.
+from a remote location. They can be implemented in a similar way.
+
+As a example we implement a custom downloader for the [ACL Anthology](https://aclanthology.org/).
```
alexfikl, line 114: The example you added here is great! However it reads more like a tutorial, so maybe it should be in the docs instead?

The docs for the plugin architecture are pretty nonexistent at the moment
https://github.com/papis/papis/blob/2fe163de082b883f804b19ed76d756b02f05dfdb/doc/source/plugins.rst
but we can start something in there.

setup.py
```diff
@@ -4,11 +4,11 @@
 #   ./venv/bin/pip install --editable .
 #   ./venv/bin/pip install --editable .[dev]  # with dev requirements, too
 
+import glob
 import os
 import sys
-import glob
 
-from setuptools import setup, find_packages
+from setuptools import find_packages, setup
```
alexfikl, line 11: I'm guessing most of these changes are from a formatter (black?). We don't really enforce one at the moment, so can you revert these changes?

papis/downloaders/acl.py
```diff
@@ -0,0 +1,89 @@
...
+    ]
+    for url, y_true in TEST_CASES:
+        d = Downloader(url)
+        d.fetch_data()
+
+        print(f"Test: {url}")
+        print("Fetched:")
+        print(f"\t {d.ctx.data}")
+        assert d.ctx.data["acl_anthology_id"] == y_true["acl_anthology_id"]
+        print("\n\n")
```
alexfikl, line 89: These could go into `tests/downloaders`. There's some existing examples in there, but let me know if you need any help setting them up!

alexfikl (commented):
The downloader itself looks good to me! (Haven't tested it yet though)

I've added some high level comments for the tests and formatting, but those are just moving things around, so hopefully not too hard to clean up.

mar4th3:
Hi!

I finally managed to find some time for this. I have updated the code following your comments. Could you please take a look again?

I have added a `test_acl.py`. The tests are failing because the resource files are not created:

```
E       FileNotFoundError: [Errno 2] No such file or directory: '/home/ele/Projects/papis/tests/downloaders/resources/ACL_2021.naacl-main.208.html'
```

Stemming from `get_remote_resource`. Any idea what I'm doing wrong?

Thanks!

alexfikl:
I'll have to look a bit over the docs, but overall this looks good to me! :heart: 

Do the parsed results in the JSON resource files look good to you?

alexfikl:
> > Yeah, that part is a bit undocumented, so I added the missing resources in a few commits.
> 
> Thanks for the cllarification! What do you think about adding a final section into `plugins.rst` about testing?

That's definitely a good idea, but I think we should put it in a separate PR. I have some ideas for simplifying the workflow there and I'll take a crack at it over the weekend!

> Yeah LGTM!

Awesome! Whenever you're ready to take this out of draft status, we can merge it in!

mar4th3:
> By the way, you might need to merge main (or rebase, if that's your preference), since this branch seems to be quite a bit behind. I imagine Github will complain once this is not a draft anymore 😁

I've changed the PR status and no complaints from Github :) I noticed that there are a couple more changes in the PR on other files (e.g. `tui/widgets/diff.py`) but I guess this  is something you introduced.

alexfikl:
> I've changed the PR status and no complaints from Github :)

Yeah, everything seems ok, so in it goes! Thank you very much for all your work on this! :rocket:

> I noticed that there are a couple more changes in the PR on other files (e.g. `tui/widgets/diff.py`) but I guess this is something you introduced.

That was me! I had some local errors from `flake8`, not sure what that was about, but it looks ok on the CI!

Issue #577, opened 2023-06-22
Wrong string escape in commit message.


Sebastian-Jung:
* papis version (`$ papis --version` or commit number): `0c0beb2415cf6bf686b5d38903d3f7245587cef9`

# how to reproduce the issue
In my case, when adding a paper from `arxiv.org` and having auto-commit set to true (by using the `use-git` config option) I end up commiting the message 

```
git commit -m 'Add document: 'PAPER TITLE''
```

Since the Paper Title is in single quotes `'` and the String used for commiting aswell, the bash command fails.

I have only tested it in my setup and only for arxiv, but I assume this is a general issue.

alexfikl:
Fixed by #578.

AlexBocken:
Is a release planned soon? Currently the latest release does not include this fix which makes the git integration difficult to use (at least the auto-commting/mv etc.)

alexfikl:
> Is a release planned soon? Currently the latest release does not include this fix which makes the git integration difficult to use (at least the auto-commting/mv etc.)

@AlexBocken Yeah, there are quite a few fixes that would be good to get out there. I think we're generally ready to push out a release, but I'm afraid there's no particular timeline for it. @alejandrogallo would just need to find some time to do it.

PR #578, opened 2023-06-22
fix(#577): Add shlex.quote(...) to escape string in commit bash command

Sebastian-Jung:
Fixes Issue #577.

I'm not sure about the position of the `import shlex`, since it's also used in the `_issue_git_command` function, but only imported in this function. Should we move the import inside the `commit` function or should we keep the import outside, and remove the import inside the `_issue_git_command`? What was the idea behind keeping the import inside the function? @alexfikl

### DIFF
--- a/papis/git.py
+++ b/papis/git.py
@@ -46,7 +46,7 @@ def add(path: str, resource: str) -> None:
     :param resource: Commit resource
     """
     logger.info("Adding '%s'.", path)
-    cmd = "git add '{}'".format(shlex.quote(resource))
+    cmd = "git add {}".format(shlex.quote(resource))
     _issue_git_command(path, cmd)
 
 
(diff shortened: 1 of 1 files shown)


alexfikl:
> What was the idea behind keeping the import inside the function?

From a bit of `git blame`-ing, that was moved inside the function in #349 to remove some global imports. Since this is a stdlib module, it likely doesn't matter that much, so if it's used in multiple functions, it should be fine to move it to the top level!

papis/git.py
```diff
@@ -33,9 +35,8 @@ def commit(path: str, message: str) -> None:
     :param path: Folder where a git repo exists.
     :param message: Commit message
     """
-
     logger.info("Committing '%s' with message '%s'.", path, message)
-    cmd = "git commit -m '{}'".format(message)
+    cmd = "git commit -m {}".format(shlex.quote(message))
```
alexfikl, line 39: This looks good to me! 

`add` and `remove` also seem to do some quoting of their own. If you have the time, can you check if those need any special handling too?

Sebastian-Jung: I never had Issues with adding or removing, but since escaping the string shouldn't break it and make it more robust, I added it to `add` and `remove`. I tested it quickly and it works without issues 👍🏽

PR #579, opened 2023-06-22
Feature: Allow .pdf urls to work with arxiv importer.

Sebastian-Jung:
Currently, adding arxiv papers using
```
papis add --from arxiv STRING
```
does not allow adding the `.pdf` url as a string (e.g. `https://arxiv.org/pdf/physics/0504179.pdf`).

This pull request checks if the string retrieved from the regular expression ends with .pdf and if so removes it.
It also adds tests for the new usecases as well as a more describing assertion error.


alexfikl:
Nice! In it goes! :rocket:

PR #580, opened 2023-06-23
Unify command calling

alexfikl:
This unifies all the `subprocess` calls to a `papis.utils.run` that does the proper escaping and anything else that's needed. All commands are ported to use this.

### DIFF
--- a/papis/git.py
+++ b/papis/git.py
@@ -1,31 +1,12 @@
 """This module serves as an lightweight interface for git related functions.
 """
-import os
-from typing import List
-import shlex
+from typing import Sequence
 
+import papis.utils
 import papis.logging
 
 
 logger = papis.logging.get_logger(__name__)
-
-
-def _issue_git_command(path: str, cmd: str) -> None:
-    """Issues a general git command ``cmd`` at ``path``.
-
-    :param path: Folder where a git repo exists.
-    :param message: Git command
-    """
-    assert isinstance(cmd, str)
-    path = os.path.expanduser(path)
-    assert os.path.exists(path)
-
-    split_cmd = shlex.split(cmd)
-    logger.debug("Running command: %s.", split_cmd)
-
-    import subprocess
-    os.chdir(path)
-    subprocess.call(split_cmd)
 
 
 def commit(path: str, message: str) -> None:
@@ -35,8 +16,7 @@ def commit(path: str, message: str) -> None:
     :param message: Commit message
     """
     logger.info("Committing '%s' with message '%s'.", path, message)
-    cmd = "git commit -m {}".format(shlex.quote(message))
-    _issue_git_command(path, cmd)
+    papis.utils.run(["git", "commit", "-m", message], cwd=path)
 
 
 def add(path: str, resource: str) -> None:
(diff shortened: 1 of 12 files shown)


alexfikl:
@Sebastian-Jung This reworks all the `git` commands to use `papis.utils.run` instead of `_issue_git_command` and should hopefully quote everything correctly and uniformly without going through `shlex` so much.

It seems to all still work, but if you get a chance to try this PR out and make sure I didn't ruin anything there, that would be great!

papis/utils.py
```diff
@@ -137,15 +131,17 @@ def general_open(file_name: str,
     import shutil
     if not shutil.which(cmd[0]):
         raise FileNotFoundError(
-            "[Errno 2] No such file or directory: '{}'".format(opener))
+            "[Errno 2] No such file or directory: '{}'".format(cmd[0]))
```
Sebastian-Jung, line 134: Error string should be changed to something like `Command not found : {cmd[0]}`

papis/git.py
```diff
@@ -57,10 +36,8 @@ def remove(path: str, resource: str, recursive: bool = False) -> None:
     :param resource: Commit resource
     """
     logger.info("Removing '%s'.", path)
-    # force removal always
```
Sebastian-Jung, line 60: Was there a reason why `-f` was used initially? Does it break something to do a soft remove?

tests/commands/test_run.py
Sebastian-Jung, line 1: Looks good. In case the error in `papis/utils` run function is changed, the test should should also change the expected exception.

Maybe adding a test to `papis/tests/test_utils.py` for the `papis.utils.run`  could also be considered, although it's basically the same as `papis.commands.run`.

Sebastian-Jung (changes requested):
Besides small comments it looks very good and clean! I gave it a quick test and could not find unexpected behavior 👍🏽 Way better then the shlex hacking.

papis/git.py
alexfikl: I'm not sure, but I added it back in the updated version to not change the behavior.

papis/utils.py
alexfikl: Yeah, good idea! 

I'll leave it as a `FileNotFoundError` though, because that's what `subprocess.call` throws as well if the command is not found, e.g.
```
>>> subprocess.call(["non-existent-command"])
...
FileNotFoundError: [Errno 2] No such file or directory: 'non-existent-command'
```

papis/utils.py
```diff
@@ -100,35 +100,29 @@ def parmap(f: Callable[[A], B],
...
-    :param file_name: a file path to open.
-    :param key: a key in the configuration file to determine the opener used,
-        e.g. :ref:`config-settings-opentool`.
-    :param default_opener: an existing executable that can be used to open the
-        file given by *file_name*. By default, the opener given by
-        *key*, if any, or the default ``papis`` opener are used.
+    This is a simple wrapper around :class:`subprocess.Popen` with custom
+    defaults used to call Papis commands.
+
+    :arg cmd: a sequence of arguments to run, where the first entry is expected
```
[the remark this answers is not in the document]
alexfikl: Good idea! Added.

tests/commands/test_run.py
alexfikl: Yeah, many tests use `papis.utils.run` indirectly, so I'm not sure it's worth adding a specific little tests for it.

alexfikl:
> Besides small comments it looks very good and clean! I gave it a quick test and could not find unexpected behavior 👍🏽 Way better then the shlex hacking.

Thanks a lot for the thorough review! I think I got everything, so in it goes :rocket:

PR #594, opened 2023-07-08
Better command overwriting handling

alexfikl:
This adds some clear messages about how commands are loaded. Namely:
* If two external commands have the same name, i.e. `papis-something`, but are located in different paths, a message is printed and the last one in `$PATH` is preferred. Before no message was printed.
* If an entrypoint commands, declared in `setup.py`, and an external command have the same name, the entry point is always preferred. Before the external command was preferred.
 
The messages are only printed with `PAPIS_DEBUG=ON`, but we can change that if we want it to be a more obvious warning (?).

### DIFF
--- a/doc/source/developer_reference.rst
+++ b/doc/source/developer_reference.rst
@@ -25,6 +25,12 @@ Developer API reference
 .. automodule:: papis.cli
     :members:
 
+``papis.commands``
+^^^^^^^^^^^^^^^^^^
+
+.. automodule:: papis.commands
+    :members:
+
 ``papis.config``
 ^^^^^^^^^^^^^^^^
 
--- a/papis/commands/__init__.py
+++ b/papis/commands/__init__.py
@@ -1,8 +1,6 @@
 import os
 import re
-import glob
-from typing import Dict, Optional, Union, NamedTuple
-import difflib
+from typing import Dict, Optional, NamedTuple
 
 import click.core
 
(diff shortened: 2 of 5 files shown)


alexfikl:
@supersambo I noticed this with `papis-tui` which was installing a console script and an entry point, but the console script was overwriting the entrypoint in the commands. Unfortunately, papis looks for a `papis-short-help` at the top of the script and it wasn't finding it to display the help.

Do you have any preference about how those should be loaded?

supersambo:
I'm afraid I do not fully understand the issue :(. I remember tinkering around with this trying to get `papis --help` to display the short help. Currently (on my system) this works properly without setting putting `papis-short-help` comment anywhere. I assumed this comes from the click.command decorator?

`papis-tui` registers the standalone console_script `papis-tui` and the papis.command `tui`. It didn't occur to me that the first one could be overwriting the second one in any way because it has a different syntax (papis**-**tui vs papis tui).

I guess the `papis-tui` version is not even necessary and I could even get rid of it.  Not sure if this provides any sort of meaningful input for your question? :)

alexfikl:
@supersambo Basically papis can add commands (that show in `papis --help`) in two ways
* Using the python entrypoint mechanism (through `stevedore`); you need to add these in `setup.py`.
* As external scripts anywhere in the PATH; basically any script called e.g. `/usr/bin/papis-<something>` will get picked up and a command called `<something>` will be added.

For `papis-tui`, it had both of these and the external one from `/usr/bin/papis-tui` was overriding the entrypoint because they're loaded in that order here
https://github.com/papis/papis/blob/3c8114ae6b46daf6a0bcf5b3ff76b35b989b2ea4/papis/commands/default.py#L53-L54

This PR changes the order so that the entrypoint registered commands take priority. I was just wondering if you had any preference here, because I don't feel very strongly either way :grin:

alexfikl:
Also, for external commands, the help that gets displayed is picked up from a `papis-short-help` blurb at the top of the script, like in this example https://github.com/papis/papis/blob/main/examples/scripts/papis-mail.

supersambo:
Ok, got it now :). For the very specific case of `papis-tui` this would indeed help and I remember being somewhat confused this didn't work as expected. However, as said before, the `papis-tui` console script is actually completely redundant since it does not provide any additional functionality. 

FYI: the reason this works on my machine is again related to pipx. Given that I injected `papis-tui` into `papis` pipx venv, there wasn't any standalone console script available

alexfikl:
> Ok, got it now :). For the very specific case of `papis-tui` this would indeed help and I remember being somewhat confused this didn't work as expected. However, as said before, the `papis-tui` console script is actually completely redundant since it does not provide any additional functionality.
> 
> FYI: the reason this works on my machine is again related to pipx. Given that I injected `papis-tui` into `papis` pipx venv, there wasn't any standalone console script available

Thanks a lot for taking a look and the discussion! This also adds some debug messages saying what overwrites what, so hopefully it will be less confusing going forward.

PR #596, opened 2023-07-11
Avoid formatting exception as string that leaks into papis library

f0rki:
So my usecase is the following: I have following option in my config:
```
add-folder-name = {doc[year]}/{doc[author_list][0][family]} - {doc[year]} - {doc[title]}
```
This tries to index into the `author_list`. However, if a given document does not have an `author_list`, it will throw an exception. In this case, papis will log the exception in debug mode and return a string based on the exception. Now in my case this lead to having a bunch of `index-error-out-of-bounds` directories in my papis library, which is quite annoying.

I don't think this is a good idea to handle it this way. Imho displaying an exception log to the user is preferable to this behavior, because then at least you have a chance at figuring out what went wrong. I changed the `papis.format.format` function to not catch exception and instead propagate them to the caller and handle them e.g., by falling back to sane defaults (as I did for the add command).


alexfikl:
Definitely agree that returning the exception in the format is quite unhelpful and just adds nonsense to the database, so we should definitely improve on it! Bubbling up the exception seems like it's going to cause a lot of pain though: a lot of places call `papis.format.format` and would need to be wrapped in a `try..except`.

What do you think about just adding an optional default value instead? For the use case in `add` that you modified, it could be
```
papis.format.format(file_name_opt,
                    papis.document.from_data(data),
                    default=os.path.basename(original_filepath))
```
? That should be less noisy for callers and still allow setting a reasonable default if the format fails.

f0rki:
Yeah makes sense too. I'll update the PR.

f0rki:
@alexfikl I updated the PR. In case there is no reasonable default, the function will still throw an exception.

alexfikl:
> @alexfikl I updated the PR. In case there is no reasonable default, the function will still throw an exception.

That sounds perfect! Thanks for working on this and updating everything!

papis/format.py
```diff
@@ -153,11 +147,20 @@ def get_formater(name: Optional[str] = None) -> Formater:
...
-    return formater.format(fmt, doc, doc_key=doc_key, additional=additional)
+    try:
+        return formater.format(fmt, doc, doc_key=doc_key, additional=additional)
+    except Exception as e:
+        if default is not None:
+            logger.debug("Failed to format according to '%s' - %s", fmt, e)
+            return default
+        else:
+            logger.warning("Failed to format according to '%s' - %s", fmt, e)
+            raise
```
alexfikl, line 166:
```suggestion
    except Exception as exc:     
        if default is not None:
        	logger.warning("Could not format string '%s'.", fmt, exc_info=exc)   
            return default
        else:
            raise FormatFailedError(fmt) from exc
```
What do you think about something like this? i.e. adding a `FormatFailedError` exception that would catch all the errors thrown here, so that callers would only have to deal with that.

papis/format.py
```diff
@@ -66,13 +66,7 @@ def format(self,
...
 
         doc_name = doc_key or self.default_doc_name
-        try:
-            return fmt.format(**{doc_name: doc}, **additional)
-        except Exception as exc:
-            logger.debug("Could not format string '%s' for document '%s'",
-                         fmt, papis.document.describe(doc), exc_info=exc)
-
-            return "{}: {}".format(type(exc).__name__, exc)
+        return fmt.format(**{doc_name: doc}, **additional)
```
alexfikl, line 69: This should also be changed in the jinja formatter.

papis/document.py
```diff
@@ -421,7 +421,7 @@ def describe(document: Union[Document, Dict[str, Any]]) -> str:
     """
     return papis.format.format(
         papis.config.getstring("document-description-format"),
-        document)
+        document, default="")
```
alexfikl, line 424:
```suggestion
        document, default=doc.get("papis_id", ""))
```
? The `papis_id` key should at least be in all the documents saved to disk..

alexfikl (changes requested):
Left a few nitpicks and a suggestion to add a special exception for format errors. Should be good to go once those are addressed! :tada:

papis/format.py
f0rki: implemented.

f0rki:
@alexfikl all good now?

alexfikl:
> @alexfikl all good now?

Updated the docs slightly, but otherwise this is good to go! Thanks for all the work here! It's gonna make this a lot less nasty to work with.

Will merge it when the tests pass! :rocket:

alexfikl (approved):

PR #597, opened 2023-07-11
Updated Dockerfile to correctly execute all testcases

f0rki:
The `test_edit.py` testcase executes itself on the command line and failed because it would not find the `tests` module. installing with `-e` seems to fix the issue. Also python 3.8 seems to be required now.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,4 @@
+Dockerfile
 papis/deps/*
 .vim-run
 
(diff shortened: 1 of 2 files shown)


alexfikl:
This looks good to me! 

That didn't occur to me, but it's fine if it requires an editable install (with `pip install -e`) to run the tests, right?

f0rki:
I don't think it makes a big difference, but with `-e` then you could mount your local changes over `/papis` and likely it would still work without rebuilding the container.

PR #599, opened 2023-07-12
Another Dockerfile update

f0rki:
* Updated `Dockerfile` to run the ci tests by default instead of pytest only.
* Added a `release.Dockerfile` that could be used to publish to ghcr.io or docker hub. Might be useful if someone wants to run a `papis serve` on a home server or similar or on container focused distributions. Publishing to ghcr.io is pretty easy with github workflows.


Dockerfile
```diff
@@ -11,4 +14,5 @@ COPY . /papis
 
 RUN pip install -e .[optional,develop]
 
-CMD ["pytest", "tests", "papis", "--cov", "papis"]
+# CMD ["pytest", "tests", "papis", "--cov", "papis"]
```
alexfikl, line 17:
```suggestion
```
? Doesn't seem needed anymore.

HACKING.md
```diff
@@ -36,6 +36,17 @@ The docs can be generated with
 make doc
 ```
 
+To quickly get things up and running, you can also use docker/podman:
+
+```
+# build the image
+docker build -t papis .
+# to run the CI tests
+podman run -v $(pwd):/papis --rm -it papis
```
alexfikl, line 45: I don't use docker or podman, so not sure what a good workflow would look like here. Why not just suggest one or the other, e.g.
```
docker run -it --rm papis 
```
?

alexfikl (commented):
Left a comment (more of a question for my curiosity) but otherwise this looks good to me!

EDIT: I ran the commands below and the tests failed. It seemed like it didn't find the `bibtex` command, which might mean `papis` wasn't installed correctly. Have you tried this? Might just be something off with my setup..
```bash
docker build -t papis .
docker run -it --rm papis
```

HACKING.md
f0rki: oops. I use podman on some machines and docker on others, but they are pretty much interchangable, so that is a copy-paste mistake!

f0rki:
@alexfikl you are right, it seems I only tested with `podman run -v $(pwd):/papis --rm -it papisdev` and not with `podman run --rm -it papisdev`.

alexfikl:
Everything works now! Thanks again for working on this! :rocket:

PR #600, opened 2023-07-14
Added `flake.nix` to enable usage with the nix package manager.

f0rki:
Two of my usecases are covered with this `flake.nix` (based on the code from nixpkgs itself):

1. `nix develop` - quickly set up a development environment using the `nix` package manager (actually also sets up a venv to get some of the typing packages).
2. "Install from main" with `nix profile install '.#papis'` if you have cloned the repo, or also `nix profile install 'github.com:papis/papis#papis'` should work.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,145 @@
+{
+  description = "Papis - Powerful command-line document and bibliography manager";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = {
+    self,
+    nixpkgs,
+    flake-utils,
+  }:
+    flake-utils.lib.eachDefaultSystem (system: let
+      pkgs = nixpkgs.legacyPackages.${system};
+      python = pkgs.python310;
+      pypkgs = pkgs.python310Packages;
+      lib = pkgs.lib;
+
+      runtime_py_deps = with pypkgs; [
+        pyyaml
+        arxiv2bib
+        beautifulsoup4
+        bibtexparser
+        chardet
+        click
+        colorama
+        dominate
+        filetype
+        habanero
+        isbnlib
+        lxml
+        prompt_toolkit
+        pygments
+        pyparsing
+        python-doi
+        python-slugify
+        requests
+        stevedore
+        tqdm
+        whoosh
+      ];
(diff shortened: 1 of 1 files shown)


alexfikl:
Do these files usually go with the package? It seems highly specific to one package manager. I'm mostly familiar with `conda` and there the [feedstocks](https://github.com/conda-forge) are kept separate.

I have no experience with nix, so can't really say if the build file is reasonable otherwise :(

alexfikl:
@alejandrogallo Any thoughts on including this? There's some mention of NixOS in the docs that was added a while ago, so we could make it easier for people.

jghauser:
I'm actually on NixOS these days, and this would be very useful! Though I can understand the misgivings about having files specific to given package managers clutter the repo...

alexfikl:
> I'm actually on NixOS these days, and this would be very useful! Though I can understand the misgivings about having files specific to given package managers clutter the repo...

Fair enough, I'm all for it if it's useful for development! If it ever becomes unmaintained and a burden, we can just remove it..

f0rki:
> Do these files usually go with the package? It seems highly specific to one package manager. 

Yes, in the case of a `flake.nix`, it must be tracked along with the code. You can actually think about a `flake.nix` file similar to a `Dockerfile`, but additionally you can commit the `flake.lock` file to lock the versions of all packages (similar to a `Cargo.lock`, `Pipfile.lock`, etc.)

f0rki:
>  If it ever becomes unmaintained and a burden

Shall I check whether there is a useful CI job to monitor that?

alexfikl:
> Shall I check whether there is a useful CI job to monitor that?

The `Dockerfile` is in a similar predicament and it isn't checked either, so no biggie. It's always welcome though! :grin:

alejandrogallo:
I support this, I think we could even write a CI workflow that checks for the building of these kind of packages (in the future or if someone feels like it please go ahead).
I am a NixOS user for some years so I might be biased.
Maybe we could save these files in a `packaging` directory or something like this.
But for now, as you all said, since we have rather few of these files we can just leave it like this.

alexfikl:
@f0rki Do you want to work on adding some CI for this? No hurry or pressure! Otherwise we can just merge it, since it seems it will be very appreciated :grin:

f0rki:
@alexfikl you can merge this PR. don't know when I come around to add CI to this. I'll do a separate PR then.

PR #602, opened 2023-07-17
add a more complex query syntax introducing literal and, or, not and parenthesis

supersambo:
This adds a more complex syntax to papis docmatcher introducing an explicit **and**, **or**, **not** and the use of **parenthesis**.

*This is my first pull request ever! Please let me know if I did something wrong or completely missed any step in the desired procedure. Given the importance of the docmatcher in papis, please consider this as a starting point for a discussion on how to best implement this.*

The way this works is relatively simple: 
1. `papis_query` matches each query instance and the available operators and parenthesis
2. before returning the parsed query missing `and` operators get added via the `proofread` function. This is supposed to make sure the original syntax still works
3. `return_if_match` then passes each query to the matcher and puts everything together again. e.g.:
    - `author: einstein or title: photons` -> `True or False`
    - `einstein and photons not albert` -> `True and True and not True`
4. This resulting boolean operation is then evaluated using vanilla python (e.g. `True or False` = `True`)

Here are some advantages and disadvantages to consider:

**Pros:**
- In theory this introduces the possibility to write very complex queries. 

**Cons:**
- It is slower than the original implementation since it cannot stop once any expression evaluates to `False`. Instead, it always evaluates each single query. I guess this is not an issue for normal usage with a decent amount of documents and a similarly decent level of query complexity. However, when using this for scripting with large numbers of concatenated `or` operators, this might become an issue.
- Of course this is more error prone than the original implementation. At this point the `proofread` function only catches missing `and` operators using a very simple logic, but it does not fix other syntax errors. e.g. "einstein or and albert" will just fail to evaluate. I guess one could come up with some rules for these cases as well but probably not for everything. Also, this may confuse users even more. Maybe it would be necessary to be more verbose here: 
  - Report exactly where `and` operators where inserted
  - When evaluation fails, catch `SyntaxError` message and use it to hint to where the error is

**Some more thoughts to consider:**

As expected, this doesn't pass the `test_parse_query` test, since I changed the logic of the output format. 

This uses literal `and` `or` etc. I was not sure wether Alejandro was suggesting some sort of "escaped" operators such as `.or.` `.and.` and so forth.

This introduces some minor changes in how queries are evaluated. For instance "einstein albert" was translated to a single regex pattern but is parsed as "einstein and albert" now. This means that the order of words does not matter any more. To me personally, this makes more sense but of course this is a matter of taste and you might not want to change how things work for the users. 

I selfishely added `alphas8bit` to the `papis_value_word` matcher because I realized that I was'nt able to search for my own surname :). `papis_value_word` also allows for parnthesis which of course is not the case anymore in this pull request. 

*edit note: fix typos, add ideas to make this more verbose*

### DIFF
--- a/papis/docmatcher.py
+++ b/papis/docmatcher.py
@@ -37,12 +37,13 @@ class MatcherCallable(Protocol):
     .. automethod:: __call__
     """
 
-    def __call__(self,
-                 document: papis.document.Document,
-                 search: Pattern[str],
-                 match_format: Optional[str] = None,
-                 doc_key: Optional[str] = None,
-                 ) -> Any:
+    def __call__(
+        self,
+        document: papis.document.Document,
+        search: Pattern[str],
+        match_format: Optional[str] = None,
+        doc_key: Optional[str] = None,
+    ) -> Any:
         """Match a document's keys to a given search pattern.
 
         The matcher can decide whether the *match_format* or the *doc_key* take
@@ -92,8 +93,8 @@ class DocMatcher:
 
     @classmethod
     def return_if_match(
-            cls,
-            doc: papis.document.Document) -> Optional[papis.document.Document]:
+        cls, doc: papis.document.Document
+    ) -> Optional[papis.document.Document]:
         """Use :attr:`DocMatcher.parsed_search` to match the *doc* against the query.
 
             >>> import papis.document
@@ -117,13 +118,19 @@ class DocMatcher:
         if cls.parsed_search is None or cls.matcher is None:
             return match
 
+        tokens = []
         for p in cls.parsed_search:
-            match = (
-                doc if cls.matcher(doc, p.pattern, cls.match_format, p.doc_key)
-                else None)
+            if type(p) == ParseResult:
+                tokens.append(
+                    "True"
+                    if cls.matcher(doc, p.pattern, cls.match_format, p.doc_key)
+                    else "False"
+                )
+            else:
+                tokens.append(p)
 
-            if not match:
-                break
+        result = eval(" ".join(tokens))
+        match = doc if result else None
 
         return match
 
@@ -131,9 +138,9 @@ class DocMatcher:
     def set_search(cls, search: str) -> None:
         """Set the search for this instance of the matcher.
 
-            >>> DocMatcher.set_search('author:Hummel')
-            >>> DocMatcher.search
-            'author:Hummel'
+        >>> DocMatcher.set_search('author:Hummel')
+        >>> DocMatcher.search
+        'author:Hummel'
         """
         cls.search = search
 
@@ -141,17 +148,18 @@ class DocMatcher:
     def set_matcher(cls, matcher: MatcherCallable) -> None:
         """Set the matcher callable for the search.
 
-            >>> from papis.database.cache import match_document
-            >>> DocMatcher.set_matcher(match_document)
+        >>> from papis.database.cache import match_document
+        >>> DocMatcher.set_matcher(match_document)
         """
         cls.matcher = matcher
 
     @classmethod
-    def parse(cls, search: Optional[str] = None) -> List[ParseResult]:
-        """Parse the main query text.
+    def parse(cls, search: Optional[str] = None) -> List[ParseResult | str]:
+        """Parse the main query text and check its syntax.
 
         This method will also set :attr:`DocMatcher.parsed_search` to the
-        resulting parsed query and it will return it too.
+        resulting parsed query and it will return it too. If there is a syntax
+        error it will log an error message and return None.
 
             >>> print(DocMatcher.parse('hello author : einstein'))
             [['hello'], ['author', 'einstein']]
@@ -169,7 +177,13 @@ class DocMatcher:
         """
         if search is None:
             search = cls.search
-        cls.parsed_search = parse_query(search)
+        parsed_search = parse_query(search)
+
+        if test_syntax(parsed_search):
+            cls.parsed_search = parsed_search
+        else:
+            cls.parsed_search = None
+
         return cls.parsed_search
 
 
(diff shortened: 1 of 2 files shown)


alejandrogallo:
This is a very nice idea to do this now, I'll be taking a look at it soon also!

papis/docmatcher.py
```diff
@@ -169,7 +191,13 @@ def parse(cls, search: Optional[str] = None) -> List[ParseResult]:
         """
         if search is None:
             search = cls.search
-        cls.parsed_search = parse_query(search)
+        parsed_search = parse_query(search)
+
+        if check_syntax(parsed_search):
+            cls.parsed_search = parsed_search
+        else:
+            cls.parsed_search = []
```
alexfikl, line 199: I think this sort of check should be just part of the tests. Normally `pyparsing` would throw an error if it does not manage to parse something and our own post-processing should be relatively simple.

papis/docmatcher.py
```diff
@@ -37,12 +51,13 @@ class MatcherCallable(Protocol):
...
-                 match_format: Optional[str] = None,
-                 doc_key: Optional[str] = None,
-                 ) -> Any:
+    def __call__(
+        self,
+        document: papis.document.Document,
+        search: Pattern[str],
+        match_format: Optional[str] = None,
+        doc_key: Optional[str] = None,
+    ) -> Any:
```
alexfikl, line 60: Using black? There are a few unrelated changes like this and it would be helpful to revert them.

papis/docmatcher.py
```diff
@@ -117,50 +132,57 @@ def return_if_match(
...
     @classmethod
     def set_search(cls, search: str) -> None:
         """Set the search for this instance of the matcher.
 
-            >>> DocMatcher.set_search('author:Hummel')
-            >>> DocMatcher.search
-            'author:Hummel'
+        >>> DocMatcher.set_search('author:Hummel')
+        >>> DocMatcher.search
+        'author:Hummel'
```
alexfikl, line 157: Is this a formatting change or do the doctests not run if they're indented? I should really check that :worried:

papis/docmatcher.py
```diff
@@ -186,8 +214,57 @@ def get_regex_from_search(search: str) -> Pattern[str]:
...
+
+    try:
+        eval(" ".join(test))
+        return True
+    except SyntaxError:
+        logger.error("Malformed query.")
+        return False
+
+
+def proofread(parsed: List[ParseResult]) -> List[ParseResult]:
```
alexfikl, line 241: This could benefit from a more obvious name. Perhaps `add_implicit_query_operators` or something? I'm not good at naming :(

papis/docmatcher.py
```diff
@@ -186,8 +214,57 @@ def get_regex_from_search(search: str) -> Pattern[str]:
...
+            # add 'and' when queries or syntax have have no operator in between
+            result.append(
+                ParseResult(
+                    syntax=True, string="and", pattern=get_regex_from_search("")
+                )
+            )
+            fixes += 1
+
+    if fixes > 0:
+        logger.debug("Fixed query by adding %s missing 'and' operator(s).", fixes)
```
alexfikl, line 265:
```suggestion
    logger.debug("Fixed query by adding %s missing 'and' operator(s).", fixes)
```
? Since it's a debug message, might as well always print it.

alexfikl (commented):
This is really awesome! Thanks a lot for working on it :heart:
I left a couple of nitpicks while reading through it. 

I do have one bigger question though. Currently the parsing just uses a `list` of results because it's always just a `A and B and C and D` query. However, with the operators you added, the parsed query becomes more of a tree. Do you think it's worth changing the data structure for that (to a nested `dict`)?

@alejandrogallo probably has a better idea about how that should look like, so we should wait until he has a moment to give it a look too!

papis/docmatcher.py
supersambo: Yes, I ran black on this. Will revert.

papis/docmatcher.py
supersambo: Not sure how this happened. Will revert.

papis/docmatcher.py
supersambo: You're right proofread is rather abstruse. I will use your suggestion or something similar!

papis/docmatcher.py
supersambo: I guess this is a somewhat misleading function name. `check_syntax` actually checks for logical consistency of a query that is not controlled by `pyparsing`. So it catches everything that cannot be evaluated as an expression of boolean values combined with operators such as `True or and False` or `True or or False` or `True and False or` etc. Btw. I put this in `parse` because otherwise `return_if_match` throws an error for each document in the library.

papis/docmatcher.py
supersambo: Agree!

supersambo:
Thanks for your comments and suggestions! I will add them soon!

I'm not entirely sure about the tree structure. I understand that this actually becomes a tree especially when using parenthesis. However, at this point it does not yet care about the tree structure, neither does it get parsed as such because figuring that out is left to the pythons `eval` function.

I do like the simplicity of this but (as mentioned above) this might not be the most efficient way to do this. For instance:
```
einstein and (albert or mileva or elsa or hans or eduard or lieserl or ...)
```
If we would know about the tree structure and the first expression `einstein` evaluates to `False`, then we could spare ourselves the pain of evaluating all the expressions in the parenthesis (similar to what papis does now). However, I believe implementing this with more operators would add quite a lot of complexity.

I hope this relates to your question? :)

supersambo:
Hey there, I was wondering if there is still interest in implementing this. If so, I could give it another try going in the direction @alexfikl suggested before.

alexfikl:
> Hey there, I was wondering if there is still interest in implementing this. If so, I could give it another try going in the direction @alexfikl suggested before.

Hey! Personally I'm still not sold on the approach, but the idea is definitely still of interest (see the recent #1128). Some things to consider:
1. #1007 should hopefully go in sooner rather than later. That also has a pretty usable search syntax (with built in AND/OR and some wildcards).
2. The current code using `pyparsing` is pretty horrible, e.g. it doesn't support unicode #1004. I started playing around with the [lark](https://github.com/lark-parser/lark) parser, which should make it a bit easier to add both unicode and this. What do you think?

I started experimenting with `lark` in here https://github.com/alexfikl/papis/blob/docmatcher-lark/papis/docmatcher.py, but didn't get very far.

supersambo:
Ok, great. I will look into this and see if I can contribute something usefull

alexfikl:
Superseded by #1165.

PR #603, opened 2023-07-18
Papis cache command

alejandrogallo:
This pr introduces `papis cache` command.

The command `update-older` could be interesting for updating the cache automatically
after a `git pull`, since it looks around the documents and checks the modification time
against the cache file and if it is newer it updates only that document in the cache.

There is still some documentation to be written and maybe some tests, the commands are

```
 papis cache update
 papis cache clear
 papis cache reset
 papis cache add
 papis cache rm
 papis cache pwd
 papis cache update-older
```


CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+VERSION v0.14
+=============
+
+## Add `cache` command (remove of `papis --cc/--clear-cache`) ([#603](https://github.com/papis/papis/pull/603))
```
alexfikl, line 4:
```suggestion
## Add `cache` command (remove `papis --cc/--clear-cache`) ([#603](https://github.com/papis/papis/pull/603))
```

papis/commands/cache.py
```diff
@@ -0,0 +1,213 @@
+"""
+This command allows the user to interact to the papis cache or papis database.
```
alexfikl, line 2:
```suggestion
This command allows the user to interact with the papis cache or papis database.
```

papis/commands/cache.py
```diff
@@ -0,0 +1,213 @@
...
+
+Before ``v0.14``, you could only clear the cache by doing
+
+::
+
+    papis --clear-cache
+
+Now you can do the same by typing the following command
```
alexfikl, line 10:
```suggestion
To clear the cache (remove it from the filesystem), you can run the following
command
```
? It's probably just confusing to mention the old way of doing it (that gets removed anyway)?

papis/commands/cache.py
```diff
@@ -0,0 +1,213 @@
...
+
+::
+
+    papis cache clear
+
+This command is also useful for plugin developers.
+Let us suppose that you are editing the yaml file of a document at path
```
alexfikl, line 17:
```suggestion
Let us suppose that you are editing the YAML file of a document at path
```

papis/commands/cache.py
```diff
@@ -0,0 +1,213 @@
...
+
+or maybe by query
+
+::
+
+    papis cache update query-matching-document
+
+Furthermore, a noteworthy subcommand is ``update-newer``, which
+updates the cache for those documents whose info file is newer than
+the cache itself.  This subcommand has the same interface than most
```
alexfikl, line 35:
```suggestion
the cache itself.  This subcommand has the same interface as most
```

papis/commands/cache.py
```diff
@@ -0,0 +1,213 @@
...
+papis commands, so that if you want to check all documents you have to
+input
+
+::
+
+        papis cache update-newer --all
+
+This command is much faster than rebuilding the cache in general.
```
alexfikl, line 43:
```suggestion
This command is much faster than rebuilding the cache from scratch.
```

papis/commands/cache.py
```diff
@@ -0,0 +1,213 @@
...
+You can therefore run this command once in a while in order to update
+the cache for those documents that have been synchronised by the means
+of synchronization that you are using, for instance using git, syncthing,
+dropbox, etc.
```
alexfikl, line 47:
```suggestion
of synchronization that you are using, for instance using git, Syncthing,
Dropbox, etc.
```

papis/database/base.py
```diff
@@ -38,6 +38,12 @@ def get_dirs(self) -> List[str]:
         """
         return self.lib.paths
 
+    def get_cache_path(self) -> str:
+        """Get the path to the actual cache file or directory.
+        """
+        raise NotImplementedError("Match not defined for backend '{}'"
```
alexfikl, line 44:
```suggestion
        raise NotImplementedError("Cache path not defined for backend '{}'"
```
?

alexfikl (approved):
I was curious how this was implemented and took a quick look. It's very nice and simple! :heart:

Left a few comments for typos in the docs, but otherwise this is great!

papis/commands/cache.py
alejandrogallo: agreed!

alejandrogallo:
Thank you for your comments! I think this will be quite useful as a starting point for future functionality. Maybe now it is also the time to make papis jump into the help whenever we simply type `papis`, don't you think? 
Before the only reason not to do this was that the command `--cc` had to be issued somehow, I think now just typing `papis` should give the list of commands available and the flags for papis. What do you think ?

alexfikl:
> Thank you for your comments! I think this will be quite useful as a starting point for future functionality. 

Definitely! It looks very helpful already.

> Maybe now it is also the time to make papis jump into the help whenever we simply type `papis`, don't you think? Before the only reason not to do this was that the command `--cc` had to be issued somehow, I think now just typing `papis` should give the list of commands available and the flags for papis. What do you think ?

Yeah, that sounds like a good idea! Now it seems like it does nothing when you call just `papis`, which might be a bit confusing.

PR #605, opened 2023-07-19
document: fix format default in describe

alexfikl:
@alejandrogallo Hopefully this fixes your issue! There's probably more lurking around though :worried:


alejandrogallo:
awesome thanks!

PR #607, opened 2023-07-21
Switch to pyproject.toml

alexfikl:
This PR moves all the configuration to a `pyproject.toml` file from `setup.cfg` and `setup.py`. It uses [hatchling](https://hatch.pypa.io/latest/) as a build backend instead of setuptools because it seems like it has good support for editable installs and other build-related things.

Unfortunately, we still need:
* ~`setup.cfg` for configuring `flake8`.~
	- ~This could be switched to something like `ruff` in the future?~
	- ~Using `flake8-pyproject` to get around flake8 not having direct support for `pyproject.toml`~
* `setup.py` for configuring and installing the bash/fish/zsh completion.
	- ~Not quite sure what the situation is with this; will need to hunt down some projects installing completions and manpages.~
	- ~This seems to be possible by adding them in a `[tool.setuptools.data-files]` section, but it's not platform independent.~


alejandrogallo:
I'm for this, I think it makes everything cleaner, that we need the cfg and setup.py files is not a biggie in my opinion.

alexfikl:
> I'm for this, I think it makes everything cleaner, that we need the cfg and setup.py files is not a biggie in my opinion.

@alejandrogallo Awesome! Can you check that the packaging on your end also works?

I just ran `tools/upload-pypi.sh` (without the upload) on both main and this branch and it seems to be generating the same sdist. I also modified a Arch Linux PKGBUILD to use this branch and it seems to install everything correctly as well!
https://gitlab.com/alexfikl/pkgbuild/-/blob/f820c40a0f0fffaec3d3949ffa8cf9179269d880/papis-git/PKGBUILD

The remaining issues are:
* `setup.cfg` needs to stay around for `flake8` (which does not seem to want to add support any time soon).
* the data files (bash completion, manpages) are installed on all platforms, which may not be desired.

f0rki:
@alexfikl FYI: I have tested ruff in the docker branch and also pushed some fixes (missing noqa for ruff and removal of unnecessary noqa).

alexfikl:
> @alexfikl FYI: I have tested ruff in the docker branch and also pushed some fixes (missing noqa for ruff and removal of unnecessary noqa).

I mostly switched to ruff in my own projects too and it's pretty nifty. However, for papis we were supporting python 3.6 just a couple of months ago, so I'm weary of moving to the latest tools too fast..

f0rki:
btw. after (or along with) the switch to poetry, we should switch to [poetry2nix](https://github.com/nix-community/poetry2nix) in the `flake.nix`.

kiike:
I have been trying to achieve conditional inclusion of files depending on the platform and I think I managed to do that.

For this purpose, I had to change the build backend from setuptools to hatchling, which is the one used by airflow, for instance. It's the backend with the highest bus factor but it looks like it is being adopted progressively. I chose this backend because it let me create a hook to include files conditionally in a very easy way.

It's a little bit late so I just had time to test on Linux, but the wheel is more or less built the same, with the data files. They are in a different root dir, named "data" instead of "papis-0.13.data". I can change that easily if needed, but I'd like to understand if that is a problem first.

[My commit and the CI results](https://github.com/kiike/papis/pull/3/commits/debe98e27749e3fb4f8878f9ae5798df85c6f8ba).

alexfikl:
> I have been trying to achieve conditional inclusion of files depending on the platform and I think I managed to do that.
> 
> For this purpose, I had to change the build backend from setuptools to hatchling, which is the one used by airflow, for instance. It's the backend with the highest bus factor but it looks like it is being adopted progressively. I chose this backend because it let me create a hook to include files conditionally in a very easy way.

Hm, from looking at our data-files, we try and package a desktop file, some man files, some completions and the README + LICENSE. As far as I can tell:
* on Linux, these would all be handled by the distro package in a better way, since they know where all those files should go (e.g. on Arch, the license goes in `/usr/share/licenses`, not in `/usr/share/doc`). 
* Not sure how this works on macOS, but I imagine our paths are generally incorrect there too and Homebrew / MacPorts does that better.
* On Windows it's probably a whole different format for all of those.

So I would be more inclined to remove those things all together and let more OS-specific packaging handle them. They're all easy to generate, e.g. completions just need a call to
```sh
_PAPIS_COMPLETE=zsh_source papis
```

kiike:
Thank you for taking a look and considering the changes, @alexfikl :heart:!

> So I would be more inclined to remove those things all together and let more OS-specific packaging handle them. They're all easy to copy over or generate, e.g. completions just need a call to

Here is a list of files I think are ok to **remove** from `shared-data`:

- README: no need to build it inside the wheel, as it is read by the build backend and embedded into `Description-Content-Type` inside the `METADATA` in `papis-0.13.dist-info`
- LICENSE: implicit text in the license identifier in `License` in the previously-mentioned file. Explicit contents inside papis-0.13.dist-info/licenses/LICENSE.
- AUTHORS: embedded into `METADATA` too (for some weird reason inside `dist-info/licenses`, but setuptools did that too).

As for the rest of the files

- manpages: I want to help this project as long as I use it and right now I know I am outsider, but my opinion is that the manpages, along with the desktop file, are **worth keeping** inside the wheel. Although it bloats the installation by almost 200K :wink:, if a user installs papis from the wheel into a global namespace, they'll get the manpages inside a standard path. And if they install with pipx, they will get discovered and installed along with other pipx-managed manpages. And, for the adventorous who installs with pip, they will be indeed inside site-packages :disappointed: , but it's just a matter of appending to `MANPATH` if they're interested in those manpages at all. Nothing to do if you don't want manpages.

- desktop file: it works fine if installing to system or `pip` with or without `--user`; they'll be in an expected location. For packagers, they can take the default location for the `installer` module or move the file to a more suitable path. If installing into a venv, it's the same story: if the user wants to use that desktop, they'll have to copy it or symlink to `~/.local/share/applications`. I do it for `yt-dlp`, for instance.

- completions: I think it's OK to keep them, as it's the same procedure as desktop files. You can link them to your local shell completion dir if you wish or importing with the command you wrote. [yt-dlp](https://github.com/yt-dlp/yt-dlp/blob/8463fb510a58050ec118b3ae17bf00d08ea7b881/pyproject.toml#L111) or [jupyter](https://github.com/jupyter/notebook/blob/d45362b5492cf4f72f4aa696c7190083a6509a62/pyproject.toml#L81) ship them in the wheel, just like papis is doing now. 

I agree that package managers will make a better job to install these files at whichever locations are best for the system, so they can [take the files as is (example from jupyter in Arch)](https://gitlab.archlinux.org/archlinux/packaging/packages/certbot/-/blob/main/PKGBUILD?ref_type=heads#L92) or [move them wherever](https://build.opensuse.org/package/view_file/home:mcepl/papis/papis.spec?expand=1).

I would like to help delete setup.py, so I've deleted the hook I made yesterday and committed building the project with hatchling if you're interested in trying it. I've removed the files that are OK to keep out of  `shared-data`, as they are included anyway inside the wheel metadata.

alexfikl:
> Thank you for taking a look and considering the changes, @alexfikl ❤️!
> 
> > So I would be more inclined to remove those things all together and let more OS-specific packaging handle them. They're all easy to copy over or generate, e.g. completions just need a call to
> 
> Here is a list of files I think are ok to **remove** from `shared-data`:

That's a good point, we can definitely just remove these.
 
> As for the rest of the files
>     * manpages: 
>     * desktop file:  
>     * completions: 

These are definitely all useful things to have. However, it seems to me that there's always extra work to do if you want them:
* If they're inside the wheel, you need to start symlinking or changing various paths
* If they're not in the wheel, you need to generate them and copy them over to wherever they need to go.

Also, my understanding is that they won't get installed to the system paths even when the package itself is installed globally -- they'll always be in `site-packages`. Maybe `pip` does some hacks to put them in a global place? Other installers such as [build](https://github.com/pypa/build) don't do that anymore.
 
> I would like to help delete setup.py, so I've deleted the hook I made yesterday and committed building the project with hatchling if you're interested in trying it. I've removed the files that are OK to keep out of `shared-data`, as they are included anyway inside the wheel metadata.

The `setup.py` can be deleted now too, it's essentially empty and `setuptools` takes all the information from `pyproject.toml`. I just left it in for some small backwards compatibility. What benefit does hatch bring here?

kiike:
That was fast!

> These are definitely all useful things to have. However, it seems to me that there's always extra work to do if you want them:

>     If they're inside the wheel, you need to start symlinking or changing various paths
>    If they're not in the wheel, you need to generate them and copy them over to wherever they need to go.

> Also, my understanding is that they won't get installed to the system paths even when the package itself is installed globally -- they'll always be in site-packages. Maybe pip does some hacks to put them in a global place? Other installers such as [build](https://github.com/pypa/build) and [installer](https://github.com/pypa/installer) don't do that anymore.

EDIT: Hm, they seem to install things properly at least in a venv, nevermind! That's weird..

Yeah it's pretty confusing! So, if I were a package maintainer from, say, Arch, I [wouldn't need to touch anything](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=papis-git#n105) after unpacking the wheel. So I agree it's a nice to have at most.

> The setup.py can be deleted now too, it's essentially empty and setuptools takes all the information from pyproject.toml. I just left it in for some small backwards compatibility. What benefit does hatch bring here?

Quoting [from the pr from the hatchling's author](https://github.com/tartley/colorama/pull/338)
> In comparison to setuptools, Hatchling has reproducible builds, supports the modern [PEP 660](https://peps.python.org/pep-0660/) editable installation standard, has a more intuitive plugin system, and the codebase is an order of magnitude smaller thus making it more auditable and less prone to bugs (setuptools is based on a [deprecated](https://peps.python.org/pep-0632/) standard library module that it now must vendor).

[Other projects ](https://archlinux.org/packages/extra/any/python-hatchling/) moved to hatchling too, I guess to avoid the setuptools' baggage. Two prominent projects that moved are jupyter and airflow.

alexfikl:
> Yeah it's pretty confusing! So, if I were a package maintainer from, say, Arch, I [wouldn't need to touch anything](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=papis-git#n105) after unpacking the wheel. So I agree it's a nice to have at most.

Agreed, I just don't understand why it works given https://github.com/pypa/installer/issues/128. 
 
> Quoting [from the pr from the hatchling's author](https://github.com/tartley/colorama/pull/338)

Setuptools supports both `pyproject.toml` and editable installs these days (and reproducible builds?), its `data-files` seem to work the same as `shared-data` in hatch, so the only benefit is the better codebase. Given how ubiquitous setuptools is and how it's seeing consistent development, that's not such a big deal..

kiike:
> > Yeah it's pretty confusing! So, if I were a package maintainer from, say, Arch, I [wouldn't need to touch anything](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=papis-git#n105) after unpacking the wheel. So I agree it's a nice to have at most.
> 
> Agreed, I just don't understand why it works given [pypa/installer#128](https://github.com/pypa/installer/issues/128).
> 
> > Quoting [from the pr from the hatchling's author](https://github.com/tartley/colorama/pull/338)
> 
> Setuptools supports both `pyproject.toml` and editable installs these days (and reproducible builds?), its `data-files` seem to work the same as `shared-data` in hatch, so the only benefit is the better codebase. Given how ubiquitous setuptools is and how it's seeing consistent development, that's not such a big deal..

Fair enough! It was just an idea given big projects do it.

alexfikl:
> Fair enough! It was just an idea given big projects do it.

It's very easy to move to hatch or pdm or flit or whatnot after this gets in, so I'm not very bothered about choosing one now, since for our purposes they mostly do the same thing anyway.

I guess the main thing holding this back is just that we agree to go for it (it's a big packaging change and I was weary of just merging it). @alejandrogallo @jghauser What do you think? :grin:

alexfikl:
@kiike I was just reading through the [Why Hatch?](https://hatch.pypa.io/latest/why/) thing and it says that the editable installs with setuptools don't work with VSCode. Is that true? That would be a very good reason to switch from it..

kiike:
> @kiike I was just reading through the [Why Hatch?](https://hatch.pypa.io/latest/why/) thing and it says that the editable installs with setuptools don't work with VSCode. Is that true? That would be a very good reason to switch from it..

It's hard to say. Editable installs with setuptools sometimes break but not only with VS Code. My current experience with VS Code is that setup.py-based projects definitely work worse (no or broken debugging and test discovery) than pyproject.toml-based ones.

I can't point to setuptools directly being the reason, but I strongly believe that setup.py has to do with the issues more than the build backend.

I am adding a feature to papis zotero and, since I was working on main of both projects, suddenly debugging doesn´t work due to setup.py. I'll move to the pyproject branch on papis, then create and move papis-zotero to a pyproject branch too. Then I'll tell you if setuptools makes it work or not.

alexfikl:
> EDIT: Right now, in my machine the problem is setup.py, not setuptools. It really irks me that both things are mixed up. I have a VS Code workspace with both papis and papis-zotero open, I edited the server and ran the program without rebuilding anything, just saving the file. And the edit worked perfectly:

Thanks for testing! I went ahead and removed the `setup.py` completely. That seems to break the nix install, but it should be fixable.

kiike:
> Thanks for testing! I went ahead and removed the `setup.py` completely.

Awesome!

> That seems to break the nix install, but it should be fixable.

It is!: https://nixos.wiki/wiki/Packaging/Python has the instructions to fix it.

kiike:
Wait. Editable installs with setuptools+wheels+pyproject.toml **are** broken. But not only in VS Code, but in my regular installation as well as an ephemeral distrobox:

I wanted to run `make mypy` on papis-zotero before my pull request a while ago, and it failed with  import-not-found. I thought i broke my environment or something, so I created a throwaway dev machine.

I recreated the setup, tried to run `make mypy` and, indeed:

image

Unusable imports. My papis-zotero venv has `papis` installed as an editable install. When I uninstalled it and installed it as a regular install, then it works.

However, same setup with hatchling works perfectly with editable installs. So it really helps in that regard.

image

If you want to test for yourself, you just need distrobox. On my machine it makes a vm with an openSUSE tumbleweed (a rolling release distro). The repos have Python 3.11 and setuptools 65 as default when you install python3.

```bash
distrobox-ephemeral --unshare-all --home /tmp/eph
cd
sudo zypper in git python3 make
git clone https://github.com/alexfikl/papis && cd papis && git checkout pyproject
git clone https://github.com/papis/papis-zotero
cat > pyproject <<EOF
[build-system]
build-backend = "setuptools.build_meta"
requires = [
    "setuptools>=69",
    "wheel",
]

[project]
name = "papis-zotero"
description = "Interact with Zotero using papis"
version = "0.1.2"
authors = [{ name = "Alejandro Gallo", email = "aamsgallo@gmail.com" }]
maintainers = [{ name = "Alejandro Gallo", email = "aamsgallo@gmail.com" }]
readme = "README.rst"
license = { text = "GPLv3" }
classifiers = [
    "Environment :: Console :: Curses",
    "Environment :: Console",
    "Intended Audience :: Developers",
    "Intended Audience :: Education",
    "Intended Audience :: Science/Research",
    "Intended Audience :: System Administrators",
    "License :: OSI Approved :: GNU General Public License v3 (GPLv3)",
    "Operating System :: MacOS",
    "Operating System :: Microsoft",
    "Operating System :: OS Independent",
    "Operating System :: POSIX",
    "Operating System :: Unix",
    "Programming Language :: Python :: 3 :: Only",
    "Programming Language :: Python :: 3.8",
    "Programming Language :: Python :: 3.9",
    "Programming Language :: Python :: 3.10",
    "Programming Language :: Python :: 3.11",
    "Programming Language :: Python :: 3.12",
    "Topic :: Utilities",
]
keywords = ["bibtex", "biliography", "cli", "management", "papis", "zotero"]
dependencies = ["papis==0.13"]

[project.urls]
Repository = "https://github.com/papis/papis-zotero"

[project.optional-dependencies]
"develop" = [
    "flake8",
    "flake8-bugbear",
    "flake8-quotes",
    "mypy>=0.7",
    "pep8-naming",
    "pytest",
    "pytest-cov",
    "python-coveralls",
    "types-PyYAML",
    "types-tqdm",
]

[project.entry-points."papis.command"]
"zotero" = "papis_zotero:main"

[tool.setuptools]
packages = ["papis_zotero"]
EOF
rm setup.py
python3 -m venv .venv && source .venv/bin/activate
pip install -e /tmp/eph/papis
pip install -e .
make mypy
```

It will fail.

Now, change both pyproject.toml's build-system to:

```
[build-system]
build-backend = "setuptools.build_meta"
requires = [
    "setuptools>=65",
    "wheel",
]
```
Next,
```
pip uninstall papis
pip uninstall papis-zotero
pip install -e /tmp/eph/papis
pip install -e .
```

It will work.

alexfikl:
> FINAL EDIT (this is fun): It's a [known issue with mypy+setuptools editable installs](https://github.com/python/mypy/issues/13392)

Can confirm this is actually happening and it's very annoying. I wasn't seeing it because I had papis installed globally and it picked up the `py.typed` from there (I guess?).

It does seem like `setuptools` has quite a few annoying bug around editable support and that sort of stuff, so I'm all for chancing to `hatchling` instead. I'll go ahead and cherrypick your commit in here! 

Thanks again for looking into this and testing!

jghauser:
I had a quick look at the whole python build landscape, and while I have to say that it's a little over my head, I think moving to `pyproject.toml` and `hatchling` is a solid choice. We wanted to move to `pyproject.toml` anyway (which I'm looking forward to, as it should make nix packaging quite a bit smoother) and if there are (as you've discussed above) issues with `setuptools`, then why not modernise things in one fell swoop. `hatchling` seems already well established and a pretty future-proof choice. Thanks you two for all the work figuring this out :heart:

alexfikl:
Thank you everybody for taking a look at this and testing! I'll go ahead and merge it :rocket:

kiike:
That's awesome! Thanks @alexfikl! :heart:

PR #612, opened 2023-07-23
Enabled two new CI workflows to build, test, (and release) containers.

f0rki:
Test the nix and docker-based install methods.

* `container-build.yml` - does the docker magic
    * build a docker container with the default `Dockerfile` and check whether `make ci-test` works.
    * build a "release" container that is pushed to `ghcr.io/papis/papis`, i.e., papis packaged as a container.
        * this container is tested by running a couple of papis commands `tools/run-integration-test.sh`
* `nix-flake.yml` - does the nix magic
    * runs `nix build` - checks the flake build
    * runs `nix develop` - checks whether pytest passes in the nix development environment.

Packaging papis in docker is a nice thing to have and quite easy with github's container registry (`ghcr.io`). For example, if you like to run `papis serve` on your home server this might be a convenient thing to run.


.github/workflows/container-build.yml
```diff
@@ -0,0 +1,91 @@
...
+        # script to avoid accidentally destroying files of folks running this
+        # script.
+        run: >
+          docker run \
+            --rm \
+            -v $(realpath ./tools/):/tools/ \
+            -e PAPIS_CI_RUN=42 \
+            --entrypoint /bin/bash \
+            ${{ fromJSON(steps.meta.outputs.json).tags[0] }} \
+            /tools/run-integration-test.sh
```
alexfikl, line 80: What's the difference between this and just running the tests above?

.github/workflows/container-build.yml
```diff
@@ -0,0 +1,91 @@
+name: Build/Release Container Image
+
+on:
+  push:
+    branches: [ "main", "ci-*" ]
```
alexfikl, line 5: What's the `ci-*` branch? Does that get created by the docker action?

.github/workflows/main.yml
```diff
@@ -16,7 +19,8 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: ['3.8', '3.9', '3.10', '3.11']
-      fail-fast: False
+      # fail fast only on pull-requests!
+      fail-fast: ${{ github.event_name == 'pull_request' }}
```
alexfikl, line 23: Why this change? The CI often fails on Windows but not Linux or on some python versions and not others, so it's very useful to let it all run.

.github/workflows/main.yml
```diff
@@ -1,9 +1,12 @@
...
+name: CI Testing Pipeline
 
 on:
   push:
-    branches: [ "main" ]
+    # run for all main branch commits and on ci-* branches for testing new CI
+    # features.
+    branches: [ "main", "ci-*" ]
   pull_request:
+    # run on all pull requests against main.
```
alexfikl, line 9:
```suggestion
    branches: [ "main", "ci-*" ]
  pull_request:
```
? These comments are a bit redundant and just add noise.

.github/workflows/nix-flake.yml
```diff
@@ -0,0 +1,22 @@
+name: "Test ./flake.nix"
```
alexfikl, line 1:
```suggestion
name: "Test flake.nix"
```

.github/workflows/nix-flake.yml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: cachix/install-nix-action@v22
+      with:
+        github_access_token: ${{ secrets.GITHUB_TOKEN }}
+    - run: nix build
+    - run: nix flake check
+    # for some reason in nix develop we lack type definitions for python-doi
+    # so the mypy checks fail.
```
alexfikl, line 21: `python-doi` has had annotations for a long time and should install the `py.typed` file. Guessing this is some nix issue?

tests/test_config.py
```diff
@@ -53,7 +53,7 @@ def test_get_config_dirs(tmp_config: TemporaryConfiguration, monkeypatch) -> Non
         dirs = papis.config.get_config_dirs()
         assert os.environ.get("XDG_CONFIG_DIRS") is None
         assert len(dirs) == 2
-        assert os.path.join("/", "tmp", "papis") == dirs[0]
+        assert os.path.join(tmpdir, "papis") == dirs[0]
```
alexfikl, line 56: Nice catch!

alexfikl (commented):
Left a few questions, but overall this looks good to me!

Does this require some special access to publish the Docker images?

.github/workflows/container-build.yml
f0rki: This uses the papis cli as it is installed. The second container (release.Dockerfile) does not have pytest, etc. installed.

.github/workflows/container-build.yml
f0rki: It is just to test CI thing, i.e. any branch with the `ci-` prefix can be used to develop/test new CI things.

.github/workflows/main.yml
f0rki: Ok I can remove those.

.github/workflows/main.yml
f0rki: Ok I wanted to reduce redundant failures, e.g. because PRs forgot to format or because of some trailing whitespace. But I can remove that.

alexfikl: That's a fair point, but probably better handled by just separating the static checks (`flake8`, `mypy`) and only running `pytest` if those succeed.

papis tests are pretty fast, so we never bothered too much :\

f0rki:
Yeah I think you have to enable that workflows can write to the repo/registry in the Settings.

.github/workflows/nix-flake.yml
f0rki: Yes, this is a nix specific issue. For now I suggest to leave it like this. Sometimes nix is a bit stubborn...

.github/workflows/main.yml
f0rki: ok I separated them. Now the lints are run first on `ubuntu-latest` with latest python in a separate step before running all tests/lints with the matrix job.

alexfikl:
I'm sorry for letting this fall off the radar for quite a while now. Maybe we can get some of it in!
From a quick skim:
1. I don't have access to add the required Docker permissions to make that workflow work, so that will have to wait.
2. Splitting the linting in the main CI doesn't sound like a good idea. Time-wise, the CI is pretty fast and the OS/version matrix takes the most time. Otherwise, it's better to pass linting on all versions we support so contributors can have a clean lint even if their system is not `ubuntu-latest`.
3. There are many unrelated fixes for various things in here (from ruff?).
4. The nix stuff is done.

3 and especially 4 can go in right away, but they will have to be in separate PRs, since I can't merge this wholesale. Do you have any time to work on these? If not, I can slowly cherry pick stuff and merge it.

f0rki:
> If not, I can slowly cherry pick stuff and merge it.

Yes feel free to cherry pick the commits you think are worth merging now.

f0rki:
> There are many unrelated fixes for various things in here (from ruff?).

Yes there is a commit that adds things from `ruff --fix`

alexfikl:
@f0rki Squashed and rebased this to only contain the docker stuff, since everything else should be in main already.

alexfikl:
This seems to be currently failing: https://github.com/f0rki/papis/actions/runs/6040713352/job/16392121597 (possibly my bad :(

f0rki:
rebased and pushed fixes. should work now again.

PR #613, opened 2023-07-25
batch update add.py

Twix53791:
when batch option is set, select by default the first importer/downloader if needed without interrupting the process

### DIFF
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -596,12 +596,12 @@ def cli(files: List[str],
     matching_importers = papis.utils.get_matching_importer_by_name(
         from_importer, only_data=only_data)
 
-    if not from_importer and not batch and files:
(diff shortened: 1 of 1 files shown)


papis/commands/add.py
```diff
@@ -613,6 +613,11 @@ def cli(files: List[str],
                 "Select matching importers (for instance 0, 1, 3-10, a, all...)")
 
             matching_importers = [matching_importers[i] for i in matching_indices]
+        elif matching_importers:
+            if len(matching_importers) >= 2:
+                matching_importers = [matching_importers[1]]   
+            else:
+                matching_importers = [matching_importers[0]]
```
alexfikl, line 620: Can you explain the logic behind this? I'd expect that in batch mode, we just merge all matching importers.

PR #614, opened 2023-07-26
Avoid exception when parsing crossref event data that has no 'start' key

f0rki:
Along the way fixed format string that contained a '%s' without corresponding arg.


alejandrogallo:
This looks good, are you ready to merge?

f0rki:
Yep :) go ahead with merging!

alexfikl:
> Yep :) go ahead with merging!

Thanks for fixing this! In it goes :rocket:

PR #616, opened 2023-07-27
commands.add: --set options after importer data

Twix53791:
--set options now overwrite the imported data via importer


Twix53791:
What about this changes @alexfikl ? We merge it or not?

jghauser:
Sorry to take a while to get to this. While I've read through the issue, I'm still not super sure I get exactly what is going on.

The way I understood things, the problem is that the `--set` stuff doesn't have priority over the info from the importers? Though @alexfikl, you're saying:
> In that case, I think it makes sense to add the values from --set first and then update them as needed with data from importers.

What do you mean by this? That the stuff added with `--set` should only have an effect if the importer doesn't also provide information regarding the same key? That's I think what I'm seeing after a quick test. If I `--set author something`, then this will disappear if the importer finds another author name. I think I agree with @Twix53791 that this is unexpected -- I'd think that user supplied values should always have precedence. Maybe with the option, in the interactive merger, of picking the importer's information.

And yeah, I also think the merger thing isn't all that easy to manage. I don't think I've ever used it. Would be cool to come up with more ergonomic alternatives/complements! :)

alexfikl:
> What do you mean by this? That the stuff added with `--set` should only have an effect if the importer doesn't also provide information regarding the same key? That's I think what I'm seeing after a quick test. If I `--set author something`, then this will disappear if the importer finds another author name.

If I recall correctly, my point there was that in the interactive case, we always get to choose between the importer and the user data, so it doesn't matter if it's set before or after. Not a great argument though..

> I think I agree with @Twix53791 that this is unexpected -- I'd think that user supplied values should always have precedence. Maybe with the option, in the interactive merger, of picking the importer's information.

Yeah, I'm starting to think that's the better idea too, so let's go with that!

> And yeah, I also think the merger thing isn't all that easy to manage. I don't think I've ever used it. Would be cool to come up with more ergonomic alternatives/complements! :)

I haven't used it either :( My usual workflow is just to pick an importer on `papis add` and then `papis edit` what I need.

alexfikl:
@Twix53791 @jghauser Made some changes like discussed (@Twix53791 sorry for the back and forth on this!). So now:
* In `--batch` mode, it should merge all the importers and overwrite with the `--set` data provided by the user. We could make the merging smarter here, but that's for another time.
* In interactive mode, it should just ask the user to interactively pick the fields with the `diff` widget. It should now skip that when there's just one (or zero) importer, but otherwise it's the same as before.

It needs a bit of testing, but this looks very reasonable! @Twix53791 Does this work for you?

Twix53791:
I have tested your changes @alexfikl and it works well for me! As I see, with --batch the user --set data overwrite the importer data, and with the interactive mode, if we just "add all", also, so the only way to overwrite the user data with the importer data is to manually reject it in the "split" process of the interactive mode.
I think it is ready to merge. Thank you both for looking at this problem!

> And yeah, I also think the merger thing isn't all that easy to manage. I don't think I've ever used it. Would be cool to come up with more ergonomic alternatives/complements! :)

I am working on it [in this pull request](https://github.com/papis/papis/pull/640) ! I have to fix some bugs and after it will be ready to be reviewed and discussed with you all...

alexfikl:
> I have tested your changes @alexfikl and it works well for me! As I see, with --batch the user --set data overwrite the importer data, and with the interactive mode, if we just "add all", also, so the only way to overwrite the user data with the importer data is to manually reject it in the "split" process of the interactive mode. I think it is ready to merge. Thank you both for looking at this problem!

Awesome! Thanks for all your effort in getting this figured out! :rocket:

alexfikl:
> I am working on it [in this pull request](https://github.com/papis/papis/pull/640) ! I have to fix some bugs and after it will be ready to be reviewed and discussed with you all...

Maybe open an issue regarding that so that we can discuss a bit more generally? There's also #628 with a different approach, so maybe it's worth considering some sort of plugin system for the diff widget.

PR #620, opened 2023-08-04
Tentative implementation of papis-init

alejandrogallo:
I think we would benefit of having a clearer workflow for newcomers
when initializing a library.

Something similar to git would be a better initiation for kickstarting
a papis library.

This is a pull request / discussion to decide what is the best way
forward in this.

For now, you go to the dir

```
mkdir -p ~/Documents/papers
cd ~/Documents/papers
```

and you can then

```
papis init
```

and interactively answer the questions being prompted.
This then updates the configuration file.

You can use the `init` command to then change more options
in the future.

Also it can be a good gateway to tell users
about the most useful configuraiton settings, and they can set them
righ away.

### DIFF
--- a/papis/commands/init.py
+++ b/papis/commands/init.py
@@ -26,7 +26,48 @@ import papis.logging
 
 logger = papis.logging.get_logger(__name__)
 
-INIT_PROMPTS = ["opentool", "editor", "use-git", "notes-name"]
+INIT_PROMPTS = [
+    {
+        "key": "opentool",
+        "help": "Which program should be used to open files? e.g. evince"
+    },
+    {
+        "key": "editor",
+        "help": "Which program should be called when editing info.yaml files?"
+    },
+    {
+        "key": "use-git",
+        "help":
+        "Use git in some papis commands to issue commits automatically?"
+    },
+    {
+        "key": "notes-name",
+        "help": "The name of the file for notes."
+    },
+    {
+        "key": "database-backend",
+        "help": "You can choose between different kinds of databases."
+    },
+    {
+        "key": "bibtex-unicode",
+        "help": "Do you want to allow unicode in an exported bibtex?"
+    },
+    {
+        "key":
+        "ref-format",
+        "help":
+        "When adding a new document, "
+        "how would you like the reference string be built?"
+    },
+    {
+        "key": "multiple-authors-format",
+        "help": ""
+    },
+    {
+        "key": "citations-file-name",
+        "help": ""
+    },
+]
 
 
 @click.command("init")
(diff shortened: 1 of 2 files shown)


alejandrogallo:
that's good, we have to decide about the API, maybe given a path it should get the basename as a first name for the library and the path as its path, for instance.

That makes sense.

We can also referr at the end of the command to the link of rtd for people to read more, and give a couple of test commands for them to start using it.

alexfikl:
> that's good, we have to decide about the API, maybe given a path it should get the basename as a first name for the library and the path as its path, for instance.

That sounds good to me! It's important for these things to have good defaults so the user could plausibly just spam `<Enter>` a few of times and everything gets created and set to expected values.

> We can also refer at the end of the command to the link of rtd for people to read more, and give a couple of test commands for them to start using it.

Yes!

papis/commands/init.py
```diff
@@ -0,0 +1,87 @@
...
+
+    if papis.tui.utils.confirm(
+            "Make '{}' the default library?".format(library_name)):
+        glob["default-library"] = library_name
+
+    local["dir"] = library_path
+
+    for setting in INIT_PROMPTS:
+        local[setting] = papis.tui.utils.prompt(
+            "{}: ".format(setting),
```
alexfikl, line 79:
```suggestion
            "Set library-local '{}': ".format(setting),
```
? Or something more verbose to explain what this does. Does `papis.tui.utils.prompt` autocomplete paths at all? That's always super useful for things like `opentool`.

papis/commands/init.py
```diff
@@ -0,0 +1,87 @@
...
+import click
+
+import papis.utils
+import papis.config
+import papis.tui.utils
+import papis.logging
+
+logger = papis.logging.get_logger(__name__)
+
+INIT_PROMPTS = ["opentool", "editor", "use-git", "notes-name"]
```
alexfikl, line 29: Not sure what else to put here, so I'll just list a few that don't sound too ridiculous (not advocating anything in particular): `database-backend`, `bibtex-unicode`, `ref-format`, `multiple-authors-separator`, `multiple-authors-format`, `citations-file-name`.

Maybe `ref-format` is worth adding, since different libraries could presumably have different preferences for that?

alexfikl (commented):
Looked a bit through the code and added a few comments. Overall this already looks very helpful! Adding some links to readthedocs and other sources like you mentioned would be great :grin:

papis/commands/init.py
```diff
@@ -0,0 +1,87 @@
...
+    config = papis.config.get_configuration()
+    defaults = papis.config.get_default_settings()
+    config_file = papis.config.get_config_file()
+
+    if os.path.exists(config_file):
+        logger.warning(
+            "Config file %s already exists,"
+            " this command may change some of its contents,", config_file)
+        if not papis.tui.utils.confirm("Do you want to continue?"):
+            return
```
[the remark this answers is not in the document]
alejandrogallo: AFAIK comments might disappear, I think, and indentation might also be different when written, this might freak out some OCD folks like me, but I can test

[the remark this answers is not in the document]
alexfikl: Hm, that's a good point. I didn't think about comments.. then yes, a nice big warning is likely a good idea!

papis/commands/init.py
alejandrogallo: I added most of these settings, I think they are meaningful, I also added a sentence to every setting to give more context to the user.

Now I noticed that in the future it would be nice if we could define the options and the documentation in the same file, then I would have been able to recycle the documentation of the settings here. But it is not super important.

PR #622, opened 2023-08-06
Fix mypy errors in papis.tui

alexfikl:
This was a chunk of code that was not getting checked before. This fixes all the `mypy` errors in there and and makes the remaining `type: ignore` have an actual error code attached so we catch new errors.

### DIFF
--- a/tests/commands/test_cache.py
+++ b/tests/commands/test_cache.py
@@ -1,6 +1,5 @@
 import os
-import sys
-import pytest
+import time
 
 from papis.commands.cache import cli
 import papis.database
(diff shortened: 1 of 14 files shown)


tests/commands/test_cache.py
```diff
@@ -59,30 +57,36 @@ def test_rm_add_update(tmp_library: TemporaryLibrary) -> None:
...
-    query_results = db.query_dict({"cache": "end-testing"})
+    query_results = db.query_dict({"cache": "test-update-newer"})
     assert len(query_results) == 0
 
     result = cli_runner.invoke(cli, ["update-newer", "--all"])
     assert result.exit_code == 0
 
-    query_results = db.query_dict({"cache": "end-testing"})
+    query_results = db.query_dict({"cache": "test-update-newer"})
     assert len(query_results) == 1
```
alexfikl, line 92: @alejandrogallo I managed to hit this error on my Linux machine too. As far as I can tell, it seems to have just been an issue with `st_mtime` not catching the changes fast enough.

Adding a little `time.sleep(1)` gives an updated `mtime` and makes the test pass. There's also a `st_mtime_ns` in nanoseconds that we could use here, if it becomes an issue.

PR #625, opened 2023-08-09
crossref: remove use of interactive pick_doc

Twix53791:
--batch option set _all to True (avoid the menu asking to manually select the entries to update if multiple matches)

### DIFF
--- a/papis/commands/update.py
+++ b/papis/commands/update.py
@@ -119,6 +119,9 @@ def cli(query: str,
         set_tuples: List[Tuple[str, str]],) -> None:
     """Update a document from a given library."""
 
+    if batch:
+        _all = True
+          
     documents = papis.cli.handle_doc_folder_query_all_sort(query,
                                                            doc_folder,
                                                            sort_field,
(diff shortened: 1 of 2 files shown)


alexfikl:
Your branch seems to be out of date: ffccd0e is already committed and the last two commits seem unrelated. Can you clean this up so that only the relevant commit 47159a5 remains?

Twix53791:
It is not 'unrelated', in fact, adding the batch option to update just to set _all to true is useless as we can use -a option instead.
My purpose, adding the batch option to update, was to use --auto option without prompting a menu.
It is really necessary? That's true than using `add --from doi` gives almost the same result, but not always. From Science direct, 'add from doi' doesn't grab the abstract, rather than 'add url + update --auto' yes... 
BUT
As you say in the issue related to this pull request, in fact, the `pick_doc` is a bug.
SO
I propose now to forget about this --batch option in update.py, and rather just to correct the bug in crossref.py

alexfikl:
Thanks for the nice fix! In it goes :rocket:

PR #626, opened 2023-08-09
Update doc_folder option in cli.py

Twix53791:
--doc-folder option now accept a large string containing multiples folders 
1) the option doesn't check anymore if the folder is valid 
2) handle_doc_folder_or_query now split the string by spaces/single/doubles quotes to get a list of folders


alexfikl:
This looks good to me! Can you check out the failing test cases? Let me know if you need any help there.

Twix53791:
I think its done. Explanations.
- add.py modified by me was sending, from the tuple list of doc_folder, each folder (a string) to from_folder (document.py)
- but some others commands (run, explore, etc.) were sending the tuple, and from_folder need a string.
So, I think the best solution is to only change `from_folder` function in document.py, to make i handles a tuple OR a string, both.
```python
    if not isinstance(folder_path, str):
    # If it not a string, try to get the first item of the tuple
        try:
            folder_path = folder_path[0]
        except Exception:
        # If not (the tuple is empty I guess, I didnt achieve to reproduce the error)
        # set folder_path to an empty string (maybe folder_path = str(folder_path) is better?)
            folder_path = ""
```

alexfikl:
Looking at this a bit more, I think it needs more work to finish up.

First, this started from enhancing `papis update`, right? For that, you'll need to also update the type annotations here
https://github.com/papis/papis/blob/a82fcac13786124e9bc31741d39fba5324c41925/papis/commands/update.py#L112
to say `doc_folder: Tuple[str, ...]`. Unfortunately, `mypy` doesn't seem to be smart enough to catch that, but once you change it there it will start complaining a lot about all the places that's used, starting with `papis.cli.handle_doc_folder_query_all_sort`.

Second, the `doc_folder_option` decorator is used in a bunch of other commands: `addto`, `browse`, `cache`, `citations`, `doctor`, `edit`, `explore`, `export`, `list`, `mv`, `open`, `rename`, `rm`, `run`, `update` (from a quick grep). I think most of them use `handle_doc_folder_query_sort` or some version of that, so probably just need to take a look and see if they're ok.

Third, that fix in `papis.document.from_folder` should not be necessary. Ideally we'd rather fix the call sites and leave this function simple as it is.

That's quite a bit of work, so let me know if you need any help!

Twix53791:
`ripgrep` is my friend !

It is strange because it works perfectly for me, `papis update -a --doc-folder folder1 --doc-folder folder2` doesn't produce any errors.
update.py:
```
def cli(query: str,
...
        doc_folder: str,
...
    print(str(type(doc_folder)))
    >>>> <class 'tuple'>
```

alexfikl:
> ```
> def cli(query: str,
> ...
>         doc_folder: str,
> ...
>     print(str(type(doc_folder)))
>     >>>> <class 'tuple'>
> ```

This actually showcases the error: the type is declared as `str` and the output is `tuple`. The type annotations are not checked at runtime for correctness, so that's why this works anyway.

Twix53791:
okok. I do it quickly with `repgrep`...

Twix53791:
oups... I mess a bit with git...

PR #628, opened 2023-08-09
Diff show in fzf

Twix53791:
Add a new function, difffzf, in diff.py. When add entries to papis library, show diff between sources in a fzf menu


Twix53791:
It is not working well now... Messing the yaml formatting... I will check that later...

alexfikl:
This is pretty old now with no progress. Feel free to re-open if you get a chance to work on it again.

PR #630, opened 2023-08-10
commands.add: fixes for --batch option

Twix53791:


papis/commands/add.py
```diff
@@ -455,6 +456,10 @@ def run(paths: List[str],
             "existing document.")
 
         # NOTE: we always want the user to confirm if a duplicate is found!
+        # Except when batch option is set
+        if batch:
+            logger.warning("No document has been created, since batch option exit when a duplicate is found.")
+            return
         confirm = True
```
alexfikl, line 463:
```suggestion
        # except when batch option is set
        confirm = not batch
```
? Is this what you intended here? The `return` in your current code definitely shouldn't be there.

papis/commands/add.py
```diff
@@ -613,6 +618,11 @@ def cli(files: List[str],
                 "Select matching importers (for instance 0, 1, 3-10, a, all...)")
 
             matching_importers = [matching_importers[i] for i in matching_indices]
+        elif matching_importers:
+            if len(matching_importers) >= 2:
+                matching_importers = [matching_importers[1]]   
+            else:
+                matching_importers = [matching_importers[0]]
```
alexfikl, line 625: From the previous pull request #613:

Can you explain the logic behind this? I'd expect that in batch mode, we just merge all matching importers.

papis/commands/add.py
Twix53791: Yes! Better. I am not used to python I did'nt think about this trick. The goal was simple : when run with the batch option, add should never overwrite an existing entry.

papis/commands/add.py
Twix53791: Hum, it is right we could just merge all the importers yes! I was thinking just select one (the 1.) but in fact it could be betetr to select all. Does it not have a bad side effect? A possibility of messing things? If we merge all the importers:
1) my last commit hacking utils.py to set the fallback downloader always first is useless
2) We can just remove this block of code you quote. So if batch, the matching_importers are not modified (all are selected).

alexfikl: The system isn't quite that smart, so I'm not quite sure about the effect of merging data from multiple importers automatically. I haven't used it like that, but I would expect that
* Just a few importers match for a given query, e.g. giving a DOI won't match the other importers.
* The importers are generally pretty solid sources (as opposed to the downloaders), so even if they overwrite data, it should be relatively ok.

papis/commands/add.py
alexfikl: This won't do that, as far as I know. `confirm = True` means asking the user if the new document should be added and doesn't consider duplicates. In effect, this will always *overwrite* when `batch = True`.

Twix53791: You think? It is strange because it works when I test it. But that's true than the 'confim' options has not really this purpose.
But we agree than this block of code is about the duplication checking?? Maybe we can just return when the batch option is set?

```diff
    try:
        found_document = papis.utils.locate_document_in_lib(tmp_document)
    except IndexError:
...
    else:
+        if batch:
+            return
        click.echo("The following document is already in your library:")
...
        # NOTE: we always want the user to confirm if a duplicate is found!
        confirm = True
```

By the way, I just remember than in fact batch option set confirm always to false:
```
609-613:
    if batch:
        edit = False
        confirm = False
        open_file = False
```

alexfikl: Yeah, you're completely right! 

However, I'm not sure that never overwriting is a good choice. Overwriting duplicates with the new version seems like a reasonable way to go as well, so we might want to add a flag for that?

Twix53791: well, maybe, in fact, now, it is not overwriting the entry but adding another one (a duplicate), its a problem, after you have twice the same item in you library, its a mess. I don't think many people want that. But we can ad a flag yes if you think its a good thing... 
I think overwriting the entry with add is useless, as the info message tells you:
"Hint: Use the 'papis update' command instead to update the existing document."
To 'overwrite' an existing entry with new data, we can use `update`.
To add another entrey (duplicate of another one), we can use `add`. 
Now, is there a case a guy wants to add a duplicate with `add -b` ? Quite rare I think, who 1) wants twice the same entry 2) has to run `add` in the background?
Remember than we have to make a choice. We can let `add -b` prompt the user to ask confirmation, as we lose the all point of --batch option : running thing quietly in the background. 
If really you think its necessary, lets add a flag --force, forcing adding the entry in any case, and then:

```python
        if batch and not force: 
            logger.info("No document is created.") 
            return
```

alexfikl: > well, maybe, but in fact, now, it is not overwriting the entry but adding another one (a duplicate), its a problem, after you have twice the same item in you library, its a mess. 

Hm, if it actually does that, that's a bug indeed. It shouldn't add the same document twice.

> To 'overwrite' an existing entry with new data, we can use `update`. To add another entry (duplicate of another one), we can use `add`. 

That's a good point, we do have the `update` command to do that, so maybe the `add` command doesn't have to be that smart. I'll have to think about it a bit :confused:

alexfikl:
With some fresh eyes, I agree with your suggestions to never overwrite when using `--batch`. I added a more verbose message in that case, but otherwise this looks good to me!

Thanks for all the work and explanations! :rocket:

papis/commands/add.py
alexfikl: Yeah, I've come around on this too. Overwriting is a bit mean and we don't know how to do a smart merge of the new and existing documents if they're "duplicates", so better just let the user know and skip that document.

Thanks for your patience thinking through this! :heart:

PR #631, opened 2023-08-10
add.py : merging downloaders and crossref importers

Twix53791:
This option print on the stdin the 'out_folder_path' (the folder newly created in the papis library). usefull to grab the new path created in another program...


Twix53791:
Another way of achieve my purpose would be:
1) add a `quiet` option to `add`, disabling all the logger.messages (so the stderr is clear of content)
2) `--print-folder` print the folder to stderr
3) so `papis add --print-folder url 2>/file` print the papis folder to /file. 

Its a by clearer in the code, but a bit less reliable, as maybe we can capture unexpected errors messages. But this can be countervail by parsing the file and keep only what is a valid directory.

alexfikl:
> Another way of achieve my purpose would be:
> 
>     1. add a `quiet` option to `add`, disabling all the logger.messages (so the stderr is clear of content)
> 
>     2. `--print-folder` print the folder to stderr
> 
>     3. so `papis add --print-folder url 2>/file` print the papis folder to /file.
> 
> 
> Its a by clearer in the code, but a bit less reliable, as maybe we can capture unexpected errors messages. But this can be countervail by parsing the file and keep only what is a valid directory.

If you're willing to scrape a log file, you can also try out
```
papis --logfile log.txt add url
```
It should write out all the logging messages to that file instead of to `stderr`. Would that work? The output folder is already printed at
https://github.com/papis/papis/blob/a82fcac13786124e9bc31741d39fba5324c41925/papis/commands/add.py#L384

Twix53791:
I did'nt noticed this option! It could work, but not absolutly satisfying, because I can't get anymore the log.infos...

In fact, my goal is to chain commands:
```
papis add --print-folder /tmp/folder url
papis update --auto --doc-folder $(< /tmp/folder)
papis citations -c --doc-folder $(< /tmp/folder)
```
Why? My 'goal of the goal':
1) add from doi + abstract (the abstract seems not always imported via doi, or badly (spelling, typo errors...)
2) build a citations file every time I add an entry

I am trying to use an extended 'add process' which add the maximum possible of informations when I add something.
But maybe there is already a way tocdaccounts@protonmail.com chain commands ?

PS:
1) I noticed the `--fetch-citations` option. It could achieve one of my goal, but it seems to not working properly. At least, when I run `papis add --fetch-citations --from doi <doi>` it doesnt work. Maybe I can try to fix that
2) Still missing the abstract. Why there is no easy way to combine the downloaders and the doi importer? Almost all the time, the downloader (sciencedirect, frontiers, etc) got the doi, so we could chain the downloader and the doi importer, to get the maximum of informations!

SO, finally, I think I have a proposal:
Design proposal:
1) make --fetch-citation works properly
2) add a way to combine the doi importer with the websites specific downloaders
    It could be:
        * `--from doi url` : when a non doi url is passed to doi, it first run a downloader if one if found, then get the doi, then run the doi importer
        * `--fetch-doi url` a new option getting the doi from the downloader then running the doi importer

Twix53791:
I have plenty of new ideas ! I am closing this pull request and open new ones...

PR #634, opened 2023-08-12
commands.add: fix --fetch-citation flag

Twix53791:


Twix53791:
--fetch-citations bug:

```
    print('::::::: ' + str(fetch_citations))
    print('::::::: ' + str(papis.config.getboolean("add-fetch-citations")))
...
papis add --fetch-citations url
::::::: False
::::::: True
papis add url
::::::: True
::::::: True

papis add --fetch-citations url
::::::: False
::::::: False
papis add url
::::::: False
::::::: False
```

When 'add-fetch-citations' is set to True, it works : `fetch_citations` is set to the opposite.
When 'add-fetch-citations' is set to False : bug ! `fetch_citations` is always false.

`default=papis.config.getboolean("add-fetch-citations")` seems to fix the bug. I also tried `default=lambda: True if papis.config.getboolean("add-fetch-citations")` else False` but it doesn't work.

papis/commands/add.py
```diff
@@ -549,7 +549,7 @@ def run(paths: List[str],
     is_flag=True)
 @click.option("--fetch-citations",
               help="Fetch citations from doi",
-              default=lambda: papis.config.getboolean("add-fetch-citations"),
+              default=papis.config.getboolean("add-fetch-citations"),
```
alexfikl, line 552: This isn't the correct fix because the initial goal was to do a lazy binding to the call to `papis.config.getboolean` so that it always gets the value at runtime.

I couldn't find anything in the docs about this specific usage, but from looking at the code at
https://github.com/pallets/click/blob/388b0e14093355e30b17ffaff0c7588533d9dbc8/src/click/core.py#L2558-L2582
it seems like it sets the `flag_value = not self.default` which is always false if the default is a callable like here. Setting `flag_value=True` seems to fix it as well.

I've checked that and it seems to work. Does it work for you too?

```suggestion
              default=lambda: papis.config.getboolean("add-fetch-citations"),
              flag_value=True,
```

alexfikl: We probably need to check other flags in papis because they probably fail the same way :cry:

Twix53791: Hum, I think it fails! Did you check with changing the `add-fetch-citations` in the papis config? For me:

`add-fetch-citations = True`

```
archx@ /usr papis/commands $ papis add --fetch-citations https://www.frontiersin.org/articles/10.3389/fmicb.2023.1085176/full
::::::: True
::::::: True
archx@ /usr papis/commands $ papis add https://www.frontiersin.org/articles/10.3389/fmicb.2023.1085176/full  
::::::: True
::::::: True
```
`add-fetch-citations = False`

```
archx@ /usr papis/commands $ papis add --fetch-citations https://www.frontiersin.org/articles/10.3389/fmicb.2023.1085176/full
::::::: True
::::::: False
archx@ /usr papis/commands $ papis add https://www.frontiersin.org/articles/10.3389/fmicb.2023.1085176/full  
::::::: False
::::::: False
```

It works if the setting is set to False, but not if its True...

add.py :
```python
@click.option("--fetch-citations",
              help="Fetch citations from doi",
              default=lambda: papis.config.getboolean("add-fetch-citations"),
              flag_value=True,
              is_flag=True)
```

alexfikl: That looks like the desired result: `--fetch-citations` defaults to `add-fetch-citations` and overwrites it if necessary. What did you expect the results to be?

Twix53791: Ah! I see. I was tinking, from the behavior before, than the --fetch-option was setting to the opposite of add-fetch-citations... But OK, in fact is more logical like that, fetch_citations is set to the add-fetch-citations value by default, and --fetch-citation overwrites it... So I think we have done!

alexfikl:
Great catch with this! Thank you :rocket:

PR #635, opened 2023-08-13
doc_folder_option handles multiple folders

Twix53791:
xref #626

### DIFF
--- a/papis/cli.py
+++ b/papis/cli.py
@@ -96,7 +96,8 @@ def handle_doc_folder_or_query(
         :func:`papis.document.from_folder`).
     """
     if doc_folder:
-        return [papis.document.from_folder(doc_folder)]
+        doc_folder_list = [papis.document.from_folder(f) for f in doc_folder]
+        return doc_folder_list
     return papis.database.get().query(query)
 
 
@@ -125,7 +126,7 @@ def handle_doc_folder_query_sort(
 
 def handle_doc_folder_query_all_sort(
         query: str,
-        doc_folder: str,
+        doc_folder: Tuple[str, ...],
         sort_field: Optional[str],
         sort_reverse: bool,
         _all: bool) -> List[papis.document.Document]:
--- a/papis/cli.py
+++ b/papis/cli.py
@@ -57,6 +57,7 @@ def doc_folder_option(**attrs: Any) -> DecoratorCallable:
         "--doc-folder",
         default=None,
         type=click.Path(exists=True),
+        multiple=True,
         help="Document folder on which to apply action",
         **attrs)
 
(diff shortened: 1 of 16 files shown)


Twix53791:
1) Not sure about the changes in explore.py/cache.py/open.py (their structure was different...).

2) Question :  in cli.py need I to make theses changes? (and it is the good syntax, Optional[tuple[]] ?) 
```diff
def handle_doc_folder_or_query(
        query: str,
-        doc_folder: Optional[str]) -> List[papis.document.Document]:
+        doc_folder: Optional[Tuple[str, ...]]) -> List[papis.document.Document]:
...
def handle_doc_folder_query_sort(
        query: str,
-        doc_folder: Optional[str],
+        doc_folder: Optional[Tuple[str, ...]],
```

alexfikl:
> 1. Not sure about the changes in explore.py/cache.py/open.py (their structure was different...).

* explore: just need to do a `[papis.document.from_folder(d) for d in doc_folder]` like you did in `cli.py`.
* cache-update: should be good to go with changes to `Tuple[str, ...]`.
* cache-add: you'll need to wrap the whole thing in a `for d in doc_folder` loop.
* cache-rm: same as cache-update.
* cache-update-newer: same as cache-update.
* open: should only need the change to `Tuple[str, ...]` as well?

Let me know if I missed anything.

> 2. Question :  in cli.py need I to make theses changes? (and it is the good syntax, Optional[tuple[]] ?

Yep, that's correct! 

>     3. I don't understand the tests failing errors : it says "error: Name "Tuple" is not defined Did you forget to import it from "typing"?", but in all theses files there is already `from typing import List, Dict, Tuple` ...

I just checked `addto.py` and it does not import `Tuple`. Maybe you missed some files in there?

Twix53791:
OK.
 Are you sure about cache-rm/cache-update-newer ? Cache-add call `papis.document.from_folder`, which take a string as arg, so yes we need to loop through the doc_folder tuple. But cache-rm and cache-update-newer, like cache-update, call `papis.cli.handle_doc_folder_query_all_sort` and then after `for doc in documents:`. It already loop through the documents. 
??

alexfikl:
> OK. Are you sure about cache-rm/cache-update-newer? Cache-add call `papis.document.from_folder`, which take a string as arg, so yes we need to loop through the doc_folder tuple. But cache-rm and cache-update-newer, like cache-update, call `papis.cli.handle_doc_folder_query_all_sort` and then after `for doc in documents:`. It already loop through the documents. ??

Not sure where we got confused, but your changes are absolutely correct!

I can't think of any other places that would need to be updated, so this looks good to go from my end. Did you want to make any other changes or is it ready to merge in? :grin:

Twix53791:
I think its good like that! Thank you for your support :)

PR #637, opened 2023-08-14
sciencedirect: get full abstract

Twix53791:
Probably not the most elegant way, but it works... it overwrites the already parsed 'abstract' value with the full abstract string


alexfikl:
You can see more details if you scroll up in the CI log 
https://github.com/papis/papis/actions/runs/5855745319/job/15874134951?pr=637
and in particular here it seems like that `_` key is missing for the links we use in the tests:
```
        # Get full abstract
        if rawdata:
            rawabstract = rawdata["abstracts"]["content"][0]
            for i, r in enumerate(rawabstract):
                if i == 0:
>                   abstract = rawabstract[r][1]["$$"][0]["_"]
E                   KeyError: '_'
```
Does that help?

alexfikl:
Also, it would be very helpful if you could add another URL to https://github.com/papis/papis/blob/main/tests/downloaders/test_sciencedirect.py that tests a bit the abstract extraction that you have here.

The workflow for that is not documented, but you basically need
* add the URL to the `SCIENCE_DIRECT_URLS` list.
* Run the test with
```
PAPIS_UPDATE_RESOURCES="both" python -m pytest -v tests/downloaders/test_sciencedirect.py
```
That should create some new files in `tests/downloaders/resources` that you also need to commit here. Let me know if you need any help with that!

Twix53791:
Hum, I add:
```diff
SCIENCE_DIRECT_URLS = (
    "https://www.sciencedirect.com/science/article/abs/pii/S0009261497040141",
    "https://www.sciencedirect.com/science/article/abs/pii/S2210271X18305656",
+    "https://www.sciencedirect.com/science/article/pii/S146290112300117X",
    )
```

Then I installed pytest module : `pacman -S python-pytest`
Then I run `PAPIS_UPDATE_RESOURCES="both" python -m pytest -v tests/downloaders/test_sciencedirect.py` in the terminal, in source papis directory
And I got:
```
ERROR: usage: __main__.py [options] [file_or_dir] [file_or_dir] [...]
__main__.py: error: unrecognized arguments: --cov=papis
  inifile: /home/archx/papis/setup.cfg
  rootdir: /home/archx/papis
```

I guess I have a wrong 'dev environment'... By the way, I disagree this this kind of package design using ressources outside the compiled package like here with the tests or more critical, with the plugins management in papis (with stevedore...). It became difficult to hack for 'common people' like me who are not professionnal developers but just motivated "amateurs". It deserves the open source dynamic towards the 'mainstream world' to complicate uselessly things like that. If the tests are just scripts, why not just keeping them as simple independant scripts to run like `python script` ? But aybe github need this complexity to run automatically the tests when pushing on the remote... But for the plugins, it was really possible to just make a dictory were people can just add a downloader script and that's it! Anyway...

I give up for now about this test updating, I prefer think hard about fixing the test failing issues!

Twix53791:
The remaining error indicates than `assert extracted_data == expected_data` in tests/downloaders/test_sciencedirect.py fail. Just because the abstract formatting has changed? It seems to me than it is just about changing the format in the test file and in fact it doesn't matter about the real papis use??

alexfikl:
> The remaining error indicates than `assert extracted_data == expected_data` in tests/downloaders/test_sciencedirect.py fail. Just because the abstract formatting has changed? It seems to me than it is just about changing the format in the test file and in fact it doesn't matter about the real papis use??

Yes, the downloader tests basically have a downloaded version of the webpage they're supposed to parse and an expected result of the parsing (usually generated at the time the test was added). They just verify that the downloader still extracts the same information it used to, so every addition needs a modification to the test.

I updated the current tests for the new abstract parsing and added the additional URL you mentioned. I haven't looked at the abstract parsing code yet, but the tests seem to pass and the output looks good :grin:

Twix53791:
thank you for adding the test!

alexfikl:
I was looking over this and noticed that one of the test cases didn't have an updated abstract, so I tried to fix that and redid the general parsing of the abstract. It's still not complete, but it seems to be doing a pretty good job. For context:

* it can now parse all the abstracts for the URLs in the tests
* it concatenates paragraphs in case the abstract has more like in https://www.sciencedirect.com/science/article/abs/pii/S2210271X18305656

This should be good to go now! Thanks for all the work on figuring this out! :rocket:

Twix53791:
Good job, better now!

PR #638, opened 2023-08-14
rtd: bump image to python 3.11

alexfikl:
The readthedocs builds have been failing recently:
https://readthedocs.org/projects/papis/builds/21604354/
Not quite sure what brought that on. From a bit of googling, the error seems to be due the `sphinx_rtd_theme` not supporting Sphinx v7, but those should be fixed and updated already.

This bumps the image used by RTD to the latest versions to see if that helps at all, but generally not sure what's going on :\

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -4,9 +4,9 @@ sphinx:
   configuration: doc/source/conf.py
 
 build:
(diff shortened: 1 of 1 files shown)


alexfikl:
Well that didn't help..

Issue #556, opened 2023-05-18
By default papis does not download pdfs anymore


alejandrogallo:
Papis only downloads pdfs now if `--force-download` is passed as an option.

Are we sure we want this behaviour to be the default? This feels quite odd.

alexfikl:
Hm, from #387, it seems like the intended behavior was to download PDFs in the first case (no local file) but not the second (with local file)
```
papis add --from arxiv 'some-arxiv-url'
papis add --from arxiv 'some-arxiv-url' some-file.pdf
```
Does that not work as intended anymore or do you think it's a bad default?

alejandrogallo:
exactly, AFAIK this does not work anymore, a quick test

![22-05-23_15:05:55](https://github.com/papis/papis/assets/12715358/800c8677-eee9-43ad-9ac3-c4ee751bd00a)

but your example works, i.e.
![22-05-23_15:09:48](https://github.com/papis/papis/assets/12715358/71dead57-37be-424f-b7dc-5ed85bb4148c)

So I guess the problem is giving an explicit `--from` flag, which I never do ;)

I'll take a look at it.

PR #641, opened 2023-08-15
commands.add: (no)-download-files option

Twix53791:
If there is already another way to do this, I am listening... Or maybe its better to add a setting in papis config?? My goal is just to be sure NEVER a pdf file would be downloaded with --batch option set (so not possibility to confirm, control...)...


alexfikl:
Hm, that `--force-download` flag doesn't seem to be very useful because it tries to be too smart. I think what we want is something like
```
papis add --download-files
papis add --no-download-files
```
like all the `--open/--no-open` options that are already in there. Can you remove the `--force-download` flag and add something like this? Would this also solve your issue?

alexfikl:
@maxmahlke Git seems to blame you for adding the original `--force-download` flag :grin: . What do you think about a scheme like suggested in https://github.com/papis/papis/pull/641#issuecomment-1685078083 to allow both forcing and disabling downloads regardless of any local files passed on the command line?

maxmahlke:
@alexfikl So the idea of my original issue #386 was to make papis a bit smarter by not triggering PDF downloads if the user passes a PDF file. I see from #556 that this had unintended consequences (no more downloads without flag) as we did not check if the user passed a file or a URL as single argument.

However, I think the original motivation (articles behind paywalls, not arxiv) is still valid. This looks sane to me:

```
papis add --from doi 'some-url'   # download by default
papis add --from oid 'some-url' some-file.pdf  # do not download by default
```

@alejandrogallo then suggested to add the `--force-download` flag to be able to still trigger downloads in any case. My implementation of this flag in #387 then broke the use case in #556.

I agree that 

```
papis add https://arxiv.org/yxz.123
```

should definitely trigger a download without having to set a flag. I think the `--force-download` flag could be retained if we check that (1) the `--from` flag is set and (2) the user argument is a local PDF file.

That said, I'm also ok with explicit `--download-files` / `--no-download-files` flags for all commands. I use custom aliases for most papis interaction so I can add these flags once and won't be bothered again. It's the more consistent and explicit approach for the command structure, so I might even favour this.

Twix53791:
Hi, I was around for a couple of days. I think than having to specify always if we download or not files is not the best option. Personnally, I think it is more like a global setting set by users. Personnaly, I don't dowload sometimes files or sometimes not, I just NEVER download and I don't want papis to download files automatically, accumulating files in my disk space without control. I think maybe the best thing is to add a setting in the config : 
`download-files`, could be `never/always/on demand` and then a flag `--force-download` (renamed just `--download-file` ?) to use when the global setting is set to `on demand`.

alexfikl:
> Hi, I was around for a couple of days. I think than having to specify always if we download or not files is not the best option. Personnally, I think it is more like a global setting set by users. Personnaly, I don't dowload sometimes files or sometimes not, I just NEVER download and I don't want papis to download files automatically, accumulating files in my disk space without control. I think maybe the best thing is to add a setting in the config : `download-files`, could be `never/always/on demand` and then a flag `--force-download` (renamed just `--download-file` ?) to use when the global setting is set to `on demand`.
> 
> This solution:
> 
> 1. avoid multiple flags
> 2. is sure (you can disable completly downloads)
> 3. is explicit (you know exactly what you do)
> 4. is simple (adding a config option is really simple)

Take a look at how the `--[no-]confirm` and other similar options work
https://github.com/papis/papis/blob/5465f2fe94ff24082384b106b70d4a4c4b84d72e/papis/commands/add.py#L527-L530
This is mostly what I had in mind, so you wouldn't need to always specify the choice every time if you add it to your config file. I'm not very convinced what the `ondemand` and an additional `--force-download` flag would do here, since it seems very user dependent when they'd want additional files downloaded and from what importer / downloader (maybe a better served by `papis addto`?).

Twix53791:
I didn't understand your goal, actually the way --confirm works is perfect for that, as we can control the option via a global setting. So I tried to implement correctly this option in the new commits. 

I noticed a strange thing (but maybe designed this way?) : if I set in defaults.py the setting `download-files` to `false`, I can switch it to `true` in the papis config with `download-files = True`, but the inverse is not true, if the built-in setting is set to True, the user config set to False not overwrites it. Personnaly I don't understand, it seems like a bug, but maybe there is a reason...

alexfikl:
> I noticed a strange thing (but maybe designed this way?) : if I set in defaults.py the setting `download-files` to `false`, I can switch it to `true` in the papis config with `download-files = True`, but the inverse is not true, if the built-in setting is set to True, the user config set to False not overwrites it. Personnaly I don't understand, it seems like a bug, but maybe there is a reason...

The flag seems to work as expected, i.e. the command line flag can be used to overwrite the configuration value. So if the configuration value is `True`, we get (showing the value of the `download_files` variables that the code would see)
```
> papis add
download_files = True
> papis add --download-files
download_files = True
> papis add --no-download-files
download_files = False
```
so the `--download-files` flag does nothing in this case, since it's used to enable the option which is here the default. If the configuration value is `False`, we get
```
> papis add
download_files = False
> papis add --download-files
download_files = True
> papis add --no-download-files
download_files = False
```
so now the `--no-download-files` flag does nothing, since it's used to disable the option which is now the default. Does that make sense? How would you have expected it to work?

alexfikl:
I removed the ``--force-download`` flag and added some docs, so this now looks good to me! Does it do what you need as well? ``papis add --no-download-files`` should now never try and download files with the importers / downloaders at least.

Twix53791:
Good job. I think we can merge it ! :)

alexfikl:
> Good job. I think we can merge it ! :)

Awesome! Thanks for all the work and discussion on this :rocket:

PR #646, opened 2023-08-28
Modernize using pyupgrade

alexfikl:
This fixes some issue that came up from running
```
ruff check --target-version py37 --select UP papis
```
(essentially running [pyupgrade](https://github.com/asottile/pyupgrade)). The main change is using the more modern f-strings `f"see {value}"` formatting where it makes sense.


alexfikl:
This will likely cause some conflicts, but better to get it in sooner rather than later :(

PR #648, opened 2023-08-29
Add `papis tag` command

jghauser:
This is a draft of a new `papis tag` command that allows adding/removing/clearing tags. There are some docs inside to show you what's possible.

This is mostly just a first working draft to get some feedback. Here are some questions/considerations I'm unsure about:

- Is this a useful thing to ship with papis? We already have `papis update` that can do some of what `papis tag` can do. Is there too much overlap? Should this maybe rather be an additional script?
- What functionality would you like to see? Maybe it could be useful to apply all tags from one doc to another for instance...
- Better idea for the names/structure of the options? For instance I would have liked to use `--add` for adding tags, but the short version of that is `-a` which is already used by `--all`. I've also considered using `+TAG` and `-TAG`, but haven't found a straightforward way of implementing this in python click (and it isn't consistent with other papis commands).
- I'm not all that well versed with python, so any suggestions for improvements are very welcome.
- I've added a new config option for the `tag-format`. I've made some assumptions regarding how people list their tags, namely that they're either a list, or a string separated by `, `, `; ` or ` `. I've considered allowing any arbitrary separators, but that's not very easy to set in the config because of the white space. I've also considered trying to detect the tag-format automatically (which I actually do in `papis.nvim`), but opted against it as it isn't foolproof (which is more acceptable for `papis.nvim` as it doesn't edit any documents).

Note that there might be bugs in here, so don't use with your actual library. Also, it requires explicitly setting the `tag-format` in your config.

EDIT: Bunch of errors in the CI, looks like my winging of the typing system hasn't quite gone right (and I also introduced a bunch of trailing white space when I wrote the explanations). :sweat_smile: I'll get to these!

### DIFF
--- a/papis/commands/tag.py
+++ b/papis/commands/tag.py
@@ -1,0 +1,185 @@
+"""
+This command allows users to add and remove tags to documents in their library.
+
+Examples
+^^^^^^^^
+
+- To add a tag use the `--add` (or `--append`/`-p`) option:
+
+    .. code:: sh
+
+        papis tag --add TAG1 --add TAG2 QUERY
+
+  This will add TAG1 and TAG2 to a document matched by QUERY. You can repeat
+  `--add` as many times as you need. The query is any query supported by papis.
+  If the query matches more than one document, Papis' picker will be started to
+  let you pick the document from those matched (just as with Papis' other
+  commands).
+
+  Tags are only added if they do not already exist, which is to say that the
+  same tag cannot exist more than once in a given document.
+
+- To remove a tag use the `--remove` (or `-r`) option:
+
+    .. code:: sh
+
+        papis tag --remove TAG1 QUERY
+
+  This removes TAG1 from the document. If it didn't exist before, nothing
+  happens. You can use `--remove` as many times as you like to remove multiple
+  tags.
+
+- Use `--drop` (or `-d`) to remove all tags:
+
+    .. code:: sh
+
+        papis tag --drop --add TAG1 QUERY
+
+  This removes all tags from the document and then adds TAG1. You can of course
+  also use `--drop` by itself to simply remove all tags without adding new
+  ones.
+
+- Use `--all` (or `-a`) with any of the other options to apply the tagging
+  operation to all matching documents:
+
+    .. code:: sh
+
+        papis tag --all --add TAG1 QUERY
+
+  This adds TAG1 to all documents matching the QUERY rather than opening the
+  picker to let you choose one.
+
+"""
+from typing import List, Optional, Tuple, Any, Dict
+
+import click
+
+import papis.cli
+import papis.commands
+import papis.importer
+import papis.strings
+
+logger = papis.logging.get_logger(__name__)
+
+
+@click.command("tag")
+@click.help_option("--help", "-h")
+@papis.cli.git_option()
+@papis.cli.bool_flag(
+    "-d",
+    "--drop",
+    help="Drop all tags",
+    default=False,
+)
+@click.option(
+    "-p",
+    "--add",
+    "--append",
+    "to_add",
+    help="Add a tag",
+    multiple=True,
+    type=str,
+)
+@click.option(
+    "-r",
+    "--remove",
+    "to_remove",
+    help="Remove a tag",
+    multiple=True,
+    type=str,
+)
+@click.option(
+    "-n",
+    "--rename",
+    "to_rename",
+    help="Rename a tag (<OLD-TAG NEW-TAG>)",
+    multiple=True,
+    type=(str, str),
+)
+@papis.cli.doc_folder_option()
+@papis.cli.all_option()
+@papis.cli.sort_option()
+@papis.cli.query_argument()
+def cli(
+    git: bool,
+    drop: bool,
+    to_add: List[str],
+    to_remove: List[str],
+    to_rename: List[Tuple[str, str]],
+    sort_field: Optional[str],
+    sort_reverse: bool,
+    query: str,
+    doc_folder: Tuple[str, ...],
+    _all: bool,
+) -> None:
+    """
+    Change a document's tags.
+    """
+
+    # if do_tag:
+    documents = papis.cli.handle_doc_folder_query_all_sort(
+        query, doc_folder, sort_field, sort_reverse, _all
+    )
+
+    if not documents:
+        logger.warning(papis.strings.no_documents_retrieved_message)
+        return
+
+    from papis.commands.update import run_append, run_remove, run_drop, run_rename, run
+
+    key_types: Dict[str, type] = {"tags": list}
+
+    success = True
+    processed_documents: List[Any] = []
+    for document in documents:
+        tags = document.get("tags", [])
+        if not isinstance(tags, list):
+            processed_documents.clear()
+            logger.error(
+                "The document with papis_id '%s' contains tags that aren't "
+                "defined as a list of items. As `papis tag` only supports "
+                "lists, tagging has been aborted and no documents have been "
+                "changed. You can use `papis doctor --check key-type --fix` to "
+                "convert all your tags to use lists.",
+                document["papis_id"],
+            )
+            break
+
+        ctx = papis.importer.Context()
+
+        ctx.data.update(document)
+        if drop and success:
+            run_drop(ctx.data, ["tags"])
+
+        if to_add and success:
+            to_add_tuples = [("tags", tag) for tag in to_add]
+            success = run_append(ctx.data, to_add_tuples, key_types, False)
+
+        if to_remove and success:
+            to_remove_tuples = [("tags", tag) for tag in to_remove]
+            success = run_remove(ctx.data, to_remove_tuples, False)
+
+        if to_rename:
+            to_rename_tuples = [
+                ("tags", old_tag, new_tag) for old_tag, new_tag in to_rename
+            ]
+            success = run_rename(ctx.data, to_rename_tuples, False)
+
+        if success:
+            processed_documents.append((document, ctx.data))
+
+        if not success:
+            processed_documents.clear()
+            logger.error(
+                "Papis has encountered an unexpected error while processing "
+                "document '%s'. Tagging has been aborted and no documents have "
+                "been changed. Please report this bug at "
+                "https://github.com/papis/papis.",
+                document["papis_id"],
+            )
+            break
+
+    for document, data in processed_documents:
+        run(document, data=data, git=git, auto_doctor=False)
+
+    logger.info("Updated tags in %d documents", len(processed_documents))
--- a/setup.py
+++ b/setup.py
@@ -227,6 +227,7 @@ setup(
             "rm=papis.commands.rm:cli",
             "run=papis.commands.run:cli",
             "serve=papis.commands.serve:cli",
+            "tag=papis.commands.tag:cli",
             "update=papis.commands.update:cli",
             "cache=papis.commands.cache:cli",
             "init=papis.commands.init:cli",
--- a/tests/commands/test_tag.py
+++ b/tests/commands/test_tag.py
@@ -1,0 +1,201 @@
+import os
+
+import papis.database
+
+from papis.testing import TemporaryLibrary, PapisRunner
+
+
+def _get_resource_file(filename: str) -> str:
+    resources = os.path.join(
+        os.path.dirname(os.path.abspath(__file__)), "resources", "update"
+    )
+    filepath = os.path.join(resources, filename)
+    assert os.path.exists(filepath)
+
+    return filepath
+
+
+def test_tag_add_general_cli(tmp_library: TemporaryLibrary) -> None:
+    from papis.commands.tag import cli
+
+    cli_runner = PapisRunner()
+
+    result = cli_runner.invoke(cli, ["__no_document__"])
+    assert result.exit_code == 0
+    assert not result.output
+
+    result = cli_runner.invoke(
+        cli,
+        ["--add", "tag3"] + ["--add", "2345"] + ["krishnamurti"],
+    )
+    assert result.exit_code == 0
+
+    db = papis.database.get()
+    (doc,) = db.query_dict({"author": "Krishnamurti"})
+    assert "tag1" in doc["tags"]
+    assert 2345 in doc["tags"]
+
+
+def test_tag_add_to_new_key_cli(tmp_library: TemporaryLibrary) -> None:
+    from papis.commands.tag import cli
+
+    cli_runner = PapisRunner()
+
+    result = cli_runner.invoke(cli, ["__no_document__"])
+    assert result.exit_code == 0
+    assert not result.output
+
+    result = cli_runner.invoke(
+        cli,
+        ["--add", "tag3"] + ["doc without files"],
+    )
+    assert result.exit_code == 0
+
+    db = papis.database.get()
+    (doc,) = db.query_dict({"author": "doc without files"})
+    assert doc["tags"] == ["tag3"]
+
+
+def test_tag_add_del_duplicates_in_list_cli(
+    tmp_library: TemporaryLibrary,
+) -> None:
+    from papis.commands.tag import cli
+
+    cli_runner = PapisRunner()
+
+    result = cli_runner.invoke(cli, ["__no_document__"])
+    assert result.exit_code == 0
+    assert not result.output
+
+    result = cli_runner.invoke(
+        cli,
+        ["--add", "tag3"] + ["--add", "tag3"] + ["krishnamurti"],
+    )
+    assert result.exit_code == 0
+
+    db = papis.database.get()
+    (doc,) = db.query_dict({"author": "Krishnamurti"})
+    assert doc["tags"].count("tag3") == 1
+
+
+def test_tag_remove_general_cli(
+    tmp_library: TemporaryLibrary,
+) -> None:
+    from papis.commands.tag import cli
+
+    cli_runner = PapisRunner()
+
+    result = cli_runner.invoke(cli, ["__no_document__"])
+    assert result.exit_code == 0
+    assert not result.output
+
+    result = cli_runner.invoke(
+        cli,
+        ["--remove", "tag1"] + ["--remove", "1234"] + ["krishnamurti"],
+    )
+    assert result.exit_code == 0
+
+    db = papis.database.get()
+    (doc,) = db.query_dict({"author": "Krishnamurti"})
+    assert doc.get("tags") is None
+
+
+def test_tag_remove_from_missing_key_cli(
+    tmp_library: TemporaryLibrary,
+) -> None:
+    from papis.commands.tag import cli
+
+    cli_runner = PapisRunner()
+
+    result = cli_runner.invoke(cli, ["__no_document__"])
+    assert result.exit_code == 0
+    assert not result.output
+
+    result = cli_runner.invoke(
+        cli,
+        ["--remove", "tag"] + ["doc without files"],
+    )
+    assert result.exit_code == 0
+
+
+def test_tag_drop_general_cli(
+    tmp_library: TemporaryLibrary,
+) -> None:
+    from papis.commands.tag import cli
+
+    cli_runner = PapisRunner()
+
+    result = cli_runner.invoke(cli, ["__no_document__"])
+    assert result.exit_code == 0
+    assert not result.output
+
+    result = cli_runner.invoke(
+        cli,
+        ["--drop"] + ["krishnamurti"],
+    )
+    assert result.exit_code == 0
+
+    db = papis.database.get()
+    (doc,) = db.query_dict({"author": "Krishnamurti"})
+    assert doc.get("tags") is None
+
+
+def test_tag_drop_missing_key_cli(
+    tmp_library: TemporaryLibrary,
+) -> None:
+    from papis.commands.tag import cli
+
+    cli_runner = PapisRunner()
+
+    result = cli_runner.invoke(cli, ["__no_document__"])
+    assert result.exit_code == 0
+    assert not result.output
+
+    result = cli_runner.invoke(
+        cli,
+        ["--drop"] + ["doc without files"],
+    )
+    assert result.exit_code == 0
+
+
+def test_tag_rename_general_cli(
+    tmp_library: TemporaryLibrary,
+) -> None:
+    from papis.commands.tag import cli
+
+    cli_runner = PapisRunner()
+
+    result = cli_runner.invoke(cli, ["__no_document__"])
+    assert result.exit_code == 0
+    assert not result.output
+
+    result = cli_runner.invoke(
+        cli,
+        ["--rename", "tag1", "tag_renamed1"]
+        + ["--rename", "1234", "2345"]
+        + ["krishnamurti"],
+    )
+    assert result.exit_code == 0
+
+    db = papis.database.get()
+    (doc,) = db.query_dict({"author": "Krishnamurti"})
+    assert "tag_renamed1" in doc["tags"]
+    assert 2345 in doc["tags"]
+
+
+def test_tag_rename_missing_value_cli(
+    tmp_library: TemporaryLibrary,
+) -> None:
+    from papis.commands.tag import cli
+
+    cli_runner = PapisRunner()
+
+    result = cli_runner.invoke(cli, ["__no_document__"])
+    assert result.exit_code == 0
+    assert not result.output
+
+    result = cli_runner.invoke(
+        cli,
+        ["--rename", "tag_nonexistent", "tag_renamed1"] + ["krishnamurti"],
+    )
+    assert result.exit_code == 0


alexfikl (commented):
This is definitely great! I'm a big fan of tagging things (as opposed to folders / subfolders) :heart: 

Left a couple of nitpicks for the python stuff and some thoughts below (not all coherent, so feel free to close).
> Is this a useful thing to ship with papis? We already have papis update that can do some of what papis tag can do. Is there too much overlap? Should this maybe rather be an additional script?

Definitely! I tried to use tags many times to separate projects / papers, but the handling is pretty much non-existent, so mostly gave up :( It would be awesome to make it better with a little command like this! :heart: 

> What functionality would you like to see?

It would also be helpful to list all the existing tags and maybe do some deduplication of some sort. For example, I have some tags with `fluid-mechanics` and `fluid-dynamics` (depending on how I felt at the time) and would love to merge them!

That can be in future PRs though.

> Better idea for the names/structure of the options? For instance I would have liked to use --add for adding tags, but the short version of that is -a which is already used by --all. I've also considered using +TAG and -TAG, but haven't found a straightforward way of implementing this in python click (and it isn't consistent with other papis commands).

The `+TAG` / `-TAG` is nice, but yeah, a bit too different from how other commands work :( Not sure I have any better suggestions :\

Maybe having some subcommands like `papis bibtex`? 
```
papis tag QUERY add TAG1 TAG1 remove TAG3 TAG4
```

> I've added a new config option for the tag-format. I've made some assumptions regarding how people list their tags, namely that they're either a list, or a string separated by , , ; or  . I've considered allowing any arbitrary separators, but that's not very easy to set in the config because of the white space. I've also considered trying to detect the tag-format automatically (which I actually do in papis.nvim), but opted against it as it isn't foolproof (which is more acceptable for papis.nvim as it doesn't edit any documents).

I think a more flexible option would be to have a `tag-format-separator` and use that if the `tags` field is a string. Maybe even just standardize on the `tags` being a list in the `info.yaml` file already? Then the separator would be just a fallback thing..

papis/commands/tag.py
```diff
@@ -0,0 +1,148 @@
...
+) -> None:
+    """
+    Add and delete document tags.
+    """
+
+    if not TAG_FORMAT:
+        logger.warning(
+            "The `tag-format` option isn't set. Please define it in your config."
+        )
+        return
```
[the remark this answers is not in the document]
jghauser: The reason why I didn't do this is that we might then end up adding tags in a wrong format. For instance, let's say the default is a list (which I think would be the most sensible). The user, however, is using comma separated strings (but haven't specified that in the config). If they then add a tag to a document that hasn't any tags yet, no error will be generated (because an error requires that `make_set_of_tags()` finds a string). We then add a list item to the doc, leading to inconsistencies. Do you have some idea of how to deal with this?

papis/commands/tag.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+@click.command("tag")
+@click.help_option("--help", "-h")
+@click.option("--new", "-n", multiple=True, help="Add a tag")
+@click.option("--remove", "-r", multiple=True, help="Remove a tag")
+@click.option("--clear", "-c", is_flag=True, help="Clear all tags")
+@papis.cli.doc_folder_option()
+@papis.cli.all_option()
+@papis.cli.query_argument()
```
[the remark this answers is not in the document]
jghauser: Oh, I excellent option, somehow I still keep discovering things in papis all the time! :)

papis/commands/tag.py
```diff
@@ -0,0 +1,148 @@
...
+
+logger = papis.logging.get_logger(__name__)
+
+TAG_FORMAT = papis.config.get("tag-format")
+
+
+def make_set_of_tags(tags: List[str] | str) -> Set[str]:
+    if isinstance(tags, list) and TAG_FORMAT == "list":
+        set_of_tags = set(tags)
+    elif isinstance(tags, str) and TAG_FORMAT == "comma_separated":
```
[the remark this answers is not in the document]
jghauser: This is something I've been thinking of and decided against because the best would then be to allow arbitrary separators. For instance, we might have both '; ' and ';' (so with and without trailing white space). I didn't find any easy way to write trailing white space in the config file (though admittedly I didn't try much), and so opted to just go with descriptive names instead. Do you think it's ok to simply assume (and document) that papis always assumes that the separator is followed by a white space?

papis/commands/tag.py
```diff
@@ -0,0 +1,148 @@
...
+import papis.document
+import papis.strings
+import papis.cli
+
+logger = papis.logging.get_logger(__name__)
+
+TAG_FORMAT = papis.config.get("tag-format")
+
+
+def make_set_of_tags(tags: List[str] | str) -> Set[str]:
```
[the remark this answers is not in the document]
jghauser: I guess I saw this function to be more in the business of getting a common format from whatever the user has defined, but it does make tags unique, too. Happy to change it :)

papis/commands/tag.py
```diff
@@ -0,0 +1,148 @@
...
+
+    if not documents:
+        logger.warning(papis.strings.no_documents_retrieved_message)
+        return
+
+    for doc in documents:
+        doc.load()
+        prior_tags = list()
+        if ("tags" in doc) and (clear is not True):
+            prior_tags = doc["tags"]
```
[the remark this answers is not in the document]
jghauser: good point! I'm using lists myself and I guess it shows haha

jghauser:
Thanks a lot for the feedback, it's super useful!

>  It would also be helpful to list all the existing tags and maybe do some deduplication of some sort. For example, I have some tags with fluid-mechanics and fluid-dynamics (depending on how I felt at the time) and would love to merge them!

Yes, that's something I've been wanting as well. Do you think it might make sense to maybe implement it in `papis doctor` (but use the functionality we'll get from `papis tag`)?

> Maybe having some subcommands like papis bibtex?
        `papis tag QUERY add TAG1 TAG1 remove TAG3 TAG4`

Yes, this seems like a better way to do it. Thanks!

> I think a more flexible option would be to have a tag-format-separator and use that if the tags field is a string. Maybe even just standardize on the tags being a list in the info.yaml file already? Then the separator would be just a fallback thing..

I think it would make sense to standardise things to lists, but then I'm using lists and I'm possibly biased :sweat_smile:. In any case, that would probably be something to do at a later point and with some conversion path. I've added some comments in the responses regarding why I've required explicitly setting the `tag-format` option. If there's a way to assume lists (except if `tag-format-separator` is set) without potentially leading to inconsistencies, I'm happy to do it. Let me know what you think!

papis/commands/tag.py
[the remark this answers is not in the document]
alexfikl: Hm, we can always support some form of quoting in the value of `tag-format-separator` and do
```
separator = papis.config.getstring("tag-format-separator").strip('"').strip("'")
```
to remove any extra quotes around the string? `ConfigParser` itself seems to indeed to `value.strip()` to remove any whitespace :(

papis/commands/tag.py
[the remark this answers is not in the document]
alexfikl: Hm, yeah, you're right, the common format is more important! How about `normalize_tags`?

papis/commands/tag.py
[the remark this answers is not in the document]
alexfikl: It should be fine either way, but this might confuse `mypy` a bit. Let me know if you need any help with that!

alexfikl:
> Yes, that's something I've been wanting as well. Do you think it might make sense to maybe implement it in `papis doctor` (but use the functionality we'll get from `papis tag`)?

Yeah, that's a good idea! The `duplicate-refs` check is already very similar, so it should fit just fine.

> I think it would make sense to standardise things to lists, but then I'm using lists and I'm possibly biased 😅. In any case, that would probably be something to do at a later point and with some conversion path. I've added some comments in the responses regarding why I've required explicitly setting the `tag-format` option. If there's a way to assume lists (except if `tag-format-separator` is set) without potentially leading to inconsistencies, I'm happy to do it. Let me know what you think!

Yeah, I think it also makes sense to standardize on lists. Then we wouldn't have to worry about the tag having whitespace in it, or having the separator as part of the tag for some reason, or any other issue like that. 

Maybe we can make have support for reading the tags in a variety of formats, but when we write them back it's always a list? That way users will slowly migrate to that format in all their documents.

papis/commands/tag.py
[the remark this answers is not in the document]
alexfikl: Yeah, haven't thought of it too well. Like you say, the main issue is that the user may have tags in various formats because we haven't standardized them in any way so it's a wild wild west. Maybe we can add a big warning at the beginning on first use or something?
```
logger.warning(
	"This command assumes that your tags are separated by {separator}. If this is not correct, 
	it will not split them correctly and may give unexpected results. 
	Use `papis doctor -t normalize-tags --fix` to clean up.")
papis.tui.utils.confirm("Are you sure you want to continue?")
```

Will have to think about it some more, so for now feel free to leave it as is!

alexfikl:
> > It would also be helpful to list all the existing tags and maybe do some deduplication of some sort. For example, I have some tags with fluid-mechanics and fluid-dynamics (depending on how I felt at the time) and would love to merge them!
> 
> Yes, that's something I've been wanting as well. Do you think it might make sense to maybe implement it in `papis doctor` (but use the functionality we'll get from `papis tag`)?

Actually this should just work with
```
papis tag --add fluid-dynamics --remove fluid-mechanics QUERY
```
so maybe not a great suggestion :)) Guess this command is awesome already! :heart:

papis/commands/tag.py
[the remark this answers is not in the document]
jghauser: Yeah, a warning like that would be good, too. Ok, for now I'll leave it, we can easily change it later.

papis/commands/tag.py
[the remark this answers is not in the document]
jghauser: Ok, that should work! It's not the prettiest solution, but for sure the most versatile. And then maybe mid term, we can just move everyone to lists

jghauser:
> Actually this should just work with
> 
> ```
> papis tag --add fluid-dynamics --remove fluid-mechanics QUERY
> ```
> 
> so maybe not a great suggestion :)) Guess this command is awesome already! ❤️

Actually, I was thinking to somehow search the library for suspiciously similar tags and then propose to merge them. Not sure that would work with your example, but I've got lots that just differ in capitalisation etc.

jghauser:
> Yeah, I think it also makes sense to standardize on lists. Then we wouldn't have to worry about the tag having whitespace in it, or having the separator as part of the tag for some reason, or any other issue like that.
> 
> Maybe we can have support for reading the tags in a variety of formats, but when we write them back it's always a list? That way users will slowly migrate to that format in all their documents.

Or we could also tell people that `papis tag` only supports lists and give them the option of converting everything to lists. That way `papis tag` can be simple and we don't even need a `tag-format` option. We could have a `papis tag --convert-from "SEPARATOR"` (or so) command that converts everything...

@alejandrogallo, what do you think about standardising tags to lists? I'm asking because my suspicion is that you're using strings (as existing docs use them). Are there any use cases we're not considering for which lists aren't well suited?

Twix53791:
Great new feature than this add/remove tag options. I support it.

@jghauser, what do you mean by "standardising tags to lists" ? Writing a list format `"tag1", "tag2 with space", ...` in the tag field of the yaml file (or formated like the author_list or citations fields...) ? I putting here a python list format but I don't know what it is the best way to store a list in a yaml file... I support the idea as soon as it can be easily parsed directly from the yaml file (with a tool like python json/yq/jq...).

jghauser:
Thanks for the input @Twix53791!


> @jghauser, what do you mean by "standardising tags to lists" ? Writing a list format `"[tag1", "tag2 with space", ...]` in the tag field of the yaml file (or formated like the author_list or citations fields...) ? I putting here a python list format but I don't know what it is the best way to store a list in a yaml file... I support the idea as soon as it can be easily parsed directly from the yaml file (with a tool like python json/yq/jq...).

it would be like this:
```yaml
tags:
- shared intentionality
- cognitive science
- developmental psychology
```

> BUT, what about the databases? I use the whoosh database personally, which let you perform complex searches like `year:2020 AND (tags:neolithique OR tags:australia)`, and I am wondering about the easiness of dealing with list instead of strings in the database. I think than if this change is implemented, it is important to modify the databases accordingly if needed to be sure it doesn't disable any feature the databases previously provided.

I'm not entirely sure how this is handled (maybe @alexfikl can chime in), but it seems to work as expected. I'm also using the whooh database, and when i search for e.g. `tags:"shared intentionality"` it seems to give me all documents containing that tag.

alexfikl:
> I'm not entirely sure how this is handled (maybe @alexfikl can chime in), but it seems to work as expected. I'm also using the whooh database, and when i search for e.g. `tags:"shared intentionality"` it seems to give me all documents containing that tag.

I think it's just stringifying the list, so it does a substring search in the *string* `'["shared intentionality", "cognitive science", "developmental psychology"]'` and it finds stuff as expected.

Twix53791:
Well the format you propose is fine for me (I am telling that because I use `yq` to parse the papis yaml file in another program I am writing, a "fzf browser" for papis [papis-fzf](https://github.com/Twix53791/papis-fzf)). 
But I just tried a bit the whoosh search and I think the database, yes, stringifying the list, but not as you think @alexfikl, more like `["shared", "intentionality", "cognitive", "science", "developmental", "psychology"]`, with a AND operator between each pairs.

I tried with a document:
```yaml
tags:
- cognitive science
- intentionality
- developmental psychology
```
All these searches match it:
`papis list "tags:cognitive AND tags:intentionality"`
`papis list "tags:science AND tags:intentionality"`
`papis list "tags:science AND tags:psychology"

It is not because it matches a part of the string, `papis list "tags:scie` doesn't match.

More problematic. Two entries:
```yaml
tags:
- cognitive science
- intentionality

tags:
- science
- intentionality
- cognitive
```

Are both matched by `papis list "tags:science AND intentionality"`, _worse_ by `tags:cognitive science` and even `tags:science cognitive`...

Maybe not a big deal, but a bit dirty. At least people should be aware than multi-words tags are not safe.

alexfikl:
I'm not 100% sure, but it seems to me like all the queries you listed work as expected. I'll try go give some explanations below. If you're still not sure about it, maybe open a [Discussion](https://github.com/papis/papis/discussions) instead so we don't clutter this PR too much?

> But I just tried a bit the whoosh search and I think the database, yes, stringifying the list like that (I checked directly the whoosh database) : `['cognitive science', 'intentionality']`. But, about the search, it does not consider the list elements as one unique entity.

Yeah, that's correct, in stringifying the list, it should consider the elements all together, i.e. it no longer considers them separate tags, for better or worse.

> I tried with a document:
> 
> ```yaml
> tags:
> - cognitive science
> - intentionality
> - developmental psychology
> ```
> 
> All these searches match it: `papis list "tags:cognitive AND tags:intentionality"` `papis list "tags:science AND tags:intentionality"` `papis list "tags:science AND tags:psychology"`

Those are all expected to me, because the `tags` list contains all of those terms (although in different tags).

> It is not because it matches a part of the string, `papis list "tags:scie` doesn't match.

I said "substring search" before, but that was just a shorthand on my side, I'm sorry. The whoosh backend uses the query language defined by the database here: https://whoosh.readthedocs.io/en/latest/querylang.html. You can also control what gets put in the database with [whoosh-schema-prototype](https://papis.readthedocs.io/en/latest/configuration.html#config-settings-whoosh-schema-prototype).

I'm not very familiar with it, but it seems like you need to say `tags:"scie*"` to get partial matches. The way that works may be a bit weird, but I'm not sure we have much control over it..

> More problematic. Two entries:
> 
> ```yaml
> tags:
> - cognitive science
> - intentionality
> 
> tags:
> - science
> - intentionality
> - cognitive
> ```
> 
> Are both matched by `"tags:science AND intentionality"`, _worse_ by `tags:cognitive science` and even `tags:science cognitive`... I am not sure we can change that as it is probably a feature from whoosh?

This might be because whoosh considers the space a separator and those are actually separate queries. Have you tried if `tags:"cognitive science"` vs `tags:"science cognitive"` gives the desired result?

Twix53791:
Well, yes I think it is the normal whoosh way to handles such things. Personally, I don't like it, as in fact it create a difference with only-word tags. I terms of design, when an algorithm behaves differently according to non explicit parameters like the typo here, I think it is failed.
```yaml
tags:
- cognitive science
- developmental psychology
tags:
- cognitive_science
- developmental_psychology
```
Both tags are the same for a human mind, but whoosh will behave differently in its search....

Anyway, it is also true than the list way doesn't interference with the one-word tags syntax. So it changes absolutly nothing for someone who doesn't use multiple words tags. It is just a positive add of new functionality, whitout interference with previous fonctionality, so I think it is OK for everyone. You tested it with the native papis database also?

jghauser:
I'm finally getting some time to finish this off. I've got two last questions before I implement everything mentioned here:

1. I'm thinking of entirely ditching handling different tag formats. I think standardising to lists makes sense as it will simplify things, avoid more options, and doesn't really have any drawbacks. We could then, when releasing the next version, announce this and offer an update path (which we've already implemented in `papis doctor`, see #656 ). I would still leave a check to see that the type found in the `info.yaml` is indeed a list, and if not warn the user. Do you think we should additionally have a warning saying that the command assumes lists (so people don't inadvertently add list tag items when their library is using strings)? Does papis have some functionality for displaying a warning only when called for the first time?
2. I kind of like the `papis QUERY add TAG1 remove TAG2` synax, but I'm also a bit wary as it feels like going against the syntax used in most current papis commands. In fact, the situation is a bit confusing in papis: `papis cache` and `papis bibtex` use grouped subcommands, but most other things do not. Is there some system behind when to use what or is it mostly about what's needed to most easily implement the functionality?

alejandrogallo:
As always I apologize for the delay in my reply.
There are really great ideas in this feed, these are my comments/thoughts

> 1. I'm thinking of entirely ditching ...

You're right that I use space separated tags. In fact for me tags are just a string that I'll match agains, and I use it
for topics or projects, like `papis open tags:proj-lala`.

# Downsides of tags as a string
- In order to have spaces in the tag, it is not really possible at the moment.
  This might be very upsetting for some people.
- Inherently it is then not clear with separator to use, I have an unconscious
  and innate hatred of commas and I guess that is why papis is more geared to spaces.

# Downsides of tags as a list
- When editing manually documents, users must know that the format of the tags
  are such of a list. But the upside is that with your command
  we can have a more lean and short way of dealing with tags.
X Apparently whoosh and most databases can just store the stringified python list
  in the database, so this is unlikely to bite us in the future.
X Tags are non-greppable. (we can discuss if this is a downside)
  I think it is not valid since yaml is not supposed to be grepped for.

I would say that I support just having a way of storing tags, namely as a list.
It will make the implementation and documentation clearer.
The implementation is not really super complicated, in the sense that
we could affort having this logic here, but I think if we are
going to officially support tags like citations or notes,
then we might as well choose a single format.

- I propose writing a function that checks for the formatting of the tags
  and offers to convert tags to their list format.

> 2. I kind of like the `papis QUERY add TAG1 remove TAG2` synax...

This kind of way by now is kind of typical to papis, for better or for worse.
`papis explore` also has this kind of behaviour.
This kind of behaviour is handy whenever you need a kind of pipe.
In the case of `explore`, you want to search for documents
online and you might pick one, or export them or add one to the
library.

In the case of tags, I don't see this rationale, therefore I would
be rather in favour of `--foo` flags.


# Other commands

I would like to have a `papis tag -al` or
`papis tag --all --list` command. I think it should be straightforward.

In the future we can also implement a cache for tags.

# External API

I would suggest creating a file à la `citations.py`
in `papis/tag.py`
in order to have there two functions to do the tag
managment. This can help me then to use the functions
from the web application. Soemthing lke

papis/tag.py
```py
def get(doc: papis.document.Document):
    tags = papis.get("tags", [])
    if not isinstance(tags, list):
        raise deprecation warning and parse according to space separation
        return re.split(r"\s+", tags)
    return tags

def add(doc: papis.document.Document, tag: str) -> None:
    pass

def remove(doc: papis.document.Document, tag: str) -> None:
    pass

def remove(doc: papis.document.Document, tag: str) -> None:
    pass
```

jghauser:
Thanks a lot for the feedback @alejandrogallo, this is very useful!

> * I propose writing a function that checks for the formatting of the tags
>       and offers to convert tags to their list format.

The idea is to tell users (with that message that is shown on first use) that they can convert their tags to the required format with something like `papis --set doctor-key-type-separator ', ' doctor --check key-type --fix --all`. 

> This kind of way by now is kind of typical to papis, for better or for worse. `papis explore` also has this kind of behaviour. This kind of behaviour is handy whenever you need a kind of pipe. In the case of `explore`, you want to search for documents online and you might pick one, or export them or add one to the library.

> In the case of tags, I don't see this rationale, therefore I would be rather in favour of `--foo` flags.

Thanks for the explanation, that makes a lot of sense.

> # Other commands
> 
> I would like to have a `papis tag -al` or `papis tag --all --list` command. I think it should be straightforward.
> 
> In the future we can also implement a cache for tags.

I think this is already implemented in the draft. There's an `--all` flag that will apply the tagging to all matching documents. Is that what you had in mind?

> #
> 
> I would suggest creating a file à la `citations.py` in `papis/tag.py`

Ok!

jghauser:
Ok, I have had an entirely different idea :sweat_smile:. Instead of adding a `papis tag` command, we could also improve the `papis update` command so it can handle lists (i.e. adding and removing items to/from a list). Then changing tags would just involve that command. It would also mean we could use `papis update` for the proposed `related_to` field, even if it is a list. As far as I can tell, all list items in the `info.yaml` only include unique items, so making a command that that can handle everything doesn't seem impossible.

alexfikl:
> Ok, I have had an entirely different idea 😅. Instead of adding a `papis tag` command, we could also improve the `papis update` command so it can handle lists (i.e. adding and removing items to/from a list). Then changing tags would just involve that command. It would also mean we could use `papis update` for the proposed `related_to` field, even if it is a list. As far as I can tell, all list items in the `info.yaml` only include unique items, so making a command that that can handle everything doesn't seem impossible.

I like that! `papis update` should be able to handle most of this with a few more flags (`--append` at least) and could nicely generalize to other keys. If it turns out that we need some special handling for tags, we can always make a separate `papis tag` command that reuses some of the logic from `papis update`.

alejandrogallo:
hmmm I don't know, so right now this logic can hold.
We currently support kind of 3 things, papis_id, notes and citations.
For only one of these there is an ad-hoc command, `papis citations`.

We can update the `papis update` command nevertheless, IMHO this would be

- `papis update --append <thing> <key>`
- `papis update --list-append/--push <thing> <key>`
- `papis update --remove <key>`
- `papis update --list-remove/--pop <key>`

Only, how do we do for listing all tags? The above approach fails to deliver this.
I think from the point of view of user friendliness, it is much clearer
that we explicitly support tags if we have a tag command.

I am therefore for both, tags command and update of `update` ;)

Twix53791:
I think personnaly also than a tag command would be clearer.

jghauser:
Thanks for all of the input. I'm gonna improve on the `papis update` command (along the lines of what @alejandrogallo suggested) and then implement `papis tag` mostly just re-using the logic of `papis update` :)

jghauser:
@alexfikl, with all the work done in `papis update`, adding a `papis tag` is now quite easy. I hacked this together quickly, but it seems like everything just works as expected :laughing:. I've had one issue with typing, if you could take a look at the `NOTE`s in `tag.py`, that would be very helpful. Also, any other comments you have are of course also appreciated!

papis/commands/tag.py
```diff
@@ -0,0 +1,194 @@
...
+    processed_documents = []
+    for document in documents:
+        ctx = papis.importer.Context()
+
+        ctx.data.update(document)
+        if drop and success:
+            run_drop(ctx.data, ["tags"])
+
+        if to_add and success:
+            # NOTE: without the str(), it gets typed as Literal, which is a problem
```
alexfikl, line 164: I just ran this locally (i.e. remove all the `str`) with `python -m mypy papis` and it doesn't seem to give any errors. 

What configuration are you using? Can you remove the `str` so we can see if this shows up on the CI as well?

jghauser: Oh, my bad! I keep forgetting that the pyright lsp that I have running locally thinks certain things are problematic which actually pass our CI. This is one of them!

papis/commands/tag.py
```diff
@@ -0,0 +1,194 @@
...
+    if not papis.config.getboolean("tag-ignore-warning"):
+        logger.warning(
+            "This command assumes that your tags are a list (rather than a string). "
+            + "If this is not correct, it will not split them correctly and may give"
+            + "unexpected results. Use `papis doctor -t normalize-tags --fix` to "
+            + "clean up. To disable this warning, set `tag-ignore-warning=true` in "
+            + "your configuration."
+        )
+        if not papis.tui.utils.confirm("Are you sure you want to continue?"):
+            return
```
alexfikl, line 134: I would say just make this a hard error and tell users to go for `papis doctor` or manual fixing if their tags are not lists. Something like
```
for document in documents:
	if not isinstance(document["tags"], list):
		logger.error(...)
		break

	...
```

For what's implemented so far, we can probably make this work for string-based tags as well. However, going forward, if we want to do fancier things with tags using lists should be much nicer (e.g. don't have to worry about separators or other special characters).

alexfikl: Oh, that should also be `papis doctor --check key-type --fix`! There's no `normalize-tags` fix right now, but we can introduce one if it looks nicer..

papis/commands/tag.py
alexfikl: Hm, yeah, the CI uses `mypy` not `pyright` for the type checking, so they might give different results in some cases. Although this is a weird one.. it's being very restrictive.

papis/commands/tag.py
jghauser: The only problem with this is if the user already uses strings for their tags but happens to not have defined any yet for the documents that are being updated. In that case we can't know whether they are using lists or not. Actually, your proposal (if I'm not mistaken) wouldn't fail in that case, but would abort when the user does use lists but happens not to have defined any tags yet for the relevant documents...

alexfikl: I was just reading through the comments (it's been a long time!) and I think what @alejandrogallo wrote up in https://github.com/papis/papis/pull/648#issuecomment-1712909809 makes good sense to me too. So, ideally:
* We should only have one way to store tags and lists seem like the better option.
* We should fail early if the tags are not in the right format. Here we could direct the user to `papis doctor` and/or add a `papis tag --fix` shortcut to do the fixing automatically.

What do you think?

alexfikl: >  In that case we can't know whether they are using lists or not. 

In that case we should probably fallback on using lists, since that's the supported format.

jghauser: > We should only have one way to store tags and lists seem like the better option.

I absolutely agree.

> We should fail early if the tags are not in the right format. Here we could direct the user to papis doctor and/or add a papis tag --fix shortcut to do the fixing automatically.

The only worry I have is that users who are using string tags might not know that `papis tag` defaults to lists and then they could make their `info.yaml` files inconsistent (and I think avoiding doing surprising things to people's `info.yaml`'s should be a priority as it might break e.g. their scripts). So, if we do just default to lists, we should be very explicit when releasing the new version that there's a breaking change (of sorts) and that from now on only lists will be supported. Is there maybe a way that a user can be warned once when first using `papis tag` or even just the new version of `papis` (I'm saying 'once' because showing a warning every time definitely gets old fast)?

jghauser: Sounds good, I'll modify it like that!

jghauser:
@alexfikl thanks a lot for all the comments -- they're very helpful! Before i implement these changes and add some tests, what's your view on this (originally an edit to an above message):

> EDIT2: One more question: I'm thinking to remove the `--batch` flag, because all the ways in which `papis` update could fail (and which `--batch` is meant to ignore) can't happen with `papis tag` (I think). Now, do you think I should still check the success values returned by the various functions from `papis update` used in `papis tag`, to catch errors just in case of something unforeseen (and maybe have a message à la "If you encounter this error, please open a bug report"? Or should I just ignore all of these return values?

alexfikl:
> @alexfikl thanks a lot for all the comments -- they're very helpful! Before i implement these changes and add some tests, what's your view on this (originally an edit to an above message):
> 
> > EDIT2: One more question: I'm thinking to remove the `--batch` flag, because all the ways in which `papis` update could fail (and which `--batch` is meant to ignore) can't happen with `papis tag` (I think). 

Hm, it seems like it serves to purpose here, so I'm fine with removing it. 

> Now, do you think I should still check the success values returned by the various functions from `papis update` used in `papis tag`, to catch errors just in case of something unforeseen (and maybe have a message à la "If you encounter this error, please open a bug report"? Or should I just ignore all of these return values?

If it's not too confusing, it might be worth to check the success status and complain if something unexpected happens. The main issue with these things is, like you said, avoiding to mess with the user's library..

jghauser:
Ok perfect, that's how I see it too. Thanks for the feedback :blush:

alexfikl:
>     * In the tests, we always pass `TemporaryLibrary` into the functions, even if it's not actually used (at least that's what I saw in some random other files I looked at). I realised that both in the tests for `papis update` and `papis tag`, not a single test uses this. Should I still leave it in?

Yes! Those are pytest fixtures that set up a temporary configuration and library somewhere in `/tmp/` for use in the test. This actually gets used even if you don't use the `tmp_library` variable because it points to a different config file and a small test library. For example,
* Any time the test calls a function that calls `papis.config.get`, it looks in `/tmp/some-folder/papis/config` instead of `~/.config/papis/config` where the global user config is.
* Any query for documents will return documents in the small test library instead of some user library.

This is only relevant when running the tests locally (not the CI). Plus, I don't think this is relevant for you, since you run the tests in a docker, but I just use a virtual environment and it would pick up my global config.

>     * Messages given when running `papis tag` will sometimes come from `papis update`, which I think is a bit confusing. This reminded me of something @alejandrogallo said when we discussed this command a while ago, namely that he'd like the core functionality of `papis tag` to be accessible with an API in an `tag.py` or so in the base `papis` directory. His reasoning was to then use this from the webapp, but I guess it could also help with the messages. My question is if there's a policy/custom on how e.g. commands may refer to each other: is it ok? should it always be via an 'API' module? Should I create a `update.py` and a `tag.py`? And should we merge this first and then fix both of them (given that we've already merged the changes to `papis update`?

I don't think there's any particular policy (?) for that sort of stuff. There are some commands that call `papis edit` here and there, so it should be fine.

Not sure what @alejandrogallo had in mind, but it would be nice to have a top level `tags.py` file to help with handling tags. We can always refactor the command and move some things around though, so it should be fine to merge this now and do that when we have some time!

One issue that comes to mind with an API for this sort of stuff is that most of the difficulty here comes from the fact that we get a string from the command line and need to know what to do with it when adding it into the document. However, in code this should "just work" because we can pass in a list or an int and not do the `try_parsing_str` and `isinstance(value, key_types[key])` stuff..

alexfikl:
> Messages given when running papis tag will sometimes come from papis update, which I think is a bit confusing.

This is mostly just a small design issue. If instead of returning `success` flags, the functions raised an exception, the caller could handle the message and we could write out something specific for update vs tag.

jghauser:
> Yes! Those are pytest fixtures that set up a temporary configuration and library somewhere in `/tmp/` for use in the test. This actually gets used even if you don't use the `tmp_library` variable because it points to a different config file and a small test library. For example,

Oh, I see! Thanks for the explanation :)

> Not sure what @alejandrogallo had in mind, but it would be nice to have a top level `tags.py` file to help with handling tags. We can always refactor the command and move some things around though, so it should be fine to merge this now and do that when we have some time!
> 
> One issue that comes to mind with an API for this sort of stuff is that most of the difficulty here comes from the fact that we get a string from the command line and need to know what to do with it when adding it into the document. However, in code this should "just work" because we can pass in a list or an int and not do the `try_parsing_str` and `isinstance(value, key_types[key])` stuff..

Ok, I'll make a note of that and sort this out later.

> This is mostly just a small design issue. If instead of returning `success` flags, the functions raised an exception, the caller could handle the message and we could write out something specific for update vs tag.
> 
> I guess for now, it should be sufficient to just remove very explicit `call "papis update ..."` sort of messages from those functions? Most of them were ok from what I recall..

I just went through all the messages and only some of those that `papis tag` can't reach (because they have to do with type mismatches) contain references to `papis update`, so we should be fine. I hadn't thought of using exceptions -- that makes a lot of sense! I'll make a note of that too, and improve this later (it would make sense to do all that in one fell swoop when factoring out things to toplevel `tags.py` and `update.py` files).

jghauser:
@alexfikl, do you think this is ready to merge? :)

papis/commands/tag.py
```diff
@@ -0,0 +1,193 @@
...
+This command allows users to add and remove tags to documents in their library.
+
+Examples
+^^^^^^^^
+
+- To add a tag use the `--new` (or `-n`) option:
+
+    .. code:: sh
+
+        papis tag --new TAG1 --new TAG2 QUERY
```
alexfikl, line 11: This seems to be `papis tag --add` now?

papis/commands/tag.py
```diff
@@ -0,0 +1,193 @@
...
+    "-d",
+    "--drop",
+    help="Drop all tags.",
+    default=False,
+)
+@click.option(
+    "-p",
+    "--add",
+    "to_add",
+    help="Add a tag.",
```
alexfikl, line 83: I'm inclined to go with `--append` here to match `papis update`. What do you think? 

`--add` sounds nice, so I don't feel strongly about this.

papis/commands/tag.py
```diff
@@ -0,0 +1,193 @@
...
+logger = papis.logging.get_logger(__name__)
+
+
+@click.command("tag")
+@click.help_option("--help", "-h")
+@papis.cli.git_option()
+@papis.cli.bool_flag(
+    "-d",
+    "--drop",
+    help="Drop all tags.",
```
alexfikl, line 76:
```suggestion
    help="Drop all tags",
```
Very nitpicky, but these help messages don't usually have a full stop at the end?

papis/commands/tag.py
```diff
@@ -0,0 +1,193 @@
...
+
+    if not documents:
+        logger.warning(papis.strings.no_documents_retrieved_message)
+        return
+
+    run_append = papis.commands.update.run_append
+    run_remove = papis.commands.update.run_remove
+    run_drop = papis.commands.update.run_drop
+    run_rename = papis.commands.update.run_rename
+    run_update = papis.commands.update.run
```
alexfikl, line 136:
```suggestion
	from papis.commands.update import (
		run_append, run_remove, run_drop, run_rename, run)
```
? Looks nicer and more python-y to me.

papis/commands/tag.py
```diff
@@ -0,0 +1,193 @@
...
+
+    key_types = papis.commands.doctor.get_key_type_check_keys()
+    key_types["tags"] = type([])
```
alexfikl, line 139:
```suggestion
    key_types = {"tags": list}
```
? It doesn't seem like we need the others if we just use the tags and overwrite that key anyway.

papis/commands/tag.py
```diff
@@ -0,0 +1,193 @@
...
+
+    success = True
+    processed_documents: List[Any] = []
+    for document in documents:
+        if not isinstance(document.get("tags"), (type(None), list)):
```
alexfikl, line 144:
```suggestion
		tags = document.get("tags", [])
        if not isinstance(tags, list):
```
?

alexfikl (commented):
Left a few comments, but overall this should be almost good to go!

papis/commands/tag.py
jghauser: Yeah, I was hesitant about this, too. `add` sounds natural, and `append` is consistent with the other command. We could also just add `--append` as an alternative?

alexfikl: Yeah, you're right. Let's just keep `--add`!

papis/commands/tag.py
```diff
@@ -0,0 +1,185 @@
...
+
+    .. code:: sh
+
+        papis tag --add TAG1 --add TAG2 QUERY
+
+  This will add TAG1 and TAG2 to a document matched by QUERY. You can repeat
+  `--new` as many times as you need. The query is any query supported by papis.
+  If the query matches more than one document, Papis' picker will be started to
+  let you pick the document from those matched (just as with Papis' other
+  commands).
```
alexfikl, line 17:
```suggestion
- To add a tag use the `--add` (or `-p`) option:

    .. code:: sh

        papis tag --add TAG1 --add TAG2 QUERY

  This will add TAG1 and TAG2 to a document matched by QUERY. You can repeat
  `--add` as many times as you need. The query is any query supported by papis.
  If the query matches more than one document, Papis' picker will be started to
  let you pick the document from those matched (just as with Papis' other
  commands).
```

alexfikl (approved):
There were a few more `--new` left over in there, sorry!

Otherwise this looks good to me! Feel free to merge whenever you feel ready.

papis/commands/tag.py
jghauser: My bad, I could have just searched for those... Thanks!

jghauser:
Thanks for all the help @alexfikl! :heart:

PR #649, opened 2023-08-30
Linting fixes

alexfikl:
@f0rki Cherry-picked from https://github.com/papis/papis/pull/612. This tries to preserve authorship and hopefully doesn't mess it up too much.

It also adds some extra cleanups:
* separate `tools/ci-run-test` and `tools/ci-run-lint` for the tests and the linting things (this was also in #612 in some capacity)
* also make those two separate steps in the CI (they all still run on all platforms and versions, unlike in #612)
* added some small shortcuts to the Makefile: `make flake8`, `make mypy` and `make pytest` to run the various tools in isolation.


.github/workflows/main.yml
```diff
@@ -16,18 +16,25 @@ jobs:
...
       with:
         python-version: ${{ matrix.python-version }}
 
-    - name: Install
+    - name: Install dependencies
       run: make ci-install
+      shell: bash
+
+    - name: Check linting and type annotations
+      continue-on-error: true
```
alexfikl, line 33: Well this apparently doesn't fail the CI if this step fails, which is not great :(

alexfikl: Adding this seems to work (from https://stackoverflow.com/a/73015952)
```
      if: success() || failure()
```

PR #653, opened 2023-09-01
defaults: fix refs doctor check name

alexfikl:
The default had `ref` instead of `refs`. Not sure why everything still worked..


alexfikl:
Oops. Misread that.. those were they key names, not the check names :\ Sorry!

PR #655, opened 2023-09-01
commands.doctor: add fixers for keys-exist

alexfikl:
@f0rki Cherrypicked from #598.

### DIFF
--- a/tests/test_doctor.py
+++ b/tests/test_doctor.py
@@ -51,8 +51,6 @@ def test_keys_check_authors(tmp_config: TemporaryConfiguration,
     monkeypatch.setattr(papis.api, "save_doc", lambda _: None)
 
     papis.config.set("doctor-keys-exist-keys", ["author_list", "author"])
-    papis.config.set("doctor-keys-exist-autofill", ["author_list", "author"])
-
     full_doc = papis.document.from_data(
         {
             "title": "DNA sequencing with chain-terminating inhibitors",
--- a/tests/test_doctor.py
+++ b/tests/test_doctor.py
@@ -40,8 +40,51 @@ def test_keys_check(tmp_config: TemporaryConfiguration) -> None:
         "author": "Sanger, F. and Nicklen, S. and Coulson, A. R.",
         })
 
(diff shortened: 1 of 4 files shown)


alexfikl:
@f0rki I pushed a simplified version of this in [171a607](https://github.com/papis/papis/pull/655/commits/171a60743ebbb186cfcb8354c9a2e38972a53a2a), which is more like what I had in mind in my comments in #598. What do you think? What usecases did you have in mind that would require `register_autofiller_for` and `doctor-keys-exist-autofill`?

tests/test_doctor.py
```diff
@@ -40,8 +40,49 @@ def test_keys_check(tmp_config: TemporaryConfiguration) -> None:
         "author": "Sanger, F. and Nicklen, S. and Coulson, A. R.",
         })
 
+    error1, error2 = keys_exist_check(doc)
+    assert error1.payload == "ref" or error2.payload == "ref"
+    assert error1.payload == "author_list" or error2.payload == "author_list"
+
+
+def test_keys_check_authors(tmp_config: TemporaryConfiguration,
```
alexfikl, line 48: Small note: removed a bunch of tests in here since they seemed to be checking different author separators and formats. That was already checked in #654.

f0rki:
@alexfikl I just noticed I never answered here. I think this is fine. It covers my original usecase.

In my original pR just wanted to add a lot of flexibility to auto-fixing vs. checking, e.g., 
* you want to make sure the `tags` key exists, but there is not auto-fill.
* at the same time you want the `year` key to exist and want to auto-fill it.
or you might want to extend papis with your own auto-filler code, etc.

alexfikl:
> you might want to extend papis with your own auto-filler code, etc.

Yeah, that isn't supported with my simplified version. What I was thinking there is that it should be just as easy to write a custom `my-keys-exist` as it would be to implement an `Autofiller`, so not a big win.. I don't know, we might want to revisit that in the future!

Thank you for taking a look and working on this stuff! :rocket:

PR #656, opened 2023-09-01
commands.doctor: add separator to key-check

alexfikl:
This builds on #652 and adds some extra functionality when dealing with `list -> str` or `str -> list` conversions.

@jghauser What do you think? Could this be used to "fix" tags? i.e. something like this could be used to split them up
```
papis --set doctor-key-type-separator ', ' doctor --check key-type --fix QUERY
```
I added some test cases for that specifically and it seems ok. Did you have other formats in mind?

### DIFF
--- a/doc/source/default-settings.rst
+++ b/doc/source/default-settings.rst
@@ -412,7 +412,20 @@ Bibtex options
 
    A list of two tuples with ``(key, type)`` used by the ``key-type`` check. This
    check will show an error if the key does not have the corresponding type. The
-   type should be a builtin Python type that can be used with ``eval``.
+   type should be a builtin Python type that can be used with ``eval``. For
+   example, this can be ``[("year", "int"), ("tags", "list")]`` to check that the
+   year is an integers and the tags are given as a list in a document.
+
+.. papis-config:: doctor-key-type-check-separator
+
+    A separator used by the ``key-type`` check fixer. When converting from
+    :class:`str` to :class:`list`, it is used to split the string into a list,
+    and when converting from :class:`list` to :class:`str`, it is used to join
+    list items. The split will ignore additional whitespace around the separator
+    (for instance, when set to ``,``, the string ``"extra,    whitespace"`` will
+    be converted to the list ``["extra", "whitespace"]``). To preserve leading
+    or trailing whitespace in the separator, make sure to quote it (for instance,
+    ``", "``).
 
 Citations options
 -----------------
--- a/papis/commands/doctor.py
+++ b/papis/commands/doctor.py
@@ -238,6 +238,7 @@ def refs_check(doc: papis.document.Document) -> List[Error]:
     :returns: an error if the reference does not exist or contains invalid
         characters (as required by BibTeX).
     """
+    import papis.bibtex
     from papis.api import save_doc
 
     folder = doc.get_main_folder() or ""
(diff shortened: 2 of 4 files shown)


papis/defaults.py
```diff
@@ -54,6 +54,7 @@ def get_default_opener() -> str:
     "doctor-duplicated-keys-keys": ["ref"],
     "doctor-html-codes-keys": ["title", "author", "abstract", "journal"],
     "doctor-html-tags-keys": ["title", "author", "abstract", "journal"],
+    "doctor-key-type-check-separator": " ",
```
alexfikl, line 57: Note to self: not sure what a good default would be here? Probably just `None`..

doc/source/default-settings.rst
```diff
@@ -414,6 +414,12 @@ Bibtex options
    check will show an error if the key does not have the corresponding type. The
    type should be a builtin Python type that can be used with ``eval``.
 
+.. papis-config:: doctor-key-type-check-separator
+
+   A separator used by the ``key-type`` check when the ``type`` is :class:`list`
+   to split or join strings. The split will ignore additional whitespace around the
+   separator.
+
```
jghauser, line 422: Maybe this could benefit from a little more explanation, à la: A separator used by the ``key-type`` check. When converting from :class:`str` to :class:`list`, it is used to split the string into list items, and when converting from :class:`list` to :class:`str`, it is used to join list items to form a string. The split will ignore additional whitespace around the separator (for instance, when set to `` , ``, the string ``"extra,    whitespace"`` will be converted to the list ``["extra", "whitespace"]``). To preserve leading and trailing whitespace in the separator, make sure to quote it (for instance, ``", "``).

jghauser: Not sure about the `:class:` syntax, so change that if I used that wrongly...

papis/commands/doctor.py
```diff
@@ -367,18 +367,40 @@ def key_type_check(doc: papis.document.Document) -> List[Error]:
     """
     Check document keys have expected types.
 
-    The ``doctor-key-type-check-keys`` configuration entry defines a mapping
-    of keys and their expected types.
+    The :ref:`config-settings-doctor-key-type-check-keys` configuration entry
+    defines a mapping of keys and their expected types. If the desired type is a
+    list, the :ref:`config-settings-doctor-key-type-check-separator` value can
+    be used to split an existing string.
```
jghauser, line 373: The :ref:`config-settings-doctor-key-type-check-keys` configuration entry defines a mapping of keys and their expected types. If the desired type is a list, the :ref:`config-settings-doctor-key-type-check-separator` value can be used to split an existing string (and, similarly, if the desired type is a string, it can be used to join a list of items).

jghauser (commented):
Looks great! I think this should cover all the cases. I've made some small comments to make the docs a little more explicit, other than that it looks good to go! :)

papis/defaults.py
jghauser: And yes, I think `None` would be good. Better not to assume anything here...

alexfikl: Changed in the latest version. Thanks!

doc/source/default-settings.rst
alexfikl: It was all correct -- just copied this verbatim :grin: Thanks!

papis/commands/doctor.py
alexfikl: Added this too! At the moment the docs for these methods don't get rendered anywhere, but it's good to be clear.

alexfikl:
> Looks great! I think this should cover all the cases. I've made some small comments to make the docs a little more explicit, other than that it looks good to go! :)

Thank you very much for taking a look! I tested it with my tags, which are formatted like `tag-name-1 tag-name-1` (dashed and separated by a space) and it seems to work very nicely.

alexfikl:
@jghauser I'll go ahead and merge this, but feel free to point out bugs / extensions as you're working on `papis tag`!

PR #658, opened 2023-09-02
commands.browse: add browse-key auto option

Twix53791:
A very small improvement: adding a 'browse-key' auto which allows the program to cleverly choose by itself the key if the document["url"] is not found. A step more on the road of the full automation of papis...

### DIFF
--- a/doc/source/default-settings.rst
+++ b/doc/source/default-settings.rst
@@ -334,7 +334,7 @@ Bibtex options
 
 .. papis-config:: browse-key
 
-    This settings provides the key that is used to generate a URL for the
+    This setting provides the key that is used to generate a URL for the
     ``papis browse`` command. In the simplest case, ``browse-key`` is just a
     key in the document (e.g. ``url``) that contains a URL to open. It also
     supports the following special values:
--- a/doc/source/default-settings.rst
+++ b/doc/source/default-settings.rst
@@ -348,11 +348,13 @@ Bibtex options
     to ``doi`` constructs the url from ``dx.doi.org/<DOI>``, providing a
     much more accurate url.
 
-    Default value is set to ``url``. If you need functionality
-    with the ``search-engine`` option, set the option to an empty
-    string e.g.  ::
+    The key can be any existing key into the info.yaml file, but the only ones
+    currently mapped to a specific url construction in browse.py are:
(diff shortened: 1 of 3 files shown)


alexfikl:
This looks good! Can you also add some documentation for the new value? It should go in here under the `browse-key` (feel free to rephrase the whole thing if it's not clear)
https://github.com/papis/papis/blob/main/doc/source/default-settings.rst
and maybe an example at the top of the `papis/commands/browse.py` file if you think it's worth it.

jghauser:
I wonder whether it would make sense to have 'auto' be the default? Seems like it covers most bases. Also, the docs advertise 'doi' as 'providing a much more accurate url'. If that's the case, maybe 'doi' should have priority when set to 'auto'?

Just putting this out there, I'm not really using this feature, and I'm not really familiar with how this would impact things.

alexfikl:
> I wonder whether it would make sense to have 'auto' be the default? Seems like it covers most bases. Also, the docs advertise 'doi' as 'providing a much more accurate url'. If that's the case, maybe 'doi' should have priority when set to 'auto'?

Yeah, that all sounds reasonable to as well! (but I also never used `papis browse`)

Twix53791:
OK guys, I did the changes. Feel free to edit the doc if you think my formulations are not good (I am not a native english speaker so... I am french...).

About selecting 'doi' rather than the 'url' I don't think it is a good idea @jghauser, a doi is a unique identifier leading to the original paper, but a paper can be seen often on various sites. If you link a page from a specific website in papis, who may want to browse this site and not the original doi link. For example, personaly, I like the presentation of ncbi, but if I browse the doi of [this article in ncbi](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8056514/), it redirects me to [the original Elsevier article](https://www.sciencedirect.com/science/article/pii/S1871402121001193?via%3Dihub). I think is good to keep the 'url' as default key, to allow users to browse to the page they saved and be explicit if you want to always get to the doi page.

alexfikl:
Thanks for working on this! I reworked the docs for `browse-key` a bit -- they were a bit repetitive and unclear to me when read as a whole. Hopefully they're a bit clearer now!

Everything else looks great, so in it goes :rocket:

PR #659, opened 2023-09-04
Addto.py add link option

Twix53791:
I was surprised to discover than the --link option in add.py was not available when addto, an oversight, I presume. Usefull to link already existing files in a (non papis) library to papis to a papis reference you forgot to link before, or update a ref with a file you get after, or for any other reason (you want to change the file, etc.).


alexfikl:
This looks perfect! In it goes :rocket:

PR #661, opened 2023-09-06
config: allow template marker in format config

OopsYao:
Allow template prefix marker like `[jinja2]` and `[python]` to set formater locally for fields involving templating. This saves one from setting all the other configuration explicitly when he wants to use jinja2.

### DIFF
--- a/papis/format.py
+++ b/papis/format.py
@@ -1,4 +1,5 @@
 from typing import Any, Dict, Optional
+import re
 
 import papis.config
 import papis.plugin
@@ -8,6 +9,7 @@ import papis.logging
 logger = papis.logging.get_logger(__name__)
 
 FORMATER: Optional["Formater"] = None
+FORMATER_CACHE: Dict[str, Any] = dict()
 
 #: The entry point name for formater plugins.
 FORMATER_EXTENSION_NAME = "papis.format"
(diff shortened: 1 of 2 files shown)


alexfikl:
I can see this failing in a couple of ways:

* The formatting uses plugins (`python` and `jinja2` are just the current ones). What if more are added? Third parties are also free to add plugins here, so those should be recognized as well somehow.
* This seems to look for `[python]` in the format string. What if that string appears for another reason? E.g. someone studying snakes!

I'm not quite sure why you would mix formatters. What's your usecase?

alexfikl:
> I think what might be a better solution is for sensible defaults to be set for all relevant options when you enable the jinja formatter (also, it would be nice to have an alternative option that isn't misspelled 😉). I think I have a pretty complete setup for jinja, happy to contribute them :)

Made this into an issue #662, since it's likely a bit complicated and requires some discussion :( I do agree that it would be cool to have better support for non-default formatters somehow.

alexfikl:
@OopsYao This should be replaced by #711. Can you check if that works for you?

OopsYao:
@alexfikl  Oh thanks! I'm sorry for not keeping my word. Caught up in lots of things (unexpected somehow), really got no time to complete the PR. 😢 I'll test it later today~

alexfikl:
> @alexfikl Oh thanks! I'm sorry for not keeping my word. Caught up in lots of things (unexpected somehow), really got no time to complete the PR. 😢 I'll test it later today~

No worries at all! Whenever you get the chance, some testing would be very appreciated :grin:

PR #663, opened 2023-09-08
Add BibLaTeX doctor checks

alexfikl:
Had this sitting around for a while because I needed to clean some documents with weird / missing fields that BibLaTeX didn't like. It adds 3 new checks to `papis doctor`:

* `biblatex-type-alias`: some BibTeX types are defined mostly for backwards compatibility, so this check looks for them and updates them to modern variants.
* `biblatex-key-alias`: same for BibTeX keys / fields.
* `biblatex-required-keys`: the BibLaTeX manual defines some required keys for each type, e.g. an `article` requires `("author", "title", "journal", "year")` at the least. This checks and warns about missing keys in the document.

None of these are turned on by default, so hopefully they're ok to add for anyone who wants a more BibLaTeX friendly library :grin:
https://mirrors.nxthost.com/ctan/macros/latex/contrib/biblatex/doc/biblatex.pdf

### DIFF
--- a/papis/bibtex.py
+++ b/papis/bibtex.py
@@ -1,6 +1,12 @@
+"""
+A set of utilities for working with BibTeX and BibLaTeX (as described in
+the `manual`_).
+
+.. _manual: https://ctan.org/pkg/biblatex?lang=en
+"""
 import os
 import string
-from typing import Optional, List, FrozenSet, Dict, Any, Iterator
+from typing import Optional, List, Dict, Any, Iterator
 
 import click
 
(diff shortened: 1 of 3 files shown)


alejandrogallo:
This is cool, go ahead.
I'm now thinking about when we have many checkers, it would be cool to have an option like
`papis doctor --all-checks` in order to choose all checks and try your luck.
What do you think ?

alexfikl:
> This is cool, go ahead. 

Thanks for taking a look!

> I'm now thinking about when we have many checkers, it would be cool to have an option like `papis doctor --all-checks` in order to choose all checks and try your luck. What do you think?

That definitely sounds like a good idea! I'll add it in a separate PR :grin:

PR #664, opened 2023-09-09
commands.doctor: add some colors to the output

alexfikl:
Added some colors to the output (that largely match what's in `papis.logging`). It looks like this in my terminal

Screenshot_20230909_110518

@jghauser What do you think? For a lot of errors, the output was hard to parse at a glance, so I thought some colors would do it good!

### DIFF
--- a/papis/commands/doctor.py
+++ b/papis/commands/doctor.py
@@ -644,16 +644,28 @@ def cli(query: str,
         click.echo(json.dumps(list(map(error_to_dict, errors))))
         return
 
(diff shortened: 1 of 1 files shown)


jghauser:
Looks great and rally helps understanding what's going on even at a glance. Thanks! :blush:

alejandrogallo:
It makes sense! Ship it! :D

alexfikl:
Thanks a lot for taking a look! :heart:

PR #666, opened 2023-09-11
format: fix spelling formater vs formatter

alexfikl:
This changes to the correct spelling "formatter" everywhere and adds a bunch of deprecation warnings and safeguards. Will need to test a bit, but it should be safe.

@jghauser Better? :grin:


jghauser:
Great, my ocd is very thankful indeed! haha

PR #669, opened 2023-09-14
commands.add: show document before adding

alexfikl:
After #616, `papis add` would no longer show any information about the document being added. Before it just showed the results of a diff with the importer data, so if there was no importer it also wouldn't show anything. This prints out the document on purpose before asking for confirmation. On my system it looks like:
Screenshot_20230914_101907

@jghauser I tried to make this look like the resulting `info.yaml` file and use similar styling to `tui.widgets.diff` (mainly the black background). What do you think? I mostly wanted it because it can happen that the metadata isn't correct (hit this with the `isbn` importer especially) and it's nice to see beforehand.

### DIFF
--- a/papis/tui/utils.py
+++ b/papis/tui/utils.py
@@ -70,26 +70,36 @@ def confirm(prompt_string: str,
 
 
 def text_area(text: str,
+              title: str = "",
               lexer_name: str = "") -> None:
     """
     Small implementation of a pager for small pieces of text.
 
-    :param text: text
-    :param lexer_name: If the text should be highlighted with
-        some kind of grammar, examples are ``yaml``, ``python`` ...
+    :param text: main text to be displayed.
+    :param title: a title for the text.
+    :param lexer_name: a pygments lexer name (e.g. ``yaml``, ``python``) if the
+        text should be highlighted.
     """
-    from pygments import lex
     from pygments.lexers import find_lexer_class_by_name
 
     pygment_lexer = find_lexer_class_by_name(lexer_name)
-    tokens = lex(text, lexer=pygment_lexer())  # type: ignore[no-untyped-call]
 
-    from prompt_toolkit import print_formatted_text
+    from prompt_toolkit.lexers import PygmentsLexer
+    from prompt_toolkit.shortcuts import print_container
     from prompt_toolkit.styles import Style
-    from prompt_toolkit.formatted_text import PygmentsTokens
+    from prompt_toolkit.widgets import Frame, TextArea
 
     papis_style = Style.from_dict(PAPIS_PYGMENTS_DEFAULT_STYLE)
-    print_formatted_text(PygmentsTokens(tokens), style=papis_style)
+    print_container(
+        Frame(
+            TextArea(
+                text=text,
+                lexer=PygmentsLexer(pygment_lexer),  # type: ignore[arg-type]
+            ),
+            title=title,
+        ),
+        style=papis_style,
+    )
 
 
 def yes_no_dialog(title: str, text: str) -> Any:
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -437,14 +437,15 @@ def run(paths: List[str],
                 "', '".join(papis.config.getlist("unique-document-keys"))
                 )
 
+    has_duplicate = False
     try:
         found_document = papis.utils.locate_document_in_lib(tmp_document)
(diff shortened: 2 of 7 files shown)


papis/tui/utils.py
```diff
@@ -6,6 +6,7 @@
 # Highlighting style used by pygments. This is a copy of the prompt_toolkit
 # default style, but changed to use ansi colors.
 PAPIS_PYGMENTS_DEFAULT_STYLE = {
+    "": "bg:ansiblack",
```
alexfikl, line 9: @jghauser I haven't looked at how the `prompt_toolkit` style stuff works, so I haven't managed to find a nicer way to make the background of the whole `print_formatted_text` in `text_area` black. Is there a better way to add some styling to all the elements?

jghauser:
Sorry, this has passed me by completely! I just tried it out and it looks very nice!

I noticed by chance that when a document is already in the library and thus triggers the duplicate warning, the output can be a bit confusing

```
...
...
...
type: article
url: http://dx.doi.org/10.1090/noti963
volume: '60'
year: 2013

[WARNING] commands.add: Duplication Warning
[WARNING] commands.add: A document (shown above) in the 'main' library seems to match the one to be added.
[WARNING] commands.add: Hint: Use the 'papis update' command instead to update the existing document.
author: Gratzer, G.
author_list:
- affiliation: []
  family: Gratzer
  given: G.
doi: 10.1090/noti963
...
...
...
```
Maybe we could change the following to say that the document shown below is the one about to be added. Maybe something like this:
```
[WARNING] commands.add: A document in the 'main' library (shown above) seems to match the one to be added (shown below).
```
Of course that only works if a document is really always shown below when a duplicate warning is triggered (which I thiiiink should be the case?).

papis/tui/utils.py
jghauser: One worry is that it might cause problems when the default background color of the terminal is set to something light, but I haven't actually checked yet. I'll look into it -- I have to admit I mostly made things work by trial and error :sweat_smile:

alexfikl: Hm, the diff widget should have the same problem. I was mostly looking for a way to have that chunk of text be emphasized a bit, so maybe there are better ways..

alexfikl:
> I noticed by chance that when a document is already in the library and thus triggers the duplicate warning, the output can be a bit confusing
>
> Maybe we could change the following to say that the document shown below is the one about to be added. Maybe something like this:

That's a good point, I'll test that a bit to see how it looks and add some more helpful text! The add command definitely has too many branches.. I always forget something :(

Thanks for taking a look!

alexfikl:
@jghauser You were right that the black background looks weird on a light terminal theme (tried with solarized)! I reverted that and switched to the `TextArea` widget from `prompt_toolkit`. What do you think about the below? The titles should clear up which document is being added and which is a duplicate.

Screenshot_20230925_120120

The frame may be a bit too much.. Keep it? Throw it?

papis/tui/utils.py
alexfikl: Removed this.. definitely looked weird on a white background :(

jghauser:
I like it! I don't have any strong opinion about the frames. On the one hand, I agree that it's maybe a bit much and not too aesthetically pleasing. On the other, it does really help set the document information apart. So, happy either way :smile:

alexfikl:
> I like it! I don't have any strong opinion about the frames. On the one hand, I agree that it's maybe a bit much and not too aesthetically pleasing. On the other, it does really help set the document information apart. So, happy either way 😄

I looked a bit through the other `prompt_toolkit` widgets and nothing seemed like a great fit, so I don't know either :\ I guess we can leave it like this until someone comes up with a better idea..

alexfikl:
Thanks again for your feedback! In it goes :rocket:

jghauser:
Yeah, that makes sense! I've been looking enviously at the sexy [rich](https://github.com/Textualize/rich)/[textual](https://github.com/Textualize/textual) framework, thinking I might try to use that to make Papis even prettier. One thing at a time though... haha

alexfikl:
> Yeah, that makes sense! I've been looking enviously at the sexy [rich](https://github.com/Textualize/rich)/[textual](https://github.com/Textualize/textual) framework, thinking I might try to use that to make Papis even prettier. One thing at a time though... haha

Haha, same! :heart:

PR #670, opened 2023-09-14
commands.doctor: change key-type check to avoid evals

alexfikl:
This changes the configuration setting of `doctor-key-type-check-keys`
```
from: [("year", "int"), ("files", "list")]
to:   ["year:int", "files:list"]
```
with the goal of removing the `eval` usage. 
* Changing from a tuple `("year", "int")` to `"year:int"` allows easier parsing without eval.
* Using `getattr(builtins, cls_name, None)` instead of eval. This should also enforce that the types are builtin Python types.

@alejandrogallo Did you have any fancier types in there that would require an `eval`? What do you think about this?

### DIFF
--- a/doc/source/default-settings.rst
+++ b/doc/source/default-settings.rst
@@ -423,11 +423,11 @@ BibTeX options
 
 .. papis-config:: doctor-key-type-check-keys
 
-   A list of two tuples with ``(key, type)`` used by the ``key-type`` check. This
+   A list of strings ``key:type`` used by the ``key-type`` check. This
    check will show an error if the key does not have the corresponding type. The
-   type should be a builtin Python type that can be used with ``eval``. For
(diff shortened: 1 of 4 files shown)


alejandrogallo:
I think this is good, I thought about lambdas but I don't think this is really relevant, because someone could just register a doctor in the `config.py` file if the person should really want to, so I'm fine with this! Thank yo so much!

alexfikl:
> I think this is good, I thought about lambdas but I don't think this is really relevant, because someone could just register a doctor in the `config.py` file if the person should really want to, so I'm fine with this! Thank yo so much!

Thanks for taking a look! Yeah, lambdas would be nice for fancier conversions, but it should be fairly easy to add them back if someone has an explicit use case.

PR #680, opened 2023-09-23
testing: add new module with testing helpers

alexfikl:
This moves the functionality from `tests/testlib.py` into a nice separate module `papis.testing`, since most of it could also be used by third-party plugins (tried to use it in `papis-zotero` for example). Also,

* Adds docs for all the classes and methods in `papis.testing` + adds them to the `Developer API reference`.
* Adds a `Testing` section to the docs explaining how they work and how they are intended to be used.
* Updates all the tests to use this instead of `tests.testlib`.
* Installs some `pytest` fixtures as entrypoints so that they can be recognized automatically.

@alejandrogallo What do you think about adding a module like this?
@jghauser If you have the time, it would be great if you took a look over the docs added in `docs/source/testing.rst`!
@mar4th3 This adds some docs around testing and how to update resources for downloaders like you mentioned in #575. Let me know what you think!


alejandrogallo:
wow! great job! That makes testing much more clear for future contributors.
Thank you very much for your overall job on testing, it's improved papis quality manyfold! I support this pr and it looks good!

jghauser:
This looks very useful, @alexfikl! I went through the docs and made some minor changes. I pushed the updated file directly to this PR as that seemed easiest. The only thing I was still a bit confused about is on line 155, where I'm not sure what the comment refers to. Is something missing on the next line?

alexfikl:
> This looks very useful, @alexfikl! I went through the docs and made some minor changes. I pushed the updated file directly to this PR as that seemed easiest.

Thank you very much for taking a look! Just pushing the changes was perfect.

> The only thing I was still a bit confused about is on line 155, where I'm not sure what the comment refers to. Is something missing on the next line?

Oops, yeah, forgot to add some dummy file names in there. Should be added now, so this is good to go!

PR #681, opened 2023-09-24
Update the update command

jghauser:
OK, here it is! This has taken a while and turned out to be a bit more complex than I anticipated. Given the many different ways of using it, there are bound to be errors lurking in it still. I'd appreciate if you could try out your various use cases and let me know how it's going. Tests are also still missing...

I've left a bunch of comments throughout (indicated with `TODO` and such) regarding things that I wasn't sure about. Let me know what you think. I'm sure I've also done some things in a silly/hackish way -- I would appreciate any feedback!

I've got another PR coming with the `papis tag` command. It's basically done as it'll be super simple given that it'll be using all the logic from `papis update`. At the moment, I'm importing `papis.commands.update` but I can also move the relevant functions to `papis.update` (as @alejandrogallo wanted for `papis.tag`) if that's how it's done properly).

I've changed the name of the `papis update` options a bit from what @alejandrogallo suggested, because I wanted to avoid some confusion and establish the string and list versions as having equal standing.

### DIFF
--- a/papis/commands/doctor.py
+++ b/papis/commands/doctor.py
@@ -620,15 +620,43 @@ def biblatex_required_keys_check(doc: papis.document.Document) -> List[Error]:
 KEY_TYPE_CHECK_NAME = "key-type"
 
 
-def key_type_check(doc: papis.document.Document) -> List[Error]:
+def get_key_type_check_keys() -> Dict[str, type]:
     """
-    Check document keys have expected types.
+    Check the `doctor-key-type-check-keys` configuration entry for correctness.
 
     The :ref:`config-settings-doctor-key-type-check-keys` configuration entry
     defines a mapping of keys and their expected types. If the desired type is
     a list, the :ref:`config-settings-doctor-key-type-check-separator` setting
     can be used to split an existing string (and, similarly, if the desired type
     is a string, it can be used to join a list of items).
+
+    :returns: A dictionary mapping key names to types.
+    """
+    import builtins
+
+    key_type_check_keys = papis.config.getlist("doctor-key-type-check-keys")
+    processed_key_type_check_keys: Dict[str, type] = {}
+    for value in key_type_check_keys:
+        if ":" not in value:
+            logger.error("Invalid (key, type) pair: '%s'. Must be 'key:type'.",
+                         value)
+            continue
+        key, cls_name = value.split(":")
+        key, cls_name = key.strip(), cls_name.strip()
+
+        cls = getattr(builtins, cls_name, None)
+        if not isinstance(cls, type):
+            logger.error(
+                "Invalid type for key '%s': '%s'. Only builtin types are supported",
+                key, cls_name)
+            continue
+        processed_key_type_check_keys[key] = cls
+    return processed_key_type_check_keys
+
+
+def key_type_check(doc: papis.document.Document) -> List[Error]:
+    """
+    Check document keys have expected types.
 
     :returns: a :class:`list` of errors, one for each key does not have the
         expected type (if it exists).
@@ -681,24 +709,8 @@ def key_type_check(doc: papis.document.Document) -> List[Error]:
         else:
             return fixer_convert_any
 
-    import builtins
-
     results = []
-    for value in papis.config.getlist("doctor-key-type-check-keys"):
-        if ":" not in value:
-            logger.error("Invalid (key, type) pair: '%s'. Must be 'key:type'.",
-                         value)
-            continue
-
-        key, cls_name = value.split(":")
-        key, cls_name = key.strip(), cls_name.strip()
-
-        cls = getattr(builtins, cls_name, None)
-        if not isinstance(cls, type):
-            logger.error(
-                "Invalid type for key '%s': '%s'. Only builtin types are supported",
-                key, cls_name)
-            continue
+    for key, cls in get_key_type_check_keys().items():
 
         doc_value = doc.get(key)
         if doc_value is not None and not isinstance(doc_value, cls):
--- a/papis/commands/update.py
+++ b/papis/commands/update.py
@@ -1,36 +1,124 @@
 """
-This command is to update document metadata.
+This command allows you to update the document metadata stored in the
+`info.yaml` file. With it, you can either change individual values manually
+or update a document with information automatically retrieved from a variety
+of sources.
+
+When using it to add information, Papis formatting strings and Python
+expressions can be used. See below examples for more information. The command
+also tries to sanitise filenames so that they don't contain any problematic
+characters.
+
+Normally, `papis update` will abort on encountering an error. If you want to
+skip errors and apply as many changes as possible, use the `--batch` flag.
 
 Examples
 ^^^^^^^^
 
-- Update a document automatically and interactively
-  (searching by DOI in Crossref or in other sources...)
+- Search among papers with the tag "classics" and update the author to
+  "Einstein, Albert":
 
     .. code:: sh
 
-        papis update --auto 'author : dyson'
+        papis update --set author "Einstein, Albert" "tags:classics"
 
-- Update your library from a BibTeX file, where many entries are listed.
-  We will try to look for documents in your library that match these
-  entries and will ask you entry per entry to update it. For example,
-  ``libraryfile.bib`` is a file containing many entries, then
+  This will open the picker containing all documents that match the query from
+  where you can select the document you want to update.
 
-    .. code::
+- Set the author_list to the surname "Einstein" and name "Albert":
+
+    .. code:: sh
+
+        papis update --set author_list "[{'family': 'Einstein', 'given': 'Albert'}]"
+
+  As you can see, `papis update` tries to parse the input string as a python
+  expression (such as a list or dictionary). When this succeeds, as in the above
+  example where we're setting a dictionary, it's the parsed expression that is
+  used to update the metadata.
+
+- Update the journal to "Mass and Energy" for all documents with the journal
+  "Energy and Mass":
+
+    .. code:: sh
+
+        papis update --all --set journal "Mass and Energy" "journal:'Energy and Mass'"
+
+
+  The `--all` flag means that the tag is applied to all documents that match the
+  query, rather than allowing you to pick one individual document to update.
+
+- Update a document automatically and interactively (searching by DOI in
+  Crossref or in other sources...)
+
+    .. code:: sh
+
+        papis update --auto "author:dyson"
+
+- Update your library from a BibTeX file, where many entries may be listed:
+
+    .. code:: sh
 
         papis update --from bibtex libraryfile.bib
 
-- Tag all "einstein" papers with the tag "classics"
+  Papis will try to look for documents in your library that match these
+  entries and will ask you for each entry whether you want to update it.
 
-    .. code::
+- Add the ", Albert" to the author string of a documents matching 'Einstein':
 
-        papis update --all --set tags classics einstein
+    .. code:: sh
 
-  and add the tag of "physics" to all papers tagged as "classics"
+        papis update --set author "{doc[author]}, Albert" Einstein
 
-    .. code::
+  The `papis update` command tries to format input strings using the configured
+  formatter. Here, it is used to get the existing author "Albert" and then add
+  the string ", Einstein" to end up with "Einstein, Albert"
 
-        papis update --all --set tags '{doc[tags]} physics' einstein
+- The above can also be achieved with the `--append` option:
+
+    .. code:: sh
+
+        papis update --append author ", Albert" Einstein
+
+    This appends ", Einstein" to the existing author string.
+
+- You can also append an item to a list:
+
+    .. code:: sh
+
+        papis update --append tags physics
+
+    This adds the tag 'physics' to the existing list of tags. If the list
+    doesn't yet exist, it will be created. All duplicate items will be removed
+    from the list.
+
+- To remove an item from a list, use `--remove`:
+
+    .. code:: sh
+
+        papis update --remove tags physics
+
+    If the tag "physics" is in the list of tags, this command removes it.
+
+- To remove a key-value pair entirely, use `--drop`:
+
+    .. code:: sh
+
+        papis update --drop tags
+
+    This removes the all tags.
+
+- There is also a convenience option `--rename` if you want to rename
+  a list item. It's equivalent to doing `--remove` and `--append` sequentially.
+
+    .. code:: sh
+
+        papis update --rename tags physics philosophy
+
+  This renames the tag 'physics' to 'philosophy'. Note that this option being a
+  combination of `--remove` and `--append`, it will append the desired values
+  even if the value to be removed didn't exist. Thus, the above command will add
+  the tag "philosophy" even if the tag "physics" didn't exist before the
+  operation.
 
 Command-line Interface
 ^^^^^^^^^^^^^^^^^^^^^^
@@ -39,27 +127,226 @@ Command-line Interface
     :prog: papis update
 """
 
-from typing import List, Dict, Tuple, Optional, Any
+import ast
+from typing import Any, Dict, List, Optional, Tuple
 
 import click
 
-import papis.utils
-import papis.strings
+import papis.cli
+import papis.commands.doctor
+import papis.config
 import papis.document
 import papis.format
-import papis.cli
-import papis.importer
 import papis.git
+import papis.importer
 import papis.logging
-import papis.commands.doctor
+import papis.strings
+import papis.utils
 
 logger = papis.logging.get_logger(__name__)
 
+KEY_TYPES = papis.commands.doctor.get_key_type_check_keys()
 
-def run(document: papis.document.Document,
-        data: Optional[Dict[str, Any]] = None,
-        git: bool = False,
-        auto_doctor: bool = False) -> None:
+
+def try_parsing_str(key: str, value: str) -> str:
+    """
+    Tries to parse the input string as a python expression.
+
+    :returns: The parsed input string if string is a python expression,
+        otherwise an unchanged string.
+    """
+    try:
+        value = ast.literal_eval(value)
+    except (SyntaxError, ValueError):
+        logger.debug("Value '%s' of key '%s' is not a python expression.", value, key)
+    return value
+
+
+def run_set(
+    document: papis.document.DocumentLike,
+    to_set: List[Tuple[str, str]],
+    key_types: Dict[str, type],
+) -> None:
+    """
+    Processes a list of `to_set` tuples and applies the resulting changes to the
+    input document. Each tuple is (KEY, VALUE) and results in setting the KEY to
+    the VALUE in the document.
+    """
+    for key, value in to_set:
+        value = papis.format.format(value, document, default=value)
+        value = try_parsing_str(key, value)
+        if isinstance(value, int) and key_types.get(key) is str:
+            value = str(value)
+        if key == "notes" and isinstance(value, str):
+            # TODO: handle renames/deletions of files on disk
+            document[key] = papis.utils.clean_document_name(value)
+            logger.warning(
+                "Document note renamed in the info.yaml file. This does not "
+                "rename any files on disk."
+            )
+        elif key == "files" and isinstance(value, list):
+            # TODO: handle renames/deletions of files on disk
+            document[key] = list()
+            for file in value:
+                if isinstance(file, str):
+                    document[key].append(papis.utils.clean_document_name(file))
+                else:
+                    document[key].append(value)
+            logger.warning(
+                "Document files renamed in the info.yaml file. This does not "
+                "rename any files on disk."
+            )
+        else:
+            document[key] = value
+
+
+def run_append(
+    document: papis.document.DocumentLike,
+    to_append: List[Tuple[str, str]],
+    key_types: Dict[str, type],
+    batch: bool,
+) -> bool:
+    """
+    Processes a list of `to_append` tuples and applies the resulting changes
+    to the input document. Each tuple is (KEY, VALUE) and results in appending
+    the VALUE to the KEY item.
+
+    :returns: A boolean indicating whether the update was successful.
+    """
+    success = True
+    processed_lists = set()
+    supported_keys = key_types.keys() | document
+    for key, value in to_append:
+        if key in supported_keys:
+            value = papis.format.format(value, document, default=value)
+            type_doc = type(document.get(key))
+            type_conf = key_types.get(key)
+            if type_doc is str or (type_doc is type(None) and type_conf is str):
+                document[key] = document.setdefault(key, "") + value
+            elif type_doc is list or (type_doc is type(None) and type_conf is list):
+                value = try_parsing_str(key, value)
+                if key == "files":
+                    value = papis.utils.clean_document_name(str(value))
+                document.setdefault(key, []).append(value)
+                processed_lists.add(key)
+            else:
+                logger.error(
+                    "Items of key '%s' have the type '%s', for which Papis "
+                    "doesn't support the append operation.",
+                    key,
+                    type(document[key]).__name__
+                    if document.get(key)
+                    else key_types[key].__name__,
+                )
+                if not batch:
+                    success = False
+                    break
+        else:
+            logger.error(
+                "We cannot append to key '%s', because we do not know the "
+                "intended type. Please use `papis update --set` instead.",
+                key,
+            )
+            if not batch:
+                success = False
+                break
+
+    for key in processed_lists:
+        document[key] = list(set(document[key]))
+
+    return success
+
+
+def run_remove(
+    document: papis.document.DocumentLike, to_remove: List[Tuple[str, str]], batch: bool
+) -> bool:
+    """
+    Processes a list of `to_remove` tuples and applies the resulting changes
+    to the input document. Each tuple is (KEY, VALUE) and results in removing
+    the VALUE from the KEY item.
+
+    :returns: A boolean indicating whether the update was successful.
+    """
+    success = True
+    for key, value in to_remove:
+        if key in document:
+            if isinstance(document.get(key), list):
+                try:
+                    document[key].remove(value)
+                except ValueError:
+                    try:
+                        document[key].remove(int(value))
+                    except ValueError:
+                        pass  # do nothing if there is nothing to remove
+            else:
+                logger.error(
+                    "You're trying to remove an item from '%s', which has the "
+                    "type '%s'. `papis update --remove` only supports lists.",
+                    key,
+                    type(document.get(key)).__name__,
+                )
+                if not batch:
+                    success = False
+                    break
+        else:
+            logger.info(
+                "Document doesn't have key '%s', cannot remove '%s' from it. "
+                "Continuing...",
+                key,
+                value,
+            )
+
+    return success
+
+
+def run_drop(document: papis.document.DocumentLike, to_remove: List[str]) -> None:
+    """
+    Processes a list of `to_drop` strings and applies the resulting changes
+    to the input document. Each string is a KEY whose value is set to None
+    (and then later in `run()` dropped from the document entirely).
+
+    """
+    for key in to_remove:
+        if key in document:
+            document[key] = None
+        else:
+            logger.info(
+                "Document doesn't have key '%s', cannot remove it. Continuing...", key
+            )
+
+
+def run_rename(
+    document: papis.document.DocumentLike,
+    to_rename: List[Tuple[str, str, str]],
+    batch: bool,
+) -> bool:
+    """
+    Processes a list of `to_rename` tuples and applies the resulting changes
+    to the input document. Each tuple is (KEY, VALUE_OLD, VALUE_NEW) and results in
+    rename the KEY's VALUE_OLD to VALUE_NEW.
+
+
+    :returns: A boolean indicating whether the update was successful.
+    """
+    to_remove = [x[:2] for x in to_rename]
+    to_append = [x[::2] for x in to_rename]
+    success = run_remove(document, to_remove, batch)
+    if success:
+        success = run_append(document, to_append, KEY_TYPES, batch)
+    return success
+
+
+def run(
+    document: papis.document.Document,
+    data: Optional[Dict[str, Any]] = None,
+    git: bool = False,
+    auto_doctor: bool = False,
+) -> None:
+    """
+    Updates the document in the Papis library.
+
+    :returns: None
+    """
     if data is None:
         data = {}
 
@@ -68,22 +355,32 @@ def run(document: papis.document.Document,
 
     if not folder or not info:
         from papis.exceptions import DocumentFolderNotFound
+
         raise DocumentFolderNotFound(papis.document.describe(document))
 
     document.update(data)
+
+    # delete all keys that do not have a value
+    to_drop = [k for k, v in document.items() if not v]
+    [document.pop(k) for k in to_drop]
+
     if auto_doctor:
-        logger.info("Running doctor auto-fixers on document: '%s'.",
-                    papis.document.describe(document))
+        logger.info(
+            "Running doctor auto-fixers on document: '%s'.",
+            papis.document.describe(document),
+        )
         papis.commands.doctor.fix_errors(document)
 
     from papis.api import save_doc
+
     save_doc(document)
 
     if git:
         papis.git.add_and_commit_resource(
-            folder, info,
-            "Update information for '{}'".format(
-                papis.document.describe(document)))
+            folder,
+            info,
+            "Update information for '{}'".format(papis.document.describe(document)),
+        )
 
 
 @click.command("update")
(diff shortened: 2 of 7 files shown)


papis/commands/update.py
```diff
@@ -39,25 +107,237 @@
...
+    processed_tuples: Dict[str, Any] = {}
+    success = True
+    for key, value in to_set:
+        processed_value: Any = try_parsing_str(key, value)
+        print(value)
+        # TODO: I'm aborting if an error is found rather than processing whatever is
+        #       correct. I noticed this is different from some other places in papis
+        #       (e.g. `doctor`). But I think it's better to as obvious as possible when
+        #       stuff is wrong. Let me know if I should change it (same applies to other
+        #       functions).
```
alexfikl, line 207: Yeah, I think most commands modify what they can and issue warnings for any errors. I'm not sure what's better here..

papis/commands/update.py
```diff
@@ -39,25 +107,237 @@
...
+        if key == "files":
+            files = value
+            processed_value = list()
+            for file in files:
+                processed_value.append(papis.utils.clean_document_name(file))
+
+        processed_tuples[key] = processed_value
+
+    document.update(processed_tuples)
+    return document, success
```
alexfikl, line 225: `Document` (and `dict`) are mutable collections, so you don't actually need to return it here because the input `document` will get updated.

Actually for all of these functions, maybe it would be better to not do the `document.update` part and just return a `Dict[str, Any]` with the parsed keys?

papis/commands/update.py
```diff
@@ -66,17 +346,21 @@ def run(document: papis.document.Document,
 
     if not folder or not info:
         from papis.exceptions import DocumentFolderNotFound
+
         raise DocumentFolderNotFound(papis.document.describe(document))
 
     from papis.api import save_doc
+
+    document.clear()
```
alexfikl, line 354: ? This feels odd.. why do you need to clear all the document keys?

alexfikl (commented):
I briefly looked over this and was a bit confused by all the fancy processing.

I'm also not a big fan of special flags the list and string parts, so maybe we can talk about it a bit more. I'll just blurt out some things to start :grin:

To start with the new flags, I'm hoping we can make a few with special semantics based on type. What do you think about the below?
* `--set <key> <value>`: Would just overwrite whatever is in the key with the given value, regardless of type.
* `--drop <key>`: would just delete the given key from the document, regardless of type.
* `--append <key> <value>`: would append the value to the existing value.
	- `list`: just append it to the list.
	- `string`: would need to concatenate it somehow at the end. Currently this seems to let the user add delimiters? Not sure how smart we can be here?
	- other types: skip them? Can't really append to an `int` or something.
* `--remove <key> <value>`: would remove the value from the existing value.
	- `list`: `list.remove` should do the trick.
	- `string`: `str.replace(substring)`? The only thought here would be that `tag-1 tag-2 tag-3.replace("tag-2", "")` would leave some ugly looking spaces.
	- other types: just skip as well.

For the string versions of `--append` and `--remove` we might want to add a more general version of `doctor-key-type-check-separator` that helps with these. The idea being that they're strings, but we're essentially treating them as lists here. What use case did you have in mind for `--append-to-string`?

Some other random questions:
* What to do when `--append <key> <value>` and key does not exist? Error or create a new list with `[<value>]`?
* What to do when `--remove <key> <value>` and the entry is now empty? Leave it or remove `<key>` from the document?
* Why does `papis update` check that they keys are "admissible" and do a bunch of processing on that? From what I can tell, the end goal there was to be able to go `--set key '["value", "value", "value"]'` and have it process it as a list?

jghauser:
> For the new flags, I'm hoping we can make a few with special semantics based on type. What do you think about the below?
> 
>     * `--set <key> <value>`: Would just overwrite whatever is in the key with the given value, regardless of type.
> 
>     * `--drop <key>`: would just delete the given key from the document, regardless of type.
> 
>     * `--append <key> <value>`: would append the value to the existing value.
>       
>       * `list`: just append it to the list.
>       * `string`: would need to concatenate it somehow at the end. Currently this seems to let the user add delimiters? Not sure how smart we can be here?
>       * other types: skip them? Can't really append to an `int` or something.
> 
>     * `--remove <key> <value>`: would remove the value from the existing value.
>       
>       * `list`: `list.remove` should do the trick.
>       * `string`: `str.replace(substring)`? The only thought here would be that `tag-1 tag-2 tag-3.replace("tag-2", "")` would leave some ugly looking spaces.
>       * other types: just skip as well.
> 

I like this. The main reason I didn't do it is the problem of appending a value when the key doesn't yet exist. But now I think it's better to handle that in a different way (see below).

> For the string versions of `--append` and `--remove` we might want to add a more general version of `doctor-key-type-check-separator` that helps with these. The idea being that they're strings, but we're essentially treating them as lists here. What use case did you have in mind for `--append-to-string`?

I'm not sure the complexity of handling `--remove` with strings is worth it. If the string is really a list, then I think we should encourage the user to just store it as a list. It's very error prone otherwise, for instance a using might know they have the tag 'science' they would like to remove, but they forgot that there's also 'cognitive science', leading to deletions that aren't desired. So, for tags, we should really move to them being a list. When doing `--append` with a string, I would say just add the new string to the end of the pre-existing string (as is done now).

> Some other random questions:
> 
>     * What to do when `--append <key> <value>` and key does not exist? Error or create a new list with `[<value>]`?

This is one of the main reasons I left a list and string version of the options. I think it wouldn't be very user-friendly to error when the key is missing (or has no value), especially if you're trying to, say, append a tag to a whole bunch of documents, only some of which have pre-existing tags. The problem of course is that without prior values, we can't always be sure what the type should be. However, with a fairly complete `doctor-key-type-check-separator` config, this isn't much of a problem. What do you think about the following when appending:

1. If key:value pair exists, append based on type
2. If key:value pair doesn't exist and type is defined in `doctor-key-type-check-separator`, create key and add value.
3. Else, give error

>     * What to do when `--remove <key> <value>` and the entry is now empty? Leave it or remove `<key>` from the document?

I would say remove it. This is what I have implemented at the moment. This is mostly for aesthetic reasons and because I don't see any benefit in keeping the key. Are you thinking of something?

>     * Why does `papis update` check that they keys are "admissible" and do a bunch of processing on that? From what I can tell, the end goal there was to be able to go `--set key '["value", "value", "value"]'` and have it process it as a list?

Yeah, I think this isn't currently done nicely haha. So basically, the way it works now is that we first try to parse the input string as a python expression, resulting in a value that can have any type. And then `is_value_admissible()` checks whether this is the right type for a given key. Now, it would be smarter to first check what type is expected for a key and try to parse the input string if we want a list (and error when that isn't possible). At the moment `is_value_admissible()` is also used to check, when list-appending and string-appending, whether the key is a list or a string, respectively. So it catches cases where someone mistakenly tries to append a string to a list or a list item to a string. Again, probably not the least confusing way of doing it :sweat_smile:

papis/commands/update.py
```diff
@@ -39,25 +107,237 @@
...
+) -> str:
+    """
+    Tries to format the input string with the currently configured formatter.
+
+    :returns: A string derived from the input string by formatting or an
+        unchanged string.
+    """
+    try:
+        value = papis.format.format(
+            value, document
```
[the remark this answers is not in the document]
jghauser: nice, thanks!

papis/commands/update.py
jghauser: Yeah, I noticed that too. I think it would be better (in general) not to modify anything if there's an error. A reason why this is particularly problematic for the `papis update` command is that re-running the command after fixing the error can lead to undesired results. So rather than just fixing the command you ran, you need to fix it and then remove all the things you are sure ran correctly...

papis/commands/update.py
jghauser: The only problem here is with deleting keys, which can't be nicely represented in a `Dict[str, Any]`. Although I guess we could set them to empty values or None and then remove the keys with such values?

papis/commands/update.py
jghauser: I think I added this because of deletions. If we only do `update()` then keys that have been deleted will not have any effect in the resulting document.

alexfikl:
> I'm not sure the complexity of handling `--remove` with strings is worth it. If the string is really a list, then I think we should encourage the user to just store it as a list. It's very error prone otherwise, for instance a user might know they have the tag 'science' they would like to remove, but they forgot that there's also 'cognitive science', leading to deletions that aren't desired.

Yeah, that's a very good point. I'm definitely fine with not handling `--remove` for anything but lists. This isn't really a big deal from my point of view because you can always `sed` string changes anyway

> This is one of the main reasons I left a list and string version of the options. I think it wouldn't be very user-friendly to error when the key is missing (or has no value), especially if you're trying to, say, append a tag to a whole bunch of documents, only some of which have pre-existing tags. The problem of course is that without prior values, we can't always be sure what the type should be. However, with a fairly complete `doctor-key-type-check-separator` config, this isn't much of a problem. What do you think about the following when appending:
> 1. If key:value pair exists, append based on type
> 2. If key:value pair doesn't exist and type is defined in `doctor-key-type-check-separator`, create key and add value.
> 3. Else, give error

Hm, that's also a good point, we do need some way to tell `--append` what a default value would be. I'll think about it some more, but for the moment this sounds good to me!

Just to offer an alternative that came to mind, we can also do something local like
```
--append <key> <value>[:type?list]
# or
--append <key>[type?list] <value>
# or
--append <key> <value> [type?list]
```
not sure if `click` works with optional values, but something like that. So by default `--append` assumes that it's working with a list, but if you really need to append to an non-existing value that you need to be a string, you can always do `--append key value:str`.

> Yeah, I think this isn't currently done nicely haha. So basically, the way it works now is that we first try to parse the input string as a python expression, resulting in a value that can have any type. And then `is_value_admissible()` checks whether this is the right type for a given key. Now, it would be smarter to first check what type is expected for a key and try to parse the input string if we want a list (and error when that isn't possible). At the moment `is_value_admissible()` is also used to check, when list-appending and string-appending, whether the key is a list or a string, respectively. So it catches cases where someone mistakenly tries to append a string to a list or a list item to a string. Again, probably not the least confusing way of doing it 😅

I'm not convinced we should be checking that so thoroughly. If the current key is a list, then append to it as a list, and if it's a string, do the string thing. What's the intended behavior here? E.g. if I try to append to a string (here it assumes that `tags` are stored as a string)
```
papis update --append-list tags 'tag-x'
```
it would skip it and issue a warning?

alexfikl:
Is it worth adding a `--replace key from_value to_value` as well? Mostly as a shortcut for `--remove key from_value --append key to_value`, but without changing the order, I guess?

jghauser:
> > 
> 
> Yeah, that's a very good point. I'm definitely fine with not handling `--remove` for anything but lists. This isn't really a big deal from my point of view because you can always `sed` string changes anyway

Yeah, that's how I see it too. I think for now, I'll just not allow it, and if there's demand for it (@alejandrogallo? :smile:), it should be easy to add later on.

> EDIT: Maybe we can make it `--remove key <regex>` and let people shoot themselves in the foot? :D

Haha, that's certainly a very versatile option, though not sure how much it would be used in practice. Having it under `--remove` would also be a bit of a misnomer.

> 
> Just to offer an alternative that came to mind, we can also do something local like
> 
> ```
> --append <key> <value>[:type?list]
> # or
> --append <key>[type?list] <value>
> # or
> --append <key> <value> [type?list]
> ```
> 
> not sure if `click` works with optional values, but something like that. So by default `--append` assumes that it's working with a list, but if you really need to append to an non-existing value that you need to be a string, you can always do `--append key value:str`.

On a cursory glance, I couldn't find anything that seems to implement this in `click`, but it wouldn't be too hard to check if an input string ends with `:str` or `:list`. I think if we do this, it should function as an override, so that the default behaviour depends on what's defined in `doctor-key-type-check-keys` (thus allowing nice and automatic stuff. As above, I think for now I'll change things to work as I suggested, and if we want this as well, it will be easy to add :).

> I'm not convinced we should be checking that so thoroughly. If the current key is a list, then append to it as a list, and if it's a string, do the string thing. What's the intended behavior here? E.g. if I try to append to a string (here it assumes that `tags` are stored as a string)
> 
> ```
> papis update --append-list tags 'tag-x'
> ```
> 
> it would skip it and issue a warning?

Yeah, that's what would have happened (but mostly because of possibly unjustified assumption that what we're going to find in `tags` should correspond to what is defined in `doctor-key-type-check-keys` and thus we'd try to list-append something to a string, which can't go well). 

I think this has become sort of moot if we (as I think we should) unify things to a single `--append` command. Then we just check what's currently there, and do the right thing and if nothing's there, consult `doctor-key-type-check-keys`. Checking whether an existing value corresponds to what is defined in `doctor-key-type-check-keys` is something which (as I'm guessing you're implying) belongs more properly to `papis doctor`.

jghauser:
> Is it worth adding a `--replace key from_value to_value` as well? Mostly as a shortcut for `--remove key from_value --append key to_value`, but without changing the order, I guess?

Yes, that would be a nice convenience option! I think the easiest would be to implement it as a combo of remove and append, which maybe won't preserve item order (but I don't think papis really does that anyway?). Not sure how the `list(set(["some", "list"]))` that I use to get rid of duplicates deals with order either... Do you think preserving list item order is something people expect?

jghauser:
hey @alexfikl, here's a new version of the updated `papis update` command. I've left a bunch of `NOTE`s throughout with various things I'm unsure about. Happy about any feedback and questions!

papis/commands/update.py
```diff
@@ -39,25 +120,223 @@
...
+
+    for key in processed_lists:
+        try:
+            document[key] = list(set(document[key]))
+        except Exception:
+            logger.warn(
+                "You're appending a `list` or `dict` item, which are not "
+                "hashable. Papis therefore couldn't check whether you've "
+                "introduced a duplicate in the list."
+            )
```
alexfikl, line 255: This seems like a slightly related "fix" to add here.. maybe it should be a `papis doctor` check instead?

papis/commands/update.py
```diff
@@ -39,25 +120,223 @@
...
+                    document[key].remove(value)
+            else:
+                logger.error(
+                    "You're trying to remove an item from '%s', which has the "
+                    "type '%s'. `papis update --remove` only supports lists.",
+                    key,
+                    type(document.get(key)).__name__,
+                )
+                success = False
+                break
```
alexfikl, line 284: Do we want to break on the first error here?

papis/commands/update.py
```diff
@@ -39,25 +120,223 @@
...
+    try:
+        value = ast.literal_eval(value)
+    except (SyntaxError, ValueError):
+        logger.debug("Value '%s' of key '%s' is not a python expression.", value, key)
+    return value
+
+
+def run_set(
+    document: papis.document.DocumentLike, to_set: List[Tuple[str, str]]
+) -> None:
```
alexfikl, line 160: All the new functions seem to use the `black` formatting like here. It's probably fine to leave it as is, but we should see if we want to use a formatter to be more consistent?

papis/commands/update.py
alexfikl: That's a good point. I'm not sure what the semantics of this function are supposed to be? The reason I thought it was weird is because doing
```
papis.commands.update.run(doc, {"year": 2023})
```
will just completely clear the document, i.e. it does `doc = data` instead of any form of update. Maybe we don't want to call this in the `cli`?

papis/commands/update.py
```diff
@@ -86,100 +369,163 @@ def run(document: papis.document.Document,
...
+    type=(click.Choice(papis.importer.available_importers()), str),
+    nargs=2,
+    multiple=True,
+    default=(),
+)
+@click.option(
+    "-s",
+    "--set",
+    "to_set",
+    help="Set a key's value. ",
```
alexfikl, line 388:
```suggestion
    help="Overwrite a document's key with the given value",
```
?

papis/commands/update.py
```diff
@@ -86,100 +369,163 @@ def run(document: papis.document.Document,
...
+    multiple=True,
+    type=(str, str),
+)
+@click.option(
+    "-d",
+    "--drop",
+    "to_drop",
+    help="Drop a key-value pair.",
```
alexfikl, line 396:
```suggestion
    help="Drop a key from the document.",
```

papis/commands/update.py
```diff
@@ -86,100 +369,163 @@ def run(document: papis.document.Document,
...
+    multiple=True,
+    type=str,
+)
+@click.option(
+    "-p",
+    "--append",
+    "to_append",
+    help="Append an item to a string or a list. ",
```
alexfikl, line 404:
```suggestion
    help="Append a value to document key",
```

papis/commands/update.py
```diff
@@ -86,100 +369,163 @@ def run(document: papis.document.Document,
...
+    multiple=True,
+    type=(str, str),
+)
+@click.option(
+    "-r",
+    "--remove",
+    "to_remove",
+    help="Remove an item from a list.",
```
alexfikl, line 412:
```suggestion
    help="Remove a value from the document key",
```

papis/commands/update.py
```diff
@@ -86,100 +369,163 @@ def run(document: papis.document.Document,
...
+    multiple=True,
+    type=(str, str),
+)
+@click.option(
+    "-n",
+    "--rename",
+    "to_rename",
+    help="Rename a list item.",
```
alexfikl, line 420:
```suggestion
    help="Replace a value in the document key.",
```
? Maybe we should rename this to `--replace`? Sounds a bit more general in case we ever want to do other things here..

papis/commands/update.py
```diff
@@ -86,100 +369,163 @@ def run(document: papis.document.Document,
...
+    to_set: List[Tuple[str, str]],
+    to_drop: List[str],
+    to_append: List[Tuple[str, str]],
+    to_remove: List[Tuple[str, str]],
+    to_rename: List[Tuple[str, str, str]],
+) -> None:
+    """
+    Update a document from a given library. It accepts both Papis formatting
+    strings and Python expressions. Filenames are sanitised. See documentation
+    for more information.
```
alexfikl, line 444: This seems rather long and not very consistent with the other commands?

papis/commands/update.py
```diff
@@ -86,100 +369,163 @@ def run(document: papis.document.Document,
...
+                    "automatically removed. Check importers: '%s'",
+                    "', '".join(importer.name for importer in matching_importers),
+                )
+
+                del imported.data["ref"]
+
+            ctx.data.update(imported.data)
+
+            # Delete keys that don't have any values
+            {k: v for k, v in ctx.data.items() if v}
```
alexfikl, line 519:
```suggestion
            ctx.data = {k: v for k, v in ctx.data.items() if v}
```
?

papis/commands/update.py
```diff
@@ -86,100 +369,163 @@ def run(document: papis.document.Document,
...
+            if "ref" in imported.data:
+                logger.debug(
+                    "An importer set the 'ref' key. This is not allowed and will be "
+                    "automatically removed. Check importers: '%s'",
+                    "', '".join(importer.name for importer in matching_importers),
+                )
+
+                del imported.data["ref"]
+
+            ctx.data.update(imported.data)
```
alexfikl, line 516: Probably for future PRs, but we should really be smarter about how we merge the data here. E.g. if we do `papis update --drop year` and the importers add the year back, is that a good idea?

alexfikl: Maybe worth doing a `update_doc_from_data_interactively` if nothing else?

papis/commands/update.py
```diff
@@ -86,100 +369,163 @@ def run(document: papis.document.Document,
...
+            {k: v for k, v in ctx.data.items() if v}
+
+            processed_documents.append((document, ctx.data))
+
+        if not success:
+            processed_documents.clear()
+            logger.error(
+                "Aborting operation. No documents have been changed.",
+            )
+            break
```
alexfikl, line 528: I'm not sure I understand this success flag? Why is it aborting the whole thing if one document failed?

alexfikl (commented):
Finally got around to looking at this again.

Left a bunch of nitpicks and comments.. but overall I think this is good to go, just need to clean it up a bit! :grin:

papis/commands/update.py
alexfikl: Just noticed that we already discussed this a bit in https://github.com/papis/papis/pull/681#discussion_r1336995439 and the same logic would apply here. Fair enough, I don't think I have any better suggestions here, but I may complain when I hit a use case where it would be nice to just steamroll through any errors :grin:

alexfikl: Oh, maybe steamroll through errors if `--batch` is set? That sounds like a "just do it!" sort of flag..

papis/commands/update.py
```diff
@@ -39,25 +120,223 @@
...
+            document.pop(key, None)
+            break
+        if key == "notes" and isinstance(value, str):
+            document[key] = papis.utils.clean_document_name(value)
+        elif key == "files" and isinstance(value, list):
+            for file in value:
+                if isinstance(file, str):
+                    document[key].append(papis.utils.clean_document_name(file))
+                else:
+                    document[key].append(value)
```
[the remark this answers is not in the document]
jghauser: I thought there might be the case where you've renamed a file and want to adjust the info.yaml to reflect this. Although admittedly, the `clean_document_name` part would then probably not be all that useful (I still think it might be good to keep just be consistent re file names).

papis/commands/update.py
```diff
@@ -39,25 +120,223 @@
...
+
+def run_set(
+    document: papis.document.DocumentLike, to_set: List[Tuple[str, str]]
+) -> None:
+    """
+    Processes a list of `to_set` tuples and applies the resulting changes to the
+    input document. Each tuple is (KEY, VALUE) and results in setting the KEY to
+    the VALUE in the document.
+
+    :returns: A boolean indicating whether the update was successful.
```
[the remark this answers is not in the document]
jghauser: Oops, thanks!

papis/commands/update.py
jghauser: Good point! I didn't really touch this as I don't use the importers with update. Should I add a note in the code for now to work on this later?

papis/commands/update.py
jghauser: Yes that would seem to be better haha

papis/commands/update.py
jghauser: Can you explain me your usecase? I was thinking that if there's some error then clearly you did something wrong. I guess maybe your library is inconsistent and you just want to get something done rather than fix it? 

I can live with the suggestions re steamrolling through errors when `--batch` is set though :laughing:

jghauser:
Thanks a ton for the detailed feedback :heart:. Super useful for someone like me who's not all that experienced in programming!

papis/commands/update.py
```diff
@@ -39,25 +120,223 @@
...
+                if key == "files" and isinstance(value, str):
+                    value = papis.utils.clean_document_name(value)
+                document[key] = document.setdefault(key, [])
+                document[key].append(value)
+            else:
+                logger.error(
+                    "Items of key '%s' have the type '%s', for which Papis "
+                    "doesn't support the append operation.",
+                    key,
+                    type(document.get(key)).__name__ or key_types.get(key),
```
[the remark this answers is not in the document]
alexfikl: Hm, in case there is no key in the doc, you'll get
```
>>> type({}.get("missing")).__name__
'NoneType'
```
which is still true because it's a non-empty string. Does this do what you want
```
type(  document.get(key, key_types.get(key))  ).__name__
```
?

papis/commands/update.py
```diff
@@ -39,25 +120,223 @@
...
+        #       cases. should we continue doing so? i've done so in the most
+        #       'stupid' way, allowing users to mess things up if they set
+        #       these fields to something other than strings. the reason is
+        #       mainly that otherwise it would almost make more sense to just
+        #       always check if the type is right. but doing that required code
+        #       that is really just there to allow the user to should themselves
+        #       in the foot. not sure what's best here.
+        if not value:
+            document.pop(key, None)
+            break
```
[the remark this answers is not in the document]
alexfikl: My main worry was that there's some valid looking input that would make this be empty, e.g. a typo
```
papis update --set date "{doc[zear]}" <QUERY>
```
would pop the date from the document because the formatting with fail and the `try_parse_str` will fail. I'm fine with leaving it as is though, since there's probably other things like this that can make it fail in interesting ways..

papis/commands/update.py
alexfikl: Yeah, I meant checking `processed_lists` and issuing this warning.

papis/commands/update.py
alexfikl: I'm mostly using ruff in my projects too these days, so definitely a fan! (Haven't tried the formatter yet)

papis/commands/update.py
alexfikl: Hm, that's a good point, we can leave it as is.

papis/commands/update.py
alexfikl: Ah, no no, I just meant the doc comment here. Most other commands have a one-line like
https://github.com/papis/papis/blob/4562963adedbf42ef70e88ebcc9b4fa8805c26f5/papis/commands/add.py#L501
because this will show up in `papis update --help`.

papis/commands/update.py
alexfikl: Yeah, adding a little note sounds sounds good!

papis/commands/update.py
alexfikl: What I had in mind was some more general clean up of some sort, e.g.
```
papis update --append tags --all <QUERY>
```
and one of the documents has `tags` as a string instead of as a list, so the whole thing would fail. But you're right, this is probably best solved by going in an fixing that document.

papis/commands/update.py
alexfikl: Yeah, you're right, that's too much trouble!

papis/commands/update.py
alexfikl: In the `cli` function, it's calling `run` like
```
    for document, data in processed_documents:
        run(document, data=data, git=git)
```
and we'll need another function or just write out the whole thing here? e.g. replace by
```
    for document, data in processed_documents:
		document.clear()
		document.update(data)
		if git:
			...
```

papis/commands/update.py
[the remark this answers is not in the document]
alexfikl: Yeah, you'll definitely want to rename the file on disk as well here!

jghauser:
So this took me a long time to get back to (made worse by the fact that I had obviously forgotten half of what I knew by the time that I did). BUT, I think it's done (except tests).

First, somehow my dev container broke with the last `git pull`. When I try to run the tests I get `/usr/local/bin/python: No module named pytest` and when I try to run things interactively, it says the `papis` command isn't found. I did manage earlier to test various things and the CI (is now) happy too, so I'm reasonably confident that it all works :sweat_smile:)

Here's a quick summary of what changed/how I addressed your (@alexfikl) comments:

- I decided not to handle the renaming of files on disk. Doing so in a smart/correct way isn't super simple, and I wanted to avoid feature creep. So, `papis update` now exclusively works on the document metadata. This might be something for a future PR though. You can still update filenames and notes (and various checks make sure they come out correct), but the command will only change the metadata. I left this for consistency and because it might be useful to bring metadata in line with files that have been renamed by hand.
- I added various bits in `run_set()` to handle edge cases where a user wants to set e.g. the title to '1234'. Before, this would have caused an `int` to be saved, now it's a `str`.
- Difficulty on line 274-279 that I'm not sure how to handle best. Because we can't distinguish between wanting to remove an `int` or a `str`, the code now first tries removing a `str`, and if that fails, an `int`. We might also just try both? Not sure what's best... (maybe a warning if we encounter an `int` and `str` that are the same?)
- `papis update` will still fail (and not change any data) if it encounters problems. However, there is now a `--batch` flag which allows just steamrolling through problems and updating as much as possible.
- I added a note for the `--rename` command, which clarifies that it is a combination of `--remove` and `--append`, and that this means that it will add a value even if the old value doesn't exist. I did think about first checking whether that value exists, but doing so complicates things for unclear benefits. Maybe something to be changed in the future if there is a desire.
- All the other things @alexfikl mentioned (thanks again!), I've implemented :smile:.

Let me know what you think! I haven't started writing tests, but if you have any tips (beyond what's in the docs), let me know.

papis/defaults.py
```diff
@@ -104,7 +104,21 @@ def get_default_opener() -> str:
...
+                                   "related_to:list",
+                                   "author:str",
+                                   "journal:str",
+                                   "note:str",
+                                   "number:int",
+                                   "volume:int",
+                                   "type:str",
+                                   "publisher:str",
+                                   "title:str",
+                                   "shorttitle:str"],
```
alexfikl, line 121: I don't know if all of these are fair. Maybe we can leave more of them up to the user? From a quick look:
* `tags` are mostly strings, I think? Maybe we can leave this to `papis tag`?
* `related_to`: is this a custom key?
* `number`: I sometimes use this to mean `report number` for some techreports from research institutes. I'm not sure if BibTeX says this is an int?
* `volume`: I have some volumes that use roman numerals?

alexfikl (commented):
This is long overdue! We should totally get it in! Some tests would be great, but otherwise this is great :heart: 

One question though: How does this interact with #598? I recall you had some auto-fixing in here too, but with that in maybe we can simplify some things in here and make sure things are the correct type and whatnot.

papis/defaults.py
jghauser: That is fair, I can pare down this list a bit. I think the `tags` (as well as `related_to`) are leftovers from other changes meant to happen, but we can discuss this in future PRs. And good points with the other two, I'll remove them!

jghauser:
> One question though: How does this interact with https://github.com/papis/papis/pull/598? I recall you had some auto-fixing in here too, but with that in maybe we can simplify some things in here and make sure things are the correct type and whatnot.

At the moment, the `papis update` command will try to update things in a sensible way. So, it will not allow you to update things in a way that clashes with the types in the document or defined in `defaults.py` (though, unlike `auto-doctor` the types used in the info.yaml have precedence). I'll have a look whether taking this out in some smartish fashion would make things simpler.

Off the top of my head, I not sure though whether relying on `auto-doctor` for this will help a lot: 
- Users who do not have `auto-doctor` enabled may still expect papis to not introduce type inconsistencies (`auto-doctor` being more something to wrangle external data into the right format). 
- A bunch of checks are there to ensure that operations happen in the right sort of way (e.g. append to list is different from append to string). There probably isn't much saved by using `auto-doctor`.
- We'd need to get the type information from `papis doctor` ahead of time (that is, we can't just check the document once `papis update` is done). Otherwise `papis update` cannot, for instance, append a string if this key isn't yet defined (the current behaviour when you append something to a non-existent key is to check the `defaults.py` for the type and append to an empty string/list as is suitable).

jghauser:
> I checked and you seem to have removed the `papis update --auto-doctor` stuff in this PR. Is that intentional?

Are you sure there ever was such stuff? :sweat_smile: I don't think there ever was any explicit way of turning `auto-doctor` on/off in this PR. Or am I misunderstanding something? I'm definitely happy to add a flag that just runs `auto-doctor` (and it does make sense given that we might import data as well, I always forget about that!)

alexfikl:
> Are you sure there ever was such stuff? 😅 I don't think there ever was any explicit way of turning `auto-doctor` on/off in this PR. Or am I misunderstanding something? I'm definitely happy to add a flag that just runs `auto-doctor` (and it does make sense given that we might import data as well, I always forget about that!)

Ah, sorry, I meant the version that's already merged in `main` here:
https://github.com/papis/papis/blob/e85fb2474c7310b5046a502686e417e0a9769aac/papis/commands/update.py#L74-L77
and the associated flag
https://github.com/papis/papis/blob/e85fb2474c7310b5046a502686e417e0a9769aac/papis/commands/update.py#L98-L100

These seem to have been deleted in some merge / rebase?

jghauser:
Oh, you're totally right! I'll add this back in. Should I just add it where it was? That location seems sensible so that `auto-doctor` runs once just before saving (when enabled).

alexfikl:
> Oh, you're totally right! I'll add this back in. Should I just add it where it was? That location seems sensible so that `auto-doctor` runs once just before saving (when enabled).

Yeah, that sounds good! Thanks :D

jghauser:
@alexfikl, do you have any idea why the container doesn't seem to work (or is it only broken for me)? It used to work some time ago, but when I did a `git pull` recently, it seems to have broken for me.

I get:
```
podman run -v $(pwd):/papis --rm -it papisdev
/usr/local/bin/python: No module named pytest
```

Having this working would help with writing the tests!

alexfikl:
> @alexfikl, do you have any idea why the container doesn't seem to work (or is it only broken for me)? It used to work some time ago, but when I did a `git pull` recently, it seems to have broken for me.
> 
> I get:
> 
> ```
> podman run -v $(pwd):/papis --rm -it papisdev
> /usr/local/bin/python: No module named pytest
> ```
> 
> Having this working would help with writing the tests!

Hm, I don't really use it, so haven't noticed anything wrong. I'll look into it a bit and get back to you!

papis/commands/doctor.py
```diff
@@ -620,16 +620,45 @@ def biblatex_required_keys_check(doc: papis.document.Document) -> List[Error]:
...
     can be used to split an existing string (and, similarly, if the desired type
     is a string, it can be used to join a list of items).
 
+    :returns: A dictionary mapping key names to types.
+    """
+    import builtins
+
+    key_type_check_keys = papis.config.getlist("doctor-key-type-check-keys")
+    processed_key_type_check_keys: Dict[str, type] = {}
+    # TODO: somehow i'm still getting a traceback here even when catching the exception
```
alexfikl, line 639: Is this still relevant? Just ran a quick `papis doctor --checks key-type -a` and didn't get any tracebacks.

jghauser: Great catch, i forgot to check this file again!

tests/commands/test_update.py
```diff
@@ -32,103 +31,447 @@ def test_update_run(tmp_library: TemporaryLibrary) -> None:
...
-
-    result = cli_runner.invoke(
-        cli,
-        ["krishnamurti", "--set", "ref", "NewRef"])
-    assert result.exit_code == 0
-    assert not result.output
-
-    db = papis.database.get()
-    doc, = db.query_dict({"author": "krishnamurti"})
-    assert doc["ref"] == "NewRef"
```
alexfikl, line 134: Why was this removed? Or am I missing something?

tests/commands/test_update.py
```diff
@@ -32,103 +31,447 @@ def test_update_run(tmp_library: TemporaryLibrary) -> None:
...
         assert result.exit_code == 0
 
     import papis.bibtex
-    data, = papis.bibtex.bibtex_to_dict(filename)
+
+    (data,) = papis.bibtex.bibtex_to_dict(filename)
 
     db = papis.database.get()
-    doc, = db.query_dict({"author": "Wannier"})
+    (doc,) = db.query_dict({"author": "Wannier"})
```
alexfikl, line 475: Did black reformat this entire file? Haha. Not sure how those things are configured, but maybe there's a  knob to turn it off if there's no `[tool.black]` in `pyproject.toml` or something like that?

Having said that, we should probably try out a formatter, either `black` or `ruff format` at this point, so that things are vaguely consistent :disappointed:

alexfikl (approved):
Just noticed a little test gone, but otherwise this looks good to me!

I tested it briefly the other day and it doesn't seem to explode, so I'm very happy with just merging this now and fixing any issues as they show up (since it's a pretty big change) :grin: 

So feel free to merge it whenever you feel it's ready!

tests/commands/test_update.py
jghauser: I removed this as it seemed redundant. Setting a ref is just like setting any string and that's already being tested elsewhere. If you'd like it back, happy to do so :smiley:

tests/commands/test_update.py
jghauser: Yeah, it's `ruff` doing things. I'm also not quite sure how to configure stuff to get this working right. On the one hand, I do like to have a formatter as it makes sure I don't do silly things that mess with the linter. On the other hand, well these odd changes happen. Usually I turn it off when editing a line here or there in an existing doc, but keep it on when it's a new file (or I change most things as I did here). Not sure what's the best thing to do :thinking:

alexfikl: > In this particular case, should I change at least these brackets back to what it was before?

No, it should be fine, since there's a couple of slightly different styles in the code anyway. We'll just need to add a formatter and do a cleanup.. maybe after 0.14 is released?

tests/commands/test_update.py
alexfikl: Hm, yeah, I guess it is redundant now! I added this test in #539 because before `papis update` wouldn't work on refs at all.

jghauser: Ah makes sense! I've made a small change and now explicitly set a `ref` in `test_update_set_general_cli()`. It's probably not a bad thing to have that in there given how important `ref`s are

tests/commands/test_update.py
jghauser: Sounds good!

jghauser:
Alright, I'm going to merge this as soon as CI passes. :tada: Thanks again @alexfikl, your feedback has been invaluable. This has taken a long time, but I think the end result is worth it :grinning:. Next step, `papis tag`!

PR #683, opened 2023-09-29
fix getting add-download-files

jghauser:
Had some strange behaviour with my settings not taking effect, and I think the reason is that this should be getting a boolean option.


alexfikl:
Hm, yeah, you're right that that ade no sense :\\
Can you fix the other uses of this? `add-confirm` / `add-open` / etc also use the same broken logic.. (used `grep 'True if papis\.config'`)

jghauser:
Ok, done!

alexfikl:
Realized that we can simplify those things a bit! :rocket:

PR #685, opened 2023-10-01
papis.doctor: add logging to all fixers

alexfikl:


alexfikl:
The `whoosh` tests keep failing on windows when it tries to set `papis.logging.setup("DEBUG")`. I really don't know what that could be about..

https://github.com/papis/papis/actions/runs/6371056175/job/17292678411?pr=685

PR #686, opened 2023-10-04
Advertise support for python 3.12

alexfikl:
Added on the CI to see if everything works. Python 3.12 just came out two days ago, so probably not all libraries are updated..


alexfikl:
Well this seems to have just worked! I'll go ahead and merge it since it's just for the CI. 
Hopefully it won't cause too many issues (I'll mostly be developing on 3.11 for a while).

PR #687, opened 2023-10-07
docs: use linkcode instead of viewcode

alexfikl:
This links the docs back to Github when clicking the `[source]` buttons (see [here](https://papis.readthedocs.io/en/latest/api.html)). Currently it shows something like this:
https://papis.readthedocs.io/en/latest/_modules/papis/api.html#get_lib_name
and with this change it would redirect to
https://github.com/papis/papis/blob/main/papis/api.py#L36-L46

@alejandrogallo @jghauser What do you think?

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -30,13 +30,9 @@ sys.path.append(os.path.join(os.path.dirname(__file__), "..", ".."))
 # ones.
 extensions = [
     "sphinx.ext.autodoc",
-    "sphinx.ext.coverage",
-    "sphinx.ext.doctest",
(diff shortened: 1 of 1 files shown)


jghauser:
I like this! Our source code is on Github and I think it makes sense to link there whenever we refer to it. Github's interface is much better suited to exploring the source.

alexfikl:
> I like this! Our source code is on Github and I think it makes sense to link there whenever we refer to it. Github's interface is much better suited to exploring the source.

Thanks for taking a look! I'll go ahead and merge it in and we can tweak it if necessary :rocket:

PR #695, opened 2023-10-31
doctor: add check for duplicate entries

alexfikl:
This adds a new doctor check for duplicates inside a list. The intended use case was to check if the same file was added multiple times / the same tag was added multiple times / the same author was added multiple times, etc.

@jghauser Inspired by some of your work on the update command.


alexfikl:
I'll go ahead and merge this, since it's a nice addition to `duplicated-keys`.

PR #709, opened 2023-11-16
Enhance python formatter with extra conversions

alexfikl:
This implements a special Formatter for our `PythonFormatter`
https://docs.python.org/3/library/string.html#string.Formatter
that supports some extra operations to make it slightly less likely to switch to the `jinja2` formatter:
* It adds `l`, `u`, `t` and `c` conversions for `lower`, `upper`, `title` and `capitalize`. It also adds a `y` conversion that calls `slugify`! Format strings using these look like
```
"{doc[title]!t}"
```
* It adds a funky formatting spec `S` that can select words from a string. This basically does `" ".join(value.split(" ")[istart:iend])`. It looks something like this to selects words 1 to 3
```
"{doc[title]:1.3S}"
```

@alejandrogallo @jghauser This was mostly an experiment because I stumbled upon that `Formatter` class (didn't know it existed). What do you think? Is it worth adding something like this or is it too unexpected for people used to `str.format`?

### DIFF
--- a/papis/format.py
+++ b/papis/format.py
@@ -1,3 +1,4 @@
+import string
 from typing import Any, Dict, Optional
 
 import papis.config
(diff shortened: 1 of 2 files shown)


alejandrogallo:
sorry for the delay, this looks awesome actually, for me it's fine to put this,
it's quite useful, maybe it would be also cool to select an element from a list like from `authors_list`

alejandrogallo:
ah right, yes, if authors_list is empty though, does it explode?

alexfikl:
> ah right, yes, if authors_list is empty though, does it explode?

Hm, using slices should work.. so doing `{doc[author_list][0:1]}` will just return an empty list if there's no authors but `{doc[author_list][0]}` will raise an exception.

alexfikl:
I'll go ahead and merge this, since it got a thumbs up and it generally shouldn't break anything :grin:

PR #711, opened 2023-11-24
config: add support for formatted keys

### LINKED ISSUE #662: Formatter specific defaults
> I think what might be a better solution is for sensible defaults to be set for all relevant options when you enable the jinja formatter. I think I have a pretty complete setup for jinja, happy to contribute them :)

_Originally posted by @jghauser in https://github.com/papis/papis/issues/661#issuecomment-1709585852_

Currently the defaults in `papis/defaults.py` use the `python` formatter and setting any other formatter means the user needs to provide values for all of these. Ideally we would help them along by setting some values ourselves.

This is a bit difficult to do 100% because formatters and options can be provided by third-party plugins. Possible issues:
* How would a third-party formatter provide defaults in its style?
* How would third-party commands / exporters / downloaders / etc provide defaults for all formatters?
* Would it be sufficient to just set values for plugins that are shipped with papis?

alexfikl:
This adds support for entries like these in the configuration file (see discussion in #662):
```
[settings]
add-file-name = {doc[year]}-{doc[author]}-{doc[title]}
ref-format.jinja2 = {{ doc.author_list|slice(3)|join("", attribute="family") }}{{ doc.year }}
```
where the first setting uses the default formatter (whatever `formatter` is set to) and the second setting overwrites the default and forces the `jinja2` formatter. In general, `key[.formatter]` is how this looks like.

For this, it adds some functionality:
* `papis.config.getformattedstring`: this looks thorough the config file for any version of `key.[formatter]` and returns the last one it finds.
* `papis.strings.FormattedString`: a class that's just a `(formatter, string)` tuple and gets passed to `papis.format.format` to allow picking the formatter for each individual string. This is mostly inspired by `prompt_toolkit.FormattedText`.
* `papis.cli.FormattedStringParamType`: a `click.ParamType` that automatically converts command-line parameters to `FormattedString`. It also shows a nice text `--file-name FORMATTED-TEXT` in the command line to mark entries that can be format strings.
* Update the rest of the code to use the above where appropriate!

TODO:
* [x] Need to check what other keys should use formatted strings.
* [x] Not sure the folder name construction in `papis.commands.add::run` works correctly with formatted strings. Need to add some tests for that.
* [x] Probably needs some more testing to make sure nothing is broken. (Currently using this daily, so hopefully will catch some things)

Fixes #662.

### DIFF
--- a/tests/test_format.py
+++ b/tests/test_format.py
@@ -33,6 +33,16 @@ def test_python_formatter(tmp_config: TemporaryConfiguration) -> None:
     assert papis.format.format("{doc[title]:.2S}", data) == "The Phantom"
     assert papis.format.format("{doc[title]:2S}", data) == "The Phantom"
 
+    # test out a jinja2 format
+    from papis.strings import FormattedString
+    assert len(papis.format.FORMATTER) == 1
+    assert (
+        papis.format.format(
+            FormattedString("jinja2", "{{ doc.author }}: {{ doc.title }}"),
+            document)
+        == "Fulano: A New Hope")
+    assert len(papis.format.FORMATTER) == 2
+
 
 @pytest.mark.config_setup(settings={"formatter": "jinja2"})
 def test_jinja_formatter(tmp_config: TemporaryConfiguration) -> None:
@@ -57,3 +67,35 @@ def test_jinja_formatter(tmp_config: TemporaryConfiguration) -> None:
             "{{ doc.author }}: {{ doc.title }} ({{ doc.blahblah }})",
             data)
         == ": The Phantom Menace ()")
+
+    # test out a python format
+    assert len(papis.format.FORMATTER) == 1
+    from papis.strings import FormattedString
+    assert (
+        papis.format.format(
+            FormattedString("python", "{doc[author]}: {doc[title]}"),
+            document)
+        == "Fulano: A New Hope")
+    assert len(papis.format.FORMATTER) == 2
+
+
+def test_overwritten_keys(tmp_config: TemporaryConfiguration) -> None:
+    pytest.importorskip("jinja2")
+
+    import papis.config
+
+    papis.config.set("ref-format.jinja2", "{{ doc.author|lower }}{{ doc.year }}")
+    document = papis.document.from_data({
+        "author": "Fulano", "year": 2020, "title": "A New Hope"
+    })
+
+    fmt = papis.config.getformattedstring("ref-format")
+    assert fmt.formatter == "jinja2"
+
+    ref = papis.format.format(fmt, document)
+    assert ref == "fulano2020"
+
+    from papis.bibtex import create_reference
+
+    ref = create_reference(document, force=True)
+    assert ref == "fulano2020"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,30 @@
+# VERSION NEXT (TDB)
+
+## Features
+
+### Major: Improved formatter support ([#711](https://github.com/papis/papis/pull/711))
+
+Papis supports [formatter plugins](https://papis.readthedocs.io/en/latest/configuration.html#config-settings-formatter)
+that act on certain configuration settings that can depend on the current
+document. Until now, when changing from one formatter to another, all settings
+needed to be rewritten (including default ones, since they used the `python`
+formatter).
+
+It is now possible to use the following syntax to set the formatter per
+configuration key
+```ini
+    [settings]
+    ref-format.jinja2 = {{ doc.author_list | slice(3) | join("", attribute="family") }}{{ doc.year }}
+```
+
+The syntax is always `key[.formatter]`. The formatted strings are searched
+alphabetically and the last one is picked, i.e. if both `key.python` and
+`key.jinja2` are provided, the `python` version will be chosen regardless of the
+order in the configuration file. If no formatter is provided for a formatted
+string of this type, then it will fall back to the default formatter set by the
+`formatter` setting. All default settings are now clearly marked as using the
+`python` formatter, so they no longer need to be rewritten when changing formatters.
+
 # VERSION 0.14.1 (March 1, 2025)
 
 ## Features
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -57,7 +57,6 @@ nitpick_ignore_regex = [
     ["py:class", r".*SubRequest"],
     ]
 
-
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ["_templates"]
 
--- a/doc/source/developer_reference.rst
+++ b/doc/source/developer_reference.rst
@@ -72,6 +72,8 @@ Developer API reference
 ``papis.format``
 ----------------
 
+.. autoclass:: papis.strings.FormattedString
+
 .. automodule:: papis.format
     :members:
 
--- a/papis/bibtex.py
+++ b/papis/bibtex.py
@@ -492,9 +492,11 @@ def create_reference(doc: Dict[str, Any], force: bool = False) -> str:
         return ref
 
     # Otherwise, try to generate one somehow
-    ref_format = papis.config.get("ref-format")
-    if ref_format is not None:
-        ref = papis.format.format(str(ref_format), doc, default="")
+    try:
+        ref_format = papis.config.getformattedstring("ref-format")
+        ref = papis.format.format(ref_format, doc, default="")
+    except ValueError:
+        ref = ""
 
     if not ref:
         ref = str(doc.get("doi", ""))
--- a/papis/cli.py
+++ b/papis/cli.py
@@ -11,6 +11,27 @@ import papis.database
 DecoratorCallable = Callable[..., Any]
 
 
+class FormattedStringParamType(click.ParamType):
+    #: Name of the parameter type (shown in the command-line).
+    name: str = "formatted-text"
+
+    def convert(self,
+                value: Any,
+                param: Optional[click.Parameter],
+                ctx: Optional[click.Context]) -> Any:
+        from papis.strings import FormattedString
+
+        # NOTE: this is required to handle default values which have a formatter
+        # already set and we do not want to remove it
+        if isinstance(value, FormattedString):
+            return value
+
+        return str(value)
+
+    def __repr__(self) -> str:
+        return "FORMATTEDSTRING"
+
+
 def bool_flag(*args: Any, **kwargs: Any) -> DecoratorCallable:
     """A wrapper to :func:`click.option` that hardcodes a boolean flag option."""
     # NOTE: we set the flag_value regardless because the default might be a
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -182,6 +182,7 @@ def get_file_name(
         doc: papis.document.Document,
         original_filepath: str,
         suffix: str = "",
+        file_name_format: Optional[papis.strings.AnyString] = None,
         base_name_limit: int = 150) -> str:
     warn("'get_file_name' is deprecated and will be removed in the next "
          "version. Use 'papis.paths.get_document_file_name' instead.",
@@ -212,8 +213,8 @@ def ensure_new_folder(path: str) -> str:
 
 def run(paths: List[str],
         data: Optional[Dict[str, Any]] = None,
-        folder_name: Optional[str] = None,
-        file_name: Optional[str] = None,
+        folder_name: Optional[papis.strings.AnyString] = None,
+        file_name: Optional[papis.strings.AnyString] = None,
         subfolder: Optional[str] = None,
         base_path: Optional[str] = None,
         batch: bool = False,
@@ -426,17 +427,17 @@ def run(paths: List[str],
     help="Pick from existing subfolders")
 @click.option(
     "--folder-name",
-    help="Name for the document's folder (papis format)",
-    default=lambda: papis.config.getstring("add-folder-name"))
+    help="Name format for the document main folder",
+    type=papis.cli.FormattedStringParamType(),
+    default=lambda: papis.config.getformattedstring("add-folder-name"))
 @click.option(
     "--file-name",
-    help="File name for the document (papis format)",
+    help="File name format for the document",
+    type=papis.cli.FormattedStringParamType(),
     default=None)
 @click.option(
     "--from", "from_importer",
-    help="Add document from a specific importer ({})".format(
-        ", ".join(papis.importer.available_importers())
-    ),
+    help="Add document from a specific importer",
     type=(click.Choice(papis.importer.available_importers()), str),
     nargs=2,
     multiple=True,
@@ -483,8 +484,8 @@ def cli(files: List[str],
         set_list: List[Tuple[str, str]],
         subfolder: str,
         pick_subfolder: bool,
-        folder_name: str,
-        file_name: Optional[str],
+        folder_name: papis.strings.AnyString,
+        file_name: Optional[papis.strings.AnyString],
         from_importer: List[Tuple[str, str]],
         batch: bool,
         confirm: bool,
--- a/papis/commands/addto.py
+++ b/papis/commands/addto.py
@@ -101,7 +101,8 @@ def run(document: papis.document.Document,
               type=click.Path(exists=True))
 @click.option("-u", "--urls", help="URLs to documents", multiple=True)
 @click.option("--file-name",
-              help="File name for the document (papis format)",
+              help="File name format for the document",
+              type=papis.cli.FormattedStringParamType(),
               default=None)
 @click.option(
     "--link/--no-link",
--- a/papis/commands/bibtex.py
+++ b/papis/commands/bibtex.py
@@ -131,6 +131,7 @@ import papis.cli
 import papis.config
 import papis.format
 import papis.logging
+import papis.strings
 from papis.commands import AliasedGroup
 from papis.commands.explore import get_explorer_by_name
 
@@ -338,10 +339,9 @@ def cli_open(ctx: click.Context) -> None:
 @cli.command("edit")
 @click.help_option("-h", "--help")
 @click.option("-s", "--set", "set_tuples",
-              help="Update document's information with key value. "
-              "The value can be a papis format.",
+              help="Update a document with key value pairs",
               multiple=True,
-              type=(str, str),)
+              type=(str, papis.cli.FormattedStringParamType()),)
 @papis.cli.all_option()
 @click.pass_context
 def cli_edit(ctx: click.Context,
@@ -384,11 +384,12 @@ def cli_edit(ctx: click.Context,
 
         if set_tuples:
             for k, v in set_tuples:
+                kp, vp = papis.strings.process_formatted_string_pair(k, v)
                 try:
-                    located[k] = papis.format.format(v, located)
+                    located[kp] = papis.format.format(vp, located)
                 except papis.format.FormatFailedError as exc:
                     logger.error("Could not format '%s' with value '%s'.",
-                                 k, v, exc_info=exc)
+                                 kp, vp, exc_info=exc)
 
             save_doc(located)
         else:
--- a/papis/commands/browse.py
+++ b/papis/commands/browse.py
@@ -113,7 +113,7 @@ def run(document: papis.document.Document,
         import urllib.parse
         params = {
             "q": papis.format.format(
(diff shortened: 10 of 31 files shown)


hermlon:
I just tried your fork because I need this feature, but I somehow can't start the `ref-format` with a lowercase letter. I'm using this Jinja format string:
```
ref-format.jinja2 = {{ doc.author_list[0]['family']|lower }}{{ doc.year }}
```

I guess there is some code upper-casing the `ref-format` again, since the Jinja2 documentation clearly states that the `lower` filter lowers all the characters.

alexfikl:
> I just tried your fork because I need this feature, but I somehow can't start the `ref-format` with a lowercase letter. I'm using this Jinja format string:
> 
> ```
> ref-format.jinja2 = {{ doc.author_list[0]['family']|lower }}{{ doc.year }}
> ```

Can confirm that this doesn't work as expected. I'll look into it!
Thanks a lot for testing this :heart: it touches a lot of parts, so needs quite a bit more testing besides the unit tests..

alexfikl:
@hermlon Turns out my case was just a config error. I added a test for the case you mentioned (I think), i.e. setting a `ref-format.jinja2`, and it seems to work as expected. 

Can you give more details about your setup? Do other things work, e.g. does setting `ref-format.jinja2 = {{ doc.year }}` give the expected change in the ref? Can you also set `ref-format` in the config to see if it's picking that one by mistake?

tests/test_format.py
```diff
@@ -52,3 +62,27 @@ def test_jinja_formatter(tmp_config: TemporaryConfiguration) -> None:
...
+
+    import papis.config
+
+    papis.config.set("ref-format.jinja2", "{{ doc.author|lower }}{{ doc.year }}")
+    document = papis.document.from_data({
+        "author": "Fulano", "year": 2020, "title": "A New Hope"
+    })
+    ref = papis.format.format(papis.config.getformattedstring("ref-format"),
+                              document)
+    assert ref == "fulano2020"
```
alexfikl, line 88: @hermlon This is the chunk of the test that I was talking about :grin:

OopsYao:
> I just tried your fork because I need this feature, but I somehow can't start the `ref-format` with a lowercase letter. I'm using this Jinja format string:
> 
> ```
> ref-format.jinja2 = {{ doc.author_list[0]['family']|lower }}{{ doc.year }}
> ```
> 
> But I get the `ref: Sanderson12`
> 
> I guess there is some code upper-casing the `ref-format` again, since the Jinja2 documentation clearly states that the `lower` filter lowers all the characters.

I tried the commit 126825cb; it works well, no abnormalities found yet. 👏 One thing is that something like `papis update --set ref.jinja2 [ref-format]` is not supported currently. Don't know if this needs extra heavy work.

alexfikl:
> I tried the commit [126825c](https://github.com/papis/papis/commit/126825cb47cb2745e12779e1839aa7a2952a7dcf); it works well, no abnormalities found yet. 👏 

Awesome! Thank for you testing!

> One thing is that something like `papis update --set ref.jinja2 [ref-format]` is not supported currently. Don't know if this needs extra heavy work.

Hm, it shouldn't be too hard to implement. I'll look into it! There are a few more places that take values that should be formatted from the command line, so we need some nice solution for them.

alexfikl:
@OopsYao The latest version should also support `papis update --set ref.jinja2 [format]`. Let me know if you try it and something doesn't work (I just very briefly tested it)!

OopsYao:
Great work! I tried `papis update --set` with something like `tags.jinja2` as well as `ref`, `publisher` and etc. They all work as expected.

> EDIT: I've looked a bit at this and there's two issues here:
> 
>     * For the `--set key value` option that a few of the commands take, we can just use `key[.formatter]` and select the formatter that way. This shouldn't be hard to implement.
> 
>     * Other arguments, e.g. `papis add --file-name [FORMAT] ...`, don't really have a way to specify the formatter. I'm not sure what a good way to do that would be. We can always recommend using
> 
> 
> ```
> papis --set add-file-name.jinja2 [FORMAT] add ...
> ```
> 
> instead, but not sure that's very nice..

I see, but this is indeed a workaround. And when I try `papis --set add-file-name.jinja2 [FORMAT] add ...`, the generated file name is still lower case in spite of the usage of `upper` in jinja2 format string. It seems like there is another formatter for the file name.

alexfikl:
> Great work! I tried `papis update --set` with something like `tags.jinja2` as well as `ref`, `publisher` and etc. They all work as expected.

Great thanks! Glad to hear that it works for you! We can probably fix any remaining issues as they pop up.

> I see, but this is indeed a workaround. And when I try `papis --set add-file-name.jinja2 [FORMAT] add ...`, the generated file name is still lower case in spite of the usage of `upper` in jinja2 format string. It seems like there is another formatter for the file name.

Yeah, pretty much all filenames are "cleaned" over here
https://github.com/papis/papis/blob/0b9be8bbfb47a856fd2530a14cab818dbeb0149b/papis/utils.py#L287-L292
which I don't think allows uppercase letters. References from `ref` will also be cleaned a bit, so they won't be exactly what you put in your format.

kiike:
I have been trying this feature for a couple of weeks and I really like it!

I think the Jinja2 formatting is a great addition, and the formatted keys give a lot of flexibility.

While the config is very compact, I find it unusual. Maybe I would've expected:

```
add-folder.format = {{doc.author}}
add-folder.formatter = jinja2
```

alexfikl:
> I have been trying this feature for a couple of weeks and I really like it!

Thank you for trying this out! :heart: 

> While the config is very compact, I find it unusual. Maybe I would've expected:
> 
> ```
> add-folder.format = {{doc.author}}
> add-folder.formatter = jinja2
> ```

Why is this better? I find it a bit repetitive for multiple options, e.g.
```
header-format.format = ...
header-format.formatter = jinja2
add-file-name.format = ...
add-file-name.formatter = jinja2
add-folder.format = ...
add-folder.formatter = jinja2
```
and annoyingly allows these to be set apart, e.g.
```
header-format.format = ...
picktool = ...
opentool = ...
add-confirm = ...
header-format.formatter = jinja2
```

> Thinking about it, this all adds complexity, while settling for Jinja2 as default and only formatter would simplify the code. The default Python formatter is simpler, but if we have a well-known syntax, Jinja2, which is more powerful, could that be the only option?

I'm not sure I agree with this. 

First, the jinja2 formatter has been available for a long time and you can already switch to it if it feels more natural. The main issue with that was just that all the default values in Papis are defined for the Python formatter, so it's a bit of a drag to redefine all of them.

I also don't think removing the plugins infrastructure for the formatters and having the jinja2 one as the only implementation is likely to happen. At least I have some custom formatters used to hook into formatting references and file names that would be quite annoying to do otherwise..

kiike:
> > I have been trying this feature for a couple of weeks and I really like it!
> 
> @kiike Thank you for trying this out! ❤️
> 
> > While the config is very compact, I find it unusual. Maybe I would've expected:
> > ```
> > add-folder.format = {{doc.author}}
> > add-folder.formatter = jinja2
> > ```
> 
> Why is this better? I find it a bit repetitive for multiple options, e.g.
> 
> ```
> header-format.format = ...
> header-format.formatter = jinja2
> add-file-name.format = ...
> add-file-name.formatter = jinja2
> add-folder.format = ...
> add-folder.formatter = jinja2
> ```

You're right. It gets tedious easily. 

 
> I'm not sure I agree with this.
> 
> First, the jinja2 formatter has been available for a long time and you can already switch to it if it feels more natural. The main issue with that was just that all the default values in Papis are defined for the Python formatter, so it's a bit of a drag to redefine all of them.

Yes. Having to define manually all of them from scratch is quite a drag. Plus it's not straightforward to see the changes without being put off by all the broken strings, or exceptions, and errors and warnings. 
 
> I also don't think removing the plugins infrastructure for the formatters and having the jinja2 one as the only implementation is likely to happen. At least I have some custom formatters used to hook into formatting references and file names that would be quite annoying to do otherwise..

Makes sense. Especially in your case, with the custom formatters.

Thank you for explaining this!

alexfikl:
> Yes. Having to define manually all of them from scratch is quite a drag. Plus it's not straightforward to see the changes without being put off by all the broken strings, or exceptions, and errors and warnings.

Yeah, not quite sure how to solve that in a nice way. That was mostly what this is trying to work around: instead of rewriting all the config settings using e.g. jinja2, you can just rewrite the ones where you need a bit more power.

There may be something to be said about switching to jinja2 as the default though. Especially if this goes in, we can probably do that in a backwards compatible way.

pmiam:
Great work! I remember getting excited for this months ago and I am again now that it is looking very close to ready.

I've just checked out your branch. I plan to implement the ref name algorithm defined in the `bibtex-generate-autokey` function from emacs' `bibtex.el` as a fancy jinja template. (Those citation keys are the only part of my old bibliography that I miss.)

In the process I'll probably break many things, I'll report my findings here.

alexfikl:
> I remember getting excited for this months ago and I am again now that it is looking very close to ready.

I've been using this for quite a while now too and it's been working without an issue (didn't try to force it too much though). The main reason I'm still holding off on it is because the current release already has enough big changes, so it seemed better to merge it after 0.14 is out.

> I've just checked out your branch. I plan to implement the ref name algorithm defined in the `bibtex-generate-autokey` function from emacs' `bibtex.el` as a fancy jinja template. (Those citation keys are the only part of my old bibliography that I miss.)
> 
> In the process I'll probably break many things, I'll report my findings here.

That would be great!

pmiam:
For what it's worth, I did break many things, but I'm near completely sure that none are in this PR.
Unfortunately I'm out of time, so I'll elaborate later.

pmiam:
I have continued to use this branch for some time and it has served me well. Everything I broke in my earlier escapade was due to my own experiments and config.py. I've opened a new PR to unlock some more power in-line with this feature set, though it does nothing to mitigate the possibility of other adventurous users writing their own volatile configs -- I reckon that comes with the territory.

Most of the CI builds are failing for some reason. I'll look into that hopefully within the next month.

pmiam:
Seems I found a bug. I wrote an alias to help me generate my pretty new refs after adding a library entry from e.g. a bibtex file

```
alias refgen='papis update -s ref.jinja2 "$(papis config -s lit --json | jq ".\"ref-format.jinja2\"")"'
```

For reasons I don't fully understand, this delivers my "lit" library's format string to `papis.update.run_set` as a `FormattedString` object with formatter attr `None`. Therefore the formatter value from the command line was silently dropped in `papis.strings.process_formatted_string_pair`. I patched it, but my solution might be a bit brutish:

```
From 55b883e2f1e805c9d8e972bfb01b559a58181291 Mon Sep 17 00:00:00 2001
From: Panayotis Manganaris <panos.manganaris@gmail.com>
Date: Thu, 31 Oct 2024 03:01:03 -0400
Subject: [PATCH] strings [BUGFIX] guarantee FormattedString value has
 formatter attr

---
 papis/strings.py | 5 ++---
 1 file changed, 2 insertions(+), 3 deletions(-)

diff --git a/papis/strings.py b/papis/strings.py
index 87e70ed4..c516663f 100644
--- a/papis/strings.py
+++ b/papis/strings.py
@@ -65,15 +65,14 @@ def process_formatted_string_pair(
     :param value: a unformatted value.
 
     :returns: a ``(key, value)`` pair, where the formatter was removed from the
-        key and the value is guaranted to be a :class:`FormattedString`.
+        key and the value is guaranteed to be a :class:`FormattedString`.
     """
     if "." in key:
         key, formatter = key.rsplit(".", maxsplit=1)
     else:
         formatter = None
 
-    if not isinstance(value, FormattedString):
-        value = FormattedString(formatter, value)
+    value = FormattedString(formatter, str(value))
 
     return key, value
 
-- 
2.47.0
```

alexfikl:
> Seems I found a bug. I wrote an alias to help me generate my pretty new refs after adding a library entry from e.g. a bibtex file

Thanks for reporting it! I'll take a look probably over the weekend.

Also noticed another thing that's missing. `papis config` isn't updated to work well with this, as this doesn't do what it should
```
papis config -s settings ref-format.jinja2
```
It currently considers `ref-format` as a section and fails miserably :(

pmiam:
> Also noticed another thing that's missing. `papis config` isn't updated to work well with this

Indeed. The current form of `papis.config.general_get` also causes `papis rename` to throw because https://github.com/papis/papis/blob/f390fd79df82b9458ac5f909f5cde5d1b919c9b4/papis/config.py#L404 always returns false if the user specifies `.formatter` in a config key.

also, `papis rename --folder-name` doesn't understand jinja format strings.

alexfikl:
> Seems I found a bug. I wrote an alias to help me generate my pretty new refs after adding a library entry from e.g. a bibtex file
> 
> ```
> alias refgen='papis update -s ref.jinja2 "$(papis config -s lit --json | jq ".\"ref-format.jinja2\"")"'
> ```

This should be fixed in the latest version with https://github.com/papis/papis/compare/e3f7629df475631a93b3e009f314e1cff8e1717e..bfbc54d909051f2c835bee1ab57bea3f3c081f9f.

The issue was indeed with that `process_formatted_string_pair` and how it interacted with the command line arguments transformed through `FormattedStringParamType`. It now insists on the formatter from the `key` as expected.

alexfikl:
> also, papis rename --folder-name doesn't understand jinja format strings.

@pmiam This is a problem for other commands as well and I'm not sure how to fix it in an intuitive manner. For example if you just go
```
papis rename --folder-name "SOME_FORMATTING_STRING" <QUERY>
```
there's no way to know what formatter that text is supposed to have. It works for `papis update` because it uses the key name to specify the formatter like `--set key.jinja2 value`. For something like `rename`, I would suggest
```
papis --set formatter jinja2 rename --folder-name "JINJA2_FORMAT" <QUERY>
```
Also, if you're using another formatter than then one you set in the config in that many places, it might be worth switching your defaults?

pmiam:
> I would suggest
> 
> ```
> papis --set formatter jinja2 rename --folder-name "JINJA2_FORMAT" <QUERY>
> ```

Ah, that's actually quite an elegant solution. I hadn't actually realized the set flag could be used like that. We could call attention to it somewhere in the docs.

The only alternative I see so far is a campaign to overhaul the CLI. I'm not sure how stable the basic interface is. I have seen other PRs working to merge/extend other commands, but I doubt deprecating basic flags like `--folder-name` in favor of universally relying on `--set` is something the main team is willing to do. (Or, even that something so drastic is needed).
 
> might be worth switching your defaults?

Indeed, I might switch my defaults at this point, I've become enamored with my jinja-given powers. It's important to handle this middle ground, though, as it's an explicit feature of this PR.

pmiam:
> > Indeed. The current form of papis.config.general_get also causes papis rename to throw [...]
> 
> Can you post a traceback for that? I'm not sure in what scenario that would happen.

To be more precise, the if-condition I referenced is always false in the case when I set my `add-folder-name.jinja2` key under a custom library subsection, and leave the default under the global settings section alone. 

e.g. with config:

```
[settings]
default-library = lit

[lit]
dir = ~/doc/lit
add-folder-name.jinja2 = {{ doc.author_list|map(attribute="family")|join }}
```

I run `papis rename` on your latest commit dafed8ffa034a813fac0fd5387627d7faf516bdf

```
  File "/home/panos/.local/bin/papis", line 8, in <module>
    sys.exit(run())
             ^^^^^
  File "/home/panos/src/papis/.direnv/python-3.12/lib/python3.12/site-packages/click/core.py", line 1157, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/panos/src/papis/.direnv/python-3.12/lib/python3.12/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/panos/src/papis/.direnv/python-3.12/lib/python3.12/site-packages/click/core.py", line 1686, in invoke
    sub_ctx = cmd.make_context(cmd_name, args, parent=ctx)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/panos/src/papis/.direnv/python-3.12/lib/python3.12/site-packages/click/core.py", line 943, in make_context
    self.parse_args(ctx, args)
  File "/home/panos/src/papis/.direnv/python-3.12/lib/python3.12/site-packages/click/core.py", line 1408, in parse_args
    value, args = param.handle_parse_result(ctx, opts, args)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/panos/src/papis/.direnv/python-3.12/lib/python3.12/site-packages/click/core.py", line 2396, in handle_parse_result
    value, source = self.consume_value(ctx, opts)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/panos/src/papis/.direnv/python-3.12/lib/python3.12/site-packages/click/core.py", line 2934, in consume_value
    value, source = super().consume_value(ctx, opts)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/panos/src/papis/.direnv/python-3.12/lib/python3.12/site-packages/click/core.py", line 2290, in consume_value
    value = self.get_default(ctx)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/home/panos/src/papis/.direnv/python-3.12/lib/python3.12/site-packages/click/core.py", line 2868, in get_default
    return super().get_default(ctx, call=call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/panos/src/papis/.direnv/python-3.12/lib/python3.12/site-packages/click/core.py", line 2268, in get_default
    value = value()
            ^^^^^^^
  File "/home/panos/src/papis/papis/commands/rename.py", line 112, in <lambda>
    default=lambda: papis.config.getstring("add-folder-name"))
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/panos/src/papis/papis/config.py", line 479, in getstring
    raise ValueError("Key '{}' should be a string: '{}'".format(key, result))
ValueError: Key 'add-folder-name' should be a string: ''
```

alexfikl:
>   File "/home/panos/src/papis/papis/commands/rename.py", line 112, in <lambda>
>     default=lambda: papis.config.getstring("add-folder-name"))

Oops, this just looks like I forgot to update the `rename` command! This should be `getformattedstring("add-folder-name")` :grin:

alexfikl:
> I'm not sure how stable the basic interface is. I have seen other PRs working to merge/extend other commands, but I doubt deprecating basic flags like --folder-name in favor of universally relying on --set is something the main team is willing to do. (Or, even that something so drastic is needed).

I don't think the interface is very stable.. or at least we don't do any particular testing to ensure it stays too compatible :( That being said, stuff like `--folder-name` and `--file-name` was mostly added as helpful shorthands for 
```bash
papis --set add-folder-name "MY_SPECIAL_FORMAT" add <SOMETHING>`
```
I'm not sure if we necessarily want to deprecate them..

pmiam:
I'm happy to report no more issues over the last week. Your rename fix has performed perfectly and it's helped me re-organize my largest library's file tree without pains.

pmiam:
I have another little patch to offer:
```
From 2aee8183e23dcfa49cfc94db89cae91be80e292f Mon Sep 17 00:00:00 2001
From: Panayotis Manganaris <panos.manganaris@gmail.com>
Date: Mon, 18 Nov 2024 14:35:47 -0500
Subject: [PATCH] notes [ITER] get formatted notes-name string

---
 papis/notes.py | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/papis/notes.py b/papis/notes.py
index bcc28176..4e643366 100644
--- a/papis/notes.py
+++ b/papis/notes.py
@@ -31,7 +31,7 @@ def notes_path(doc: papis.document.Document) -> str:
     from papis.paths import normalize_path
 
     if not has_notes(doc):
-        notes_name = papis.format.format(papis.config.getstring("notes-name"), doc,
+        notes_name = papis.format.format(papis.config.getformattedstring("notes-name"), doc,
                                          default="notes.tex")
         doc["notes"] = normalize_path(notes_name)
         papis.api.save_doc(doc)
-- 
2.47.0
```

alexfikl:
> I have another little patch to offer:

Nice catch! Added it in the latest rebase.

alexfikl:
I'll go ahead and merge this as soon as the CI passes. It's been long overdue and we can always fix any issues that it introduces :grin:

PR #715, opened 2023-12-07
docs: instruction on how to install papis on/with Guix

stasvlasov:


alexfikl:
This looks good to me, but it would be nicer to add that code in the top directory (like the `flake.nix` file) or at least in the `contrib` directory. 

What do you think? Would you be willing to maintain that a bit if it gets bug reports?

doc/source/install.rst
```diff
@@ -72,6 +72,285 @@ and start hacking it with:
...
+                            ))
+      (home-page "https://github.com/Knio/dominate/")
+      (synopsis
+       "Dominate is a Python library for creating and manipulating HTML documents using an elegant DOM API.")
+      (description
+       "Dominate is a Python library for creating and manipulating HTML documents using
+  an elegant DOM API.")
+      (license #f)))
+  
+  (define python-arxiv2bib
```
alexfikl, line 258: Papis doesn't use `arxiv2bib` anymore. This should probably be updated?

doc/source/install.rst
```diff
@@ -72,6 +72,285 @@ and start hacking it with:
...
+                        python-coveralls
+                        python-lsp-server
+                        python-sphinx-click
+                        python-sphinx-rtd-theme
+                        python-types-beautifulsoup4
+                        python-types-pygments
+                        python-types-python-slugify
+                        python-types-pyyaml
+                        python-types-requests
+                        python-types-tqdm))
```
alexfikl, line 342: These type annotation packages are only required for testing. Do they need to be in the recipe? Who is this intended for: developers or users?

stasvlasov: The default workflow for guix's "pyproject-build-system" includes testing so ideally these packages would be used during the make process. I, however, could not build papis with passing all its tests so I temporary switched it off. Later, if time allows, I would actually make sure that all tests are passing as for most of the other packages that had to be defined prior to papis package because they are not (yet) in the main guix repo.

doc/source/install.rst
stasvlasov: This is papis package definition for 0.13 release. As I understand it is current
stable version, right? At that point it was still rely on arxiv2bib.

alexfikl: Ah, you're right, my bad!

doc/source/install.rst
alexfikl: Fair enough! Thank you for explaining.

alexfikl:
> This is actually just a temporary fix for guix installation. Ideally I would submit this package definition to main guix repository (if licences of all dependencies allows it) or to alternative guix repos (e.g., nonguix). Otherwise papis repo can 'host' a dedicated to papis guix channel (it just a git repo) that can also provide build binaries for quicker installation.

It would be great to have this in a specialized repository instead, especially the stable releases. Maybe the papis repo can just contain a special Guix package definition for the latest `git` version of the code?

> Yes, I am ok with maintaining this piece) Perhaps somebody can also help me out as I am quite new to Guix (but I like a lot so far).

Thanks! If this gets submitted to some Guix repository, then it will probably be enough to take care of that, so the only issues that could come up here are that the build instructions are no longer up to date.

doc/source/install.rst
```diff
@@ -72,6 +72,285 @@ and start hacking it with:
...
+`guix <https://guix.gnu.org/>`__ package manager installed, you can install
+papis by running:
+
+::
+
+  guix package --install-from-file=papis.scm
+
+Where papis.scm file defines the papis package recipe (this is a first draft, some failing tests are switched off):
+
+.. code:: scheme
```
alexfikl, line 87: This whole chunk of scheme code could be a separate file in `contrib/papis.scm` and here just use
```rst
Where :download:`papis.scm <../../contrib/papis.scm>` file defines the papis package recipe.
```
? We don't need to paste the whole code in the documentation, since it's quite long.

stasvlasov:
I moved recipe to `/contrib`.

> It would be great to have this in a specialized repository instead,
> especially the stable releases. Maybe the papis repo can just contain
> a special Guix package definition for the latest `git` version of the
> code?

I am not sure how to implement the latest `git` version recipe. I think one would need to update it and all dependencies every time manually because guix recipes have to be pinned to a specific commit. Maybe this could be automated with running `guix import pypi

> Thanks! If this gets submitted to some Guix repository, then it will
> probably be enough to take care of that, so the only issues that could
> come up here are that the build instructions are no longer up to date.

No problem) I will take a note to self to update the docs if I see papis added to the guix repos (or if add it there myself).

alexfikl:
> No problem. I will take a note to self to update the docs if I see papis added to the guix repos (or if add it there myself).

That would be great! Thank you for the contribution! I'll go ahead and merge it as soon as the tests pass :rocket:

PR #716, opened 2023-12-11
commands.list: expand shown plugins

alexfikl:
This cleans up and expands `papis list` in the following ways
* Separates `run` into `list_plugins` and `list_documents` to not mix up the two.
* Adds flags for printing exporters, explorers, importers, downloaders, etc. with a short description.
* Also prints libraries and config paths.
* Removes `papis add --list-importers`, `papis doctor --list-checks` and `papis config --list-paths`, since they're now better taken care of by `papis list`.
* Updates the documentation a bit all over the place so that the plugin descriptions look nicer.


alexfikl:
This also adds some colors so the listed downloader plugins look like this

Screenshot_20231211_110430

PR #726, opened 2024-01-04
Fix typo in HACKING.md

mlq:


alexfikl:

PR #729, opened 2024-01-12
Fixed truncated file/dirnames when clean_document_name is called on a…

f0rki:
… paper title that has a directory separator (`/`) in it.

e.g.,
```
papis add 'https://www.usenix.org/conference/osdi21/presentation/zhang'
```
will break since the title is:
```
SanRazoR: Reducing Redundant Sanitizer Checks in C/C++ Programs
```

The default arg `is_path` is set to `True` in `clean_document_name`, which makes it call `os.basename` beforehand. If the document name contains a directory separator, i.e., `/`, then this will essentially truncate the document path/name. In this case to "C++ Programs", which is not intended. Instead the `/` should be removed from the document name.

### DIFF
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -196,10 +196,10 @@ def get_file_name(
         default=""
     )
 
-    file_name_base = papis.utils.clean_document_name(file_name_base)
+    file_name_base = papis.utils.clean_document_name(file_name_base, False)
     if not file_name_base:
         file_name_base = papis.utils.clean_document_name(
-            os.path.basename(original_filepath))
+            os.path.basename(original_filepath), False)
 
     if len(file_name_base) > base_name_limit:
         logger.warning(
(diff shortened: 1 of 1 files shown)


alexfikl:
Hm, I added some similar fixes in #711 for the same reasons!

I'm not quite sure why `clean_document_name` does a `os.path.basename` at all actually? It's very unintuitive and most places seem to want to skip that anyway..

What do you think about just removing the `is_path` flag and the `os.path.basename` in `clean_document_name` instead? Maybe even rename it to `clean_display_string` or something (very open to suggestions)?

f0rki:
> It's very unintuitive and most places seem to want to skip that anyway..

indeed. maybe it is best to just flip the default of the argument to `False`? Then go through the callsites and check if it is used anywhere?

alexfikl:
> indeed. maybe it is best to just flip the default of the argument to `False`? Then go through the callsites and check if it is used anywhere? or just wait something breaks somewhere else :D

Sure, that sounds good too.

alexfikl:
Just remembered this.. on a second thought, switching the flag isn't that much different, so I think this can go in as is!

Thanks for fixing these! :rocket:

PR #730, opened 2024-01-13
papis.serve: add sorting options to the tags endpoint

justinnk:
This adds two further Options besides the refresh icon to select whether to sort the list of tags alphabetically or by their number of occurrences. Formerly, the latter was the default and only option.

Why? I use tags a lot and found myself frequently searching for them on the page. Alphabetical sorting makes that much easier.

### DIFF
--- a/papis/web/tags.py
+++ b/papis/web/tags.py
@@ -51,10 +51,11 @@ def html(pretitle: str, libname: str, tags: Dict[str, int],
                              title="Sort by number of occurrences"):
                         wh.icon("arrow-down-1-9")
                 with wh.container():
-                    # either sort by number of occurences or alphabetical
-                    sort_kwargs = dict(key=lambda k: tags[k], reverse=True)
+                    # either sort by number of occurrences or alphabetical
                     if sort_by == "alpha":
-                        sort_kwargs = dict()
-                    for tag in sorted(tags, **sort_kwargs):
+                    	tags = sorted(tags)
+                    elif sort_by in ("", "numeric"):
+                    	tags = sorted(tags, key=lambda k: tags[k], reverse=True)
+                    for tag in tags:
                         _tag(tag=tag, libname=libname)
     return result
--- a/papis/web/tags.py
+++ b/papis/web/tags.py
@@ -51,14 +51,21 @@ def html(pretitle: str, libname: str, tags: Dict[str, int],
                              title="Sort by number of occurrences"):
                         wh.icon("arrow-down-1-9")
                 with wh.container():
+                    sorted_tags = []
                     # either sort by number of occurrences or alphabetical
                     # if option is not set, use the default config
                     if sort_by == "":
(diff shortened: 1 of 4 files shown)


papis/web/tags.py
```diff
@@ -35,17 +35,26 @@ def tags_list_div(tags: Tags, libname: str) -> None:
...
+         sort_by: str) -> t.html_tag:
     with papis.web.header.main_html_document(pretitle) as result:
         with result.body:
             papis.web.navbar.navbar(libname=libname)
             with wh.container():
                 with t.h1("TAGS"):
                     with t.a(href=f"/library/{libname}/tags/refresh"):
                         wh.icon("refresh")
+                    with t.a(href=f"/library/{libname}/tags?sort=alpha",
+                             title="sort by name"):
```
alexfikl, line 48:
```suggestion
                             title="Sort by name"):
```

papis/web/tags.py
```diff
@@ -35,17 +35,26 @@ def tags_list_div(tags: Tags, libname: str) -> None:
...
+                        wh.icon("arrow-down-a-z")
+                    with t.a(href=f"/library/{libname}/tags?sort=numeric",
+                             title="sort by number of occurences"):
```
alexfikl, line 51:
```suggestion
                             title="Sort by number of occurrences"):
```

papis/web/tags.py
```diff
@@ -35,17 +35,26 @@ def tags_list_div(tags: Tags, libname: str) -> None:
...
+                        wh.icon("arrow-down-1-9")
                 with wh.container():
-                    for tag in sorted(tags,
-                                      key=lambda k: tags[k],
-                                      reverse=True):
+                    # either sort by number of occurences or alphabetical
+                    sort_kwargs = dict(key=lambda k: tags[k], reverse=True)
+                    if sort_by == "alpha":
+                        sort_kwargs = dict()
+                    for tag in sorted(tags, **sort_kwargs):
```
alexfikl, line 58:
```suggestion
                    # either sort by number of occurrences or alphabetical
                    if sort_by == "alpha":
                    	tags = sorted(tags)
                    elif sort_by in ("", "numeric"):
                    	tags = sorted(tags, key=lambda k: tags[k], reverse=True)
                    for tag in tags:
```
? In case we ever want to add more sorting formats?

alexfikl (changes requested):
I'm not user of `papis serve`, but this looks good to me!

justinnk:
Thanks for the suggestions. I now also added a configuration option to set the default sorting strategy when opening the `/tags` endpoint without a query.
If there are no problems, I believe this is ready to merge.

alexfikl:
> Thanks for the suggestions. I now also added a configuration option to set the default sorting strategy when opening the `/tags` endpoint without a query. If there are no problems, I believe this is ready to merge.

This looks good to me too! Thanks for adding the config option to the docs and everything :rocket:

PR #734, opened 2024-01-28
ci: update actions

alexfikl:
Updates the versions for all the GitHub actions we use.


alexfikl:
@f0rki Just FYI, I tried updating `cachix/install-nix-action@v22` to the latest `cachix/install-nix-action@v25` and the `nix build` step failed when building `Whoosh` for some reason, so this keeps the older version for now.

PR #740, opened 2024-02-08
commands.add: add a --move flag

alexfikl:
To complement `--link` and the default `--copy` for the adventurous. The move is only performed only at the end, so that data doesn't just disappear on a weird failure.


jghauser:
Just wanted to say, I really like this. Makes my scripts a bit smoother (and less error-prone)! Thanks :tada:

Issue #738, opened 2024-02-08
papis export --format bibtex escape URL underscores


S1M0N38:
* papis version (`$ papis --version` or commit number): papis, version 0.13

# how to reproduce the issue
Command and output of `papis --log INFO export --format bibtex`:
```bib
[INFO] commands.export: Dumping to STDOUT.
@misc{SDbwLashko2019,
  author = {Lashkov, Alexander},
  journal = {GitHub repository},
  month = {3},
  publisher = {GitHub},
  title = {S\_Dbw},
  url = {https://github.com/alashkov83/S\_Dbw},
  year = {2019},
}
```

This is the result of the following info.yaml
```yaml
author: Lashkov, Alexander
author_list:
- family: Lashkov
  given: Alexander
commit: 4e2b7619e9cbe3f04aca1f5ca440198fb06d3390
files: []
journal: GitHub repository
month: '3'
papis_id: 0b5967343dc5e70f99d98a88b884a586
publisher: GitHub
ref: SDbwLashko2019
time-added: 2024-02-08-00:44:55
title: S_Dbw
type: misc
url: 'https://github.com/alashkov83/S_Dbw'
year: '2019'
```
---

When I add the produce bibtex to a tex file and compile it, I get the following:
```
[91] Alexander Lashkov. S_Dbw. Mar. 2019. URL: https://github.com/alashkov83/S%5C_Dbw.
```

This is not the correct URL (The right one is https://github.com/alashkov83/S_Dbw).
I suspect that the `papis export --format bibtex` command is attempting to escape all underscores (e.g the title of the reference is fine). However, escaping the underscore inside the URL results in the `%5C` code which I presume is the representation of `\` in a URL.

Could it be that not all underscores need to be escaped by `papis export --format bibtex`?

alexfikl:
Hm, can you share your config a bit? I've just tried this with PDFLaTeX and it seems to work, i.e. it compiles and both links are correct (somehow?)
```latex
\documentclass{article}

\usepackage[T1]{fontenc}
\usepackage{hyperref}

\begin{filecontents*}{\jobname.bib}
    @misc{code1,
        title = {Code without underscore},
        url = {https://github.com/alashkov83/S_Dbw}
    }
    @misc{code2,
        title = {Code with underscore},
        url = {https://github.com/alashkov83/S\_Dbw}
    }
\end{filecontents*}

\begin{document}

\cite{code1} and \cite{code2}

\bibliographystyle{IEEEtran}
[...]

PR #744, opened 2024-02-10
bibtex: more strict version for bibtexparser (#738)

S1M0N38:
Bibtexparser 2.x will not expose latexenc module


alexfikl:
Thanks! We can merge this as soon as the tests pass.

I did check out v2 a bit and it has quite a few breaking changes, but we'll cross that bridge when we get to it.

PR #745, opened 2024-02-10
config: use platformdirs

alexfikl:
This switches to using [platformdirs](https://platformdirs.readthedocs.io/en/latest/) for all our "getting a standard path" functions. This will mainly improves the situation on other platforms, i.e.
* On Windows the configuration should go in `%USERPROFILE%\AppData\Local\$appauthor\$appname`
* On macOS the configuration should go in `~/Library/Application Support/$appname/$version`

To help with migrating:
* config: When loading a config, it checks if the previous XDG-based `config_folder` exists and copies it over with a warning.
* cache: Nothing is done here, since we can just regenerate the cache and nothing is lost. It's not too hard to copy over older caches, so I'm open to suggestions.

On Linux, this shouldn't change anything at all really, so if someone with a different OS can chime in, that would be very appreciated!

### DIFF
--- a/papis/config.py
+++ b/papis/config.py
@@ -3,6 +3,7 @@ import configparser
 from typing import Dict, Any, List, Optional, Callable
 
 import click
+import platformdirs
 
 import papis.exceptions
 import papis.library
@@ -53,7 +54,7 @@ class Configuration(configparser.ConfigParser):
         self.file_location: str = get_config_file()
         self.default_info: PapisConfigType = {
             "papers": {
-                "dir": "~/Documents/papers"
+                "dir": os.path.join(platformdirs.user_documents_dir(), "papers")
             },
             GENERAL_SETTINGS_NAME: {
                 "default-library": "papers"
@@ -77,10 +78,32 @@ class Configuration(configparser.ConfigParser):
                 self[section][field] = self.default_info[section][field]
 
     def initialize(self) -> None:
+        deprecated_config = _get_deprecated_config_folder()
+        has_deprecated_config = (
+            self.dir_location != deprecated_config
+            and os.path.exists(deprecated_config))
+
         # ensure all configuration directories exist
-        if not os.path.exists(self.dir_location):
-            self.include_defaults()
-            return
+        if os.path.exists(self.dir_location):
+            if has_deprecated_config:
+                click.echo(
+                    f"The configuration is loaded from '{self.dir_location}'. A "
+                    "deprecated configuration folder was found at "
+                    f"'{deprecated_config}' and skipped. Please remove it to "
+                    "avoid seeing this warning in the future.")
+        else:
+            if has_deprecated_config:
+                click.echo(
+                    "A deprecated configuration folder was found at "
+                    f"'{deprecated_config}' and has been copied to the new "
+                    f"location '{self.dir_location}'. Please remove the deprecated "
+                    "folder to avoid seeing this warning in the future.")
+
+                import shutil
+                shutil.copytree(deprecated_config, self.dir_location)
+            else:
+                self.include_defaults()
+                return
 
         # load settings
         if os.path.exists(self.file_location):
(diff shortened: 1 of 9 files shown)


alexfikl:
@kiike To get this in and make progress on #795 it would just need some testing:
* On all platforms: make sure that the configuration is correctly migrated from the old paths to the new paths. I'll try to play around with this a bit on Linux at least.
* On Windows / macOS: make sure that the new paths are where they're expected to be now.

For new users, it should hopefully just work!

papis/config.py
```diff
@@ -188,61 +208,56 @@ def register_default_settings(settings_dictionary: PapisConfigType) -> None:
...
 
-    # NOTE: If no folder is found, then get the config home
-    return os.path.join(get_config_home(), "papis")
+def get_config_folder() -> str:
+    """Get the main configuration folder.
+
+    :returns: a (platform dependent) folder where the configuration files are
+        stored, e.g. ``$HOME/.config/papis`` on POSIX platforms.
+    """
+    return platformdirs.user_config_dir("papis")
```
alexfikl, line 252: This doesn't look through `XDG_CONFIG_DIRS` anymore, so it won't pick up configurations in e.g. `/etc/xdg` on Linux. It should be easy enough to add if someone needs it though.

kiike:
> @kiike To get this in and make progress on #795, it would just need some testing of the migration path:
> 
>     * On all platforms: make sure that the configuration is correctly migrated from the old paths to the new paths. I'll try to play around with this a bit on Linux at least.
> 
>     * On Windows / macOS: make sure that the new paths are where they're expected to be now.
> 
> 
> For new users, it should hopefully just work!

I nuked my windows test library and configs. The paths are correctly created (except the papers library, i think it is by design):

- library (set, not created) at ~/Documents/papers
- config at - cache at ~/AppData/Local/papis/papis/config
- cache (after first document added) at ~/AppData/Local/papis/papis/Cache

I'll try a migration path!

kiike:
I tried the migration path on Windows:

1. Nuked config and library
2. Checked out main
3. `papis init`
4. Checked out your branch
5. `papis config`
    - it shows the message about the old configuration folder and that it is being moved to the correct location. The new location is the correct one but the files are copied, not moved
    - on the next run, it says that the old config folder was found, but the new one is loaded from the new, correct location instead (maybe it could say that it can be removed after checking there are no valuable files, or prompt removal after comparing both dirs)
    - if i remove the old deprecated path, the message no longer appears

So overall I think it works fine on Windows!

alexfikl:
> So overall I think it works fine on Windows!

That's awesome! Thank you very much for testing this! :heart:

alexfikl:
> it shows the message about the old configuration folder and that it is being moved to the correct location. The new location is the correct one but the files are copied, not moved

Yeah, I thought copying instead of moving them was better to avoid any data loss. Would you expected to be moved? (Just noticed that the log message is out of sync, my bad!)

kiike:
> > it shows the message about the old configuration folder and that it is being moved to the correct location. The new location is the correct one but the files are copied, not moved
> 
> Yeah, I thought copying instead of moving them was better to avoid any data loss. Would you expected it to be moved? (Just noticed that the log message is out of sync, my bad!)

I would expect:
- copying, then checking everything went fine, then deleting
OR
- copying, and displaying this message, but inviting the user to delete the old deprecated directory to avoid seeing the message. Like:

> Papis is now loading the configuration from the new default location $LOCATION, but the deprecated folder in $DEPRECATED_LOCATION still exists. Please check there's nothing of value in the deprecated folder and delete it to avoid seeing this warning in the future.

alexfikl:
> I would expect:

That sounds a lot better! I left it at just copying and updated the logging messages to be more in line with what you suggested.

alexfikl:
This seems to be hitting some sort of race condition on Windows when using `multiprocessing`: for some reason it seems that `platformdirs` and the old `XDG_CONFIG_HOME` return different values and the config directory creation gets confused..

alexfikl:
I'll go ahead and merge this, since it would be nice to have in the next release. 

We've tested it a bit and it seems to work reasonably well, so fingers crossed :rocket:

PR #748, opened 2024-02-12
docker: fix image

alexfikl:
@jghauser Can you check if this works? I apparently managed to break it in e46fc5888ad953ad890d7eee9357e697309f742f because I don't understand how Docker files work. Sorry about that! :cry:


jghauser:
Perfect, this fixes it! Thanks a lot! :tada:

PR #752, opened 2024-02-15
docs: improve how configuration values are handled

alexfikl:
This improves how configuration values are handled in the docs:
* It adds a new directive `:confval:` that's used by the wrapping `papis-config` directive to render the configuration values in a nicer way.
* Adds a `type: ...` entry to each config to show what type we expect it to be.
* They can now be referenced normally using `:confval:<name>` just like `:class:` or `:func:` roles.

It also moves the Sphinx directives from `docs/source/conf.py` to `papis/sphinx_ext.py` so that they can be used by external plugins without copy-pasting them. And for some pretty pictures before and after:

Before:
</img>

After:
</img>


alexfikl:
I'll go ahead and merge this, but I'm happy to improve it in another PR if anyone notices anything off! :rocket:

jghauser:
looks great!

alexfikl:
> looks great!

Thank you! And thanks for taking a look :grin:

PR #756, opened 2024-02-17
add convenience commands to nix flake

jghauser:
I've been wanting a container in which I can try out various commands and see how they work. Which is why I made the below small script number 3 for my development shell. While I was doing that, it seemed logical to also create 1 + 2:

1. `papis-build-container` (builds the container)
2. `papis-run-container-tests` (runs the tests in the container)
3. `papis-run-container-interactive` (drops you into a shell in the container after populating the library)

The commands use `podman` or `docker`, depending on which is in the path.

I could also make these individual scripts so that they would be available for non-nix users, if there is a desire for that.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -35,6 +35,7 @@ The docs can be generated with
 ```
 make doc
 ```
+### Containers
 
 To quickly get things up and running, you can also use docker/podman:
 ```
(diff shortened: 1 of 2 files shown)


alexfikl:
I don't really use Docker for development, so if you think these are useful, go ahead!

jghauser:
Ok great, thanks for having a look. I've added some documentation to HACKING.md. Merging once CI completes :rocket:

PR #758, opened 2024-02-22
tests: add a fixture for doctests

### LINKED ISSUE #757: test failing when using jinja2 formatter
https://github.com/papis/papis/blob/84a90d19b17d26616a8e78b31a8b396869b0d1ba/papis/database/cache.py#L103C1-L104C10

This is the exact line that is failing. I am using jinja2 formatter. I also noticed this format works fine with my configuration

```
>>> match_document(document, regex('e in'), '{{ doc.author }}') is None
False
```

alexfikl:
This adds a `conftest.py` with a little fixture to the root `papis` folder. It's meant to run the doctests in a clean configuration (not the one in the user's `~/.config/papis`)

@alejandrogallo It's a little hacky to add a test file to the package folder, but it should help users run and pass the tests (if not using a docker container or similar). What do you think?

Fixes #757.

### DIFF
--- a/papis/testing.py
+++ b/papis/testing.py
@@ -9,6 +9,7 @@ import click
 import click.testing
 from _pytest.fixtures import SubRequest
 from _pytest.config import Config
+from _pytest.config.argparsing import Parser
 
 PAPIS_UPDATE_RESOURCES = os.environ.get("PAPIS_UPDATE_RESOURCES", "none").lower()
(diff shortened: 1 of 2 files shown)


alexfikl:
Hacked this in a nicer way by adding it to `papis.testing` and adding a little command-line flag to enable it.

PR #760, opened 2024-02-22
makefile: update nix command

alexfikl:
@jghauser I tried running some of those nix commands on just a bare nix install on Archlinux and they complained about some missing `--extra-experimental-features`. Is it ok to add those like this?


jghauser:
That makes sense, nix flakes and the new nix command are still experimental (though by now very widely used) features. Most nixos users will have the flags set globally, but by default this isn't the case. So yes, it makes sense to add them like this to the Makefile I think!

alexfikl:
Thank you for the explanation!

PR #767, opened 2024-03-07
refactor(tests/test_config): check instance instead of comparing

kiike:
I saw a very minor improvement that ruff caught:

There was a comparison that was performed instead of checking the instance.

This change makes ruff happier.

Also, this is my first of many (I hope!) commits to papis, so :wave: !

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -93,12 +93,16 @@ def test_get(tmp_config: TemporaryConfiguration) -> None:
     papis.config.set("test_getint", "42")
     assert papis.config.getint("test_getint") == 42
     assert papis.config.getint("test_getint", section=general_name) == 42
-    assert type(papis.config.getint("test_getint", section=general_name)) is int
+    assert isinstance(papis.config.getint("test_getint", section=general_name),
+                      int
+                      )
 
     papis.config.set("test_getfloat", "3.14")
     assert papis.config.getfloat("test_getfloat") == 3.14
     assert papis.config.getfloat("test_getfloat", section=general_name) == 3.14
-    assert type(papis.config.getfloat("test_getfloat", section=general_name)) is float
+    assert isinstance(papis.config.getfloat("test_getfloat", section=general_name),
+                      float
+                      )
 
     papis.config.set("test_getbool", "True")
     assert papis.config.getboolean("test_getbool") is True


alexfikl:
We can merge this, but we don't use `ruff` for linting, so I imagine there's a lot of little discrepancies here and there.

kiike:
No problem. This is just nitpicking by a tool that it is not used yet by papis.

I used ruff because I was trying out your pyproject.toml branch, because it turns out that VS Code fails to discover the tests with the current setup.py of papis. But, with pyproject.toml, which I adapted to use rye, it works fine. And rye lints with ruff by default, that's why I saw this "issue".

I also learnt isinstance as more idiomatic, and pylint, the tool that papis uses, [will probably adopt this check too](https://pylint.pycqa.org/en/latest/user_guide/messages/convention/unidiomatic-typecheck.html) as apparently it is more idiomatic.

Thanks for the quick review!

alexfikl:
> I used ruff because I was trying out your pyproject.toml branch, because it turns out that VS Code fails to discover the tests with the current setup.py of papis. But, with pyproject.toml, which I adapted to use rye, it works fine. And rye lints with ruff by default, that's why I saw this "issue".

`rye` seems to be pretty dead at this point, since it was mostly swallowed by [uv](https://github.com/astral-sh/uv), so we probably won't be switching to it even when the `pyproject` stuff gets in. That being said, `ruff` is definitely on the roadmap (hopefully after the next release)!

> I also learnt isinstance as more idiomatic, and pylint, the tool that papis uses, [will probably adopt this check too](https://pylint.pycqa.org/en/latest/user_guide/messages/convention/unidiomatic-typecheck.html) as apparently it is more idiomatic.

That's a pretty opinionated lint in general because `isinstance(obj, cls)` and `type(obj) is cls` are not the same check -- `isinstance` also accepts subclasses. It seems fair to run to spot potential bugs, but much more.

kiike:
> > I used ruff because I was trying out your pyproject.toml branch, because it turns out that VS Code fails to discover the tests with the current setup.py of papis. But, with pyproject.toml, which I adapted to use rye, it works fine. And rye lints with ruff by default, that's why I saw this "issue".
> 
> `rye` seems to be pretty dead at this point, since it was mostly swallowed by [uv](https://github.com/astral-sh/uv), so we probably won't be switching to it even when the `pyproject` stuff gets in. That being said, `ruff` is definitely on the roadmap (hopefully after the next release)!
> 
> > I also learnt isinstance as more idiomatic, and pylint, the tool that papis uses, [will probably adopt this check too](https://pylint.pycqa.org/en/latest/user_guide/messages/convention/unidiomatic-typecheck.html) as apparently it is more idiomatic.
> 
> That's a pretty opinionated lint in general because `isinstance(obj, cls)` and `type(obj) is cls` are not the same check -- `isinstance` also accepts subclasses. It seems fair to run to spot potential bugs, but not much more.

Thanks for the feedback! I agree on the opinionated lint! Sorry if the PR was out of scope then!

About rye, as far as I understand the dependency installing/solving/syncing is now handled by uv instead of using pip-tools like it was in the past. The other things rye does, like publishing, project initialization, workspace management  I think uv is not capable of doing them yet. It might be a matter of time that rye is ingested into uv maybe?

I am using rye with the uv backend to install the dependencies of papis and for other projects too and I really like the experience. I do miss the ephemeral environments of poetry, but not all packages were installable by poetry.

Which one do you prefer for papis?

alexfikl:
> About rye, as far as I understand the dependency installing/solving/syncing is now handled by uv instead of using pip-tools like it was in the past. The other things rye does, like publishing, project initialization, workspace management I think uv is not capable of doing them yet. It might be a matter of time that rye is ingested into uv maybe?

I was mostly referring to this https://astral.sh/blog/uv. It seems like rye is on maintenance mode and will slowly be incorporated into uv. It's all a bit new..
 
> I am using rye with the uv backend to install the dependencies of papis and for other projects too and I really like the experience. I do miss the ephemeral environments of poetry, but not all packages were installable by poetry.

Out of curiosity, what packages weren't installable by poetry?

> Which one do you prefer for papis?

At the moment I'm still using just plain ol' `setuptools`. Haven't seen a good reason to move to anything fancier.

kiike:
> > About rye, as far as I understand the dependency installing/solving/syncing is now handled by uv instead of using pip-tools like it was in the past. The other things rye does, like publishing, project initialization, workspace management I think uv is not capable of doing them yet. It might be a matter of time that rye is ingested into uv maybe?
> 
> I was mostly referring to this https://astral.sh/blog/uv. It seems like rye is on maintenance mode and will slowly be incorporated into uv. It's all a bit new..

I read that too when it came out, but I don't understand how that might imply rye is on maintenance mode. They say that they are using rye as testbed and incorporating uv functionality into it. To me, that is the opposite of maintenance mode, but it will be very interesting to see how this turns out!


> Out of curiosity, what packages weren't installable by poetry?

NLP libraries, mostly, like huggingface dependencies such as datasets or transformers. That was last august. Might have changed, but I haven't gone back to poetry yet.

 
> At the moment I'm still using just plain ol' `setuptools`. Haven't seen a good reason to move to anything fancier.

Fair enough! But would you like to use other tools to provide functionality on top, such as linting, version bumping, etc. such as pdm/poetry/hatch?

alexfikl:
> I read that too when it came out, but I don't understand how that might imply rye is on maintenance mode. They say that they are using rye as testbed and incorporating uv functionality into it. To me, that is the opposite of maintenance mode, but it will be very interesting to see how this turns out!

Armin seems to be clearer on those aspects https://lucumr.pocoo.org/2024/2/15/rye-grows-with-uv/. Maybe maintenance is not well put, but it doesn't seem to be meant as a longer term solution.
 
> NLP libraries, mostly, like huggingface dependencies such as datasets or transformers. That was last august. Might have changed, but I haven't gone back to poetry yet.

Ah, fair enough, those are always painful for python packaging from whichever side :(

> Fair enough! But would you like to use other tools to provide functionality on top, such as linting, version bumping, etc. such as pdm/poetry/hatch?

Haven't had much friction there yet, but I also don't maintain anything too big or too strict in that regard.
* For linting, doing a `ruff check` instead of `rye lint` doesn't seem like a big deal.
* For version bumps, we can probably improve in papis, but just setting the version in `pyproject.toml` and using `importlib.metadata` seems to be pretty straightforward.

Having all these integrated under one umbrella is probably nice though.

kiike:
> Having all these integrated under one umbrella is probably nice though. If `uv` is going to be as nice and fast as `ruff`, we should probably move to it 😁

Right now, it is! Have you tried it? I am using it to create environments on machines I'd rather have idle as short as possible and in that regard it is really helping!

Issue #768, opened 2024-03-07
Add support for Zenodo


kiike:
This is similar to Zenodo, which also has an API we can use
https://developers.zenodo.org/

With these two, we should have pretty good support for citing software as well!

_Originally posted by @alexfikl in https://github.com/papis/papis/issues/720#issuecomment-1872943423_
            
I am working on this. This will be a great opportunity to get more acquainted with the papis API.

PR #770, opened 2024-03-09
Add a prototype importer for zenodo

kiike:
As mentioned in #768, I am working on the Zenodo importing. I need to cite a couple datasets I use for my thesis and while I already have them as imported by papis-zotero, I want to contribute and learn more about papis.

This pull request implements a first version of a zenodo importer.

### Features

- [x] Fetch records metadata 
- [x] Download and import the associated files
- [ ] Search the zenodo database
- [ ] Explore the zenodo citations for a record

I think it's ready to be tested by someone who uses zenodo. I tested with a couple of records and I can retrieve and save the metadata, along with the files:

```json
{
            "name": "Python Debugger: papis zenodo",
            "type": "debugpy",
            "request": "launch",
            "program": "${workspaceFolder}/.venv/bin/papis",
            "console": "integratedTerminal",
            "args": "add --file-name '' --from zenodo 10794563",
}
```

result:

```yaml
author: Bourg, Salomé and Bolstad, Geir H. and Griffin, Donald V. and Pélabon, Christophe
  and Hansen, Thomas F.
author_list:
- affiliation: Norwegian University of Science and Technology
  family: Bourg
  given: Salomé
- affiliation: Norwegian Institute for Nature Research
  family: Bolstad
  given: Geir H.
- affiliation: Florida State University
  family: Griffin
  given: Donald V.
- affiliation: Norwegian University of Science and Technology
  family: Pélabon
  given: Christophe
- affiliation: University of Oslo
  family: Hansen
  given: Thomas F.
concept_record_id: '10794562'
day: 7
description: Epistasis is often portrayed as unimportant in evolution. While random
  patterns of epistasis may have limited effects on the response to selection, systematic
  directional epistasis can have substantial effects on evolutionary dynamics. Directional
  epistasis occurs when allelic substitutions that change a trait also modify the
  effects of allelic substitutions at other loci in a systematic direction. In this
  case, trait evolution may induce correlated changes in allelic effects and effective
  genetic variance (evolvability) that modify further evolution. Although theory thus
  suggests a potentially important role for directional epistasis in evolution, we
  still lack empirical evidence about its prevalence and magnitude. Using a new framework
  to estimate systematic patterns of epistasis from line-crosses experiments, we quantify
  its effects on 197 size-related traits from diverging natural populations in 24
  animal and 17 plant species. We show that directional epistasis is common and tends
  to become stronger with increasing morphological divergence. In animals, most traits
  displayed negative directionality toward larger size, suggesting that epistasis
  constraints reducing evolvability toward larger size may be common. Dominance was
  also common and did not systematically alter the effects of epistasis.
doi: 10.5061/dryad.66t1g1k7m
files:
- paperlist-epistasis.xls
- full-list-of-studies-epistasis-a.xls
- dataset-epistasis-b.xls
- readme-c.md
- reciprocal-line-cross-explanation-d.pdf
- aics-epistasis-e.xls
license: cc-zero
method: The data were collected through a literature search.
month: 3
notes: "Funding provided by: Norwegian research council (RCN)\\*  \nCrossref Funder
  Registry ID:   \nAward Number: 287214\n\nFunding provided by: Centre for Advanced
  Study  \nCrossref Funder Registry ID: https://ror.org/05rbhea42  \nAward Number:
  223257"
papis_id: c4b631820231b36b03b8d59ee33fae0f
ref: Directional_epi_Bourg_2024
revision: 5
tags:
- linecross
- Epistasis
- Genetic dominance
- evolvability
- Evolutionary biology
time-added: 2024-03-09-11:01:31
title: Directional epistasis is common in morphological divergence
type: dataset
url: https://zenodo.org/api/records/10794563
year: 2024
zenodo_record_id: 10794563
```


tests/test_zenodo.py
```diff
@@ -0,0 +1,37 @@
...
+        result = json.load(fd)
+
+    assert isinstance(result, dict)
+    return result
+
+
+@pytest.mark.parametrize(("zenodo_id",), [("7391177",), ("10794563",)])
+def test_zenodo_id_to_data(monkeypatch: Any, zenodo_id: str) -> None:
+    mock_data_file = "input_{}.json".format(zenodo_id)
+    expected_data_file = "output_{}.json".format(zenodo_id)
```
alexfikl, line 26: Small nitpick here: most of the test resources are named `IDENTIFIER.EXT` and `IDENTIFIER_out.EXT`. These should be named the same for consistency.

papis/zenodo.py
```diff
@@ -0,0 +1,221 @@
...
+            },
+            {"key": "day", "action": lambda x: datetime.datetime.fromisoformat(x).day},
+        ],
+    ),
+    KeyConversionPair("creators", [{"key": "author_list", "action": get_author_info}]),
+    KeyConversionPair("links", [{"key": "url", "action": lambda x: x["self"]}]),
+    KeyConversionPair("notes", [{"key": "notes", "action": get_text_from_html}]),
+    KeyConversionPair("method", [{"key": "method", "action": get_text_from_html}]),
+    KeyConversionPair(
+        "resource_type", [{"key": "type", "action": lambda x: x["type"]}]
```
alexfikl, line 93: Do these resource types match BibTeX types in any way?

papis/zenodo.py
```diff
@@ -0,0 +1,221 @@
...
+    :type data: Dict[str, Any]
+    :return: a new dictionary in a conventional papis form
+    :rtype: Dict[str, Any]
+    """
+    # Merge metadata into data
+    data.update(data["metadata"])
+    del data["metadata"]
+
+    new_data = papis.document.keyconversion_to_data(key_conversion, data)
+    new_data["file_urls"] = data["files"]
```
alexfikl, line 116: "file_urls"?

papis/zenodo.py
```diff
@@ -0,0 +1,221 @@
...
+    checksum: str
+    links: LinkInfo
+
+
+class Importer(papis.importer.Importer):
+    """Importer downloading data from a Zenodo ID"""
+
+    def __init__(self, uri: str = "") -> None:
+        super().__init__(name="zenodo", uri=uri)
+        self._file_urls: List[FileInfo] = []
```
alexfikl, line 193: Hm, this is a bit ugly. We should probably improve the API for these importers to let them cache things in a nicer way..

papis/zenodo.py
```diff
@@ -0,0 +1,221 @@
...
+    - checksum: md5sum hash of the file.
+    - links: dict with the url to download the file
+    """
+
+    id: str
+    key: str
+    size: int
+    checksum: str
+    links: LinkInfo
+
```
alexfikl, line 186: This seems a bit overkill. Why not just save the URL in the first place?

papis/zenodo.py
```diff
@@ -0,0 +1,221 @@
...
+    def fetch_files(self) -> None:
+        if not self.ctx.data:
+            return
+
+        for url_data in self._file_urls:
+            url = url_data["links"]["self"]
+            self.logger.info("Trying to download document from '%s'.", url)
+
+            filename = download_document(url, filename=url_data["key"])
```
alexfikl, line 219: These files will be renamed in `papis add` or `papis update` using `add-file-name` anyway, so I'm not sure it's worth giving them a fixed file name here.

Most of our importers just get the one file, so it's not confusing. I imagine Zenodo downloads more files, so we might need to rework that..

alexfikl (commented):
This generally looks good to me!

Zenodo is sort of the first of its kind that gets added here, so we probably need to think a bit about it. From looking over the code, at least two things come to mind:
* Papis is fairly BibTeX-centric, but this adds a lot of new fields in there that we don't know about (e.g. revision?). Are there any BibTeX extensions that work with that? `biblatex-software` comes to mind.
* Zenodo provides a bunch of files with which we probably shouldn't mess around with. Right now, these will just get renamed `something-a.zip`, `something-b.txt`, etc. which is very unhelpful..

Left a few random comments as well, but the code is in nice shape :grin:

kiike:
> This generally looks good to me!

Thanks!

> Zenodo is sort of the first of its kind that gets added here, so we probably need to think a bit about it. From looking over the code, at least two things come to mind:
> 
>     * Papis is fairly BibTeX-centric, but this adds a lot of new fields in there that we don't know about (e.g. revision?). Are there any BibTeX extensions that work with that? `biblatex-software` comes to mind.

I see. I moved to hayagriva since I submit my deliverables with Typst. Admittedly, I only cite articles, so I don't know the rules of software. I'll need to have a look and adapt the keys, then.

>     * Zenodo provides a bunch of files with which we probably shouldn't mess around with. Right now, these will just get renamed `something-a.zip`, `something-b.txt`, etc. which is very unhelpful..

Indeed. What about letting the user provide a variable for `original_filename` so you can have all the attached files renamed as, say `JohnDoe2024-foobar-v1.0.zip`, `JohnDoe2023-foobar-article.pdf?`

> Left a few random comments as well, but the code is in nice shape 😁

I am really thankful for the thorough review. It shows that you care for the project!

kiike:
@alexfikl, have a look at it if you wish. I think I addressed your feedback in the recent commits.

kiike:
I haven't found an elegant way to preserve filenames, so we'll probably have to think about that. I'll create a discussion thread.

papis/zenodo.py
```diff
@@ -0,0 +1,208 @@
...
+
+def zenodo_data_to_papis_data(data: Dict[str, Any]) -> Dict[str, Any]:
+    """Converts the dictionary from Zenodo to the conventional papis format.
+
+    :param data: the raw dictionary from Zenodo
+    :return: a new dictionary in a conventional papis form
+    """
+    # Merge metadata into data
+    data.update(data["metadata"])
+    del data["metadata"]
```
alexfikl, line 123:
```suggestion
    data.update(data.pop("metadata", {}))
```
?

papis/zenodo.py
```diff
@@ -0,0 +1,208 @@
...
+
+    new_data = papis.document.keyconversion_to_data(key_conversion, data)
+    return new_data
```
alexfikl, line 126:
```suggestion
    return papis.document.keyconversion_to_data(key_conversion, data)
```

papis/zenodo.py
```diff
@@ -0,0 +1,208 @@
...
+    def match(cls, uri: str) -> Optional[papis.importer.Importer]:
+        if is_valid_record_id(uri):
+            return Importer(uri)
+
+        return None
+
+    @classmethod
+    def match_data(cls, data: Dict[str, Any]) -> Optional["Importer"]:
+        return None
+
```
alexfikl, line 190:
```suggestion
```

papis/zenodo.py
```diff
@@ -0,0 +1,208 @@
...
+    def fetch_data(self) -> None:
+        zenodo_data = get_data(self.uri)
+        # Build a filename to URL dictionary
+        self.ctx.file_info = {
+            file["key"]: file["links"]["self"] for file in zenodo_data["files"]
+        }
```
alexfikl, line 196:
```suggestion
        self.ctx.file_info = [
            (file["key"], file["links"]["self"]) for file in zenodo_data["files"]
        ]
```
? Not sure there's any difference, so up to you!

papis/zenodo.py
```diff
@@ -0,0 +1,208 @@
...
+        self.ctx.data = zenodo_data_to_papis_data(zenodo_data)
+
+    def fetch_files(self) -> None:
+        if not self.ctx.file_info:
+            return
+
+        for filename, url in self.ctx.file_info.items():
```
alexfikl, line 203:
```suggestion
        for filename, url in self.ctx.file_info:
```
If you think the change above is a good idea.

alexfikl (commented):
Left a bunch of nitpicks, but this all looks good if the file naming thing is fixed.

> I haven't found an elegant way to preserve filenames, so we'll probably have to think about that. Maybe we can discuss this in the forum? Or here?

Have you found a non-elegant way? :grin:
Right now, papis should leave the file names alone if you don't set `add-file-name`. However, if it is set, not quite sure what's a good way to tell it "leave *THESE* files alone and rename the rest"..

kiike:
> Have you found a non-elegant way? 😁 Right now, papis should leave the file names alone if you don't set `add-file-name`. However, if it is set, not quite sure what's a good way to tell it "leave _THESE_ files alone and rename the rest"..
> 
> EDIT: Also, we can definitely merge this even if we don't have an "elegant" way to do it just yet. It would just need a bit of documentation somewhere to explain any shortcomings (or a `logger.warning`). If it "works for you" in a reasonable way, we can move ahead.

It doesn't rename the files if `add-file-name` or `filename` are set but the downloader definitely does, since [there isn't (or I can't realise)](https://github.com/papis/papis/blob/ee68780e4b963364238fae83136cbc38133f5b44/papis/downloaders/__init__.py#L475) a code path to save to a *named* file.

alexfikl:
> It doesn't rename the files if `add-file-name` or `filename` are set but the downloader definitely does, since [there isn't (or I can't realise)](https://github.com/papis/papis/blob/ee68780e4b963364238fae83136cbc38133f5b44/papis/downloaders/__init__.py#L475) a code path to save to a _named_ file.

Ah, sorry! I meant with the `filename` parameter you added in this PR. Does it work as expected with that if `add-file-name` is not set?

kiike:
> > It doesn't rename the files if `add-file-name` or `filename` are set but the downloader definitely does, since [there isn't (or I can't realise)](https://github.com/papis/papis/blob/ee68780e4b963364238fae83136cbc38133f5b44/papis/downloaders/__init__.py#L475) a code path to save to a _named_ file.
> 
> Ah, sorry! I meant with the `filename` parameter you added in this PR. Does it work as expected with that if `add-file-name` is not set?
> 
> EDIT: For context, that `download_document` function was written with the "well, papis add will rename this anyway" mentality, which is definitely incorrect if `add-file-name` is not set :\

Right now, as it is, it is the choice of the user, since for some cases it might make sense to rename (you are importing a bunch of files) or not (just an article). Your comment on the warning makes a lot of sense for this use case, but the importer cannot now if you called papis with `file-name` set. Maybe I can document this so that the user can take this into account. Where would be a good place to add this documentation?

alexfikl:
> Right now, as it is, it is the choice of the user, since for some cases it might make sense to rename (you are importing a bunch of files) or not (just an article). Your comment on the warning makes a lot of sense for this use case, but the importer cannot now if you called papis with `file-name` set. Maybe I can document this so that the user can take this into account. Where would be a good place to add this documentation?

Hm, `papis add` already says how it renamed files
https://github.com/papis/papis/blob/ee68780e4b963364238fae83136cbc38133f5b44/papis/commands/add.py#L417-L426
Would that be sufficient? i.e. we can just rename those `logger.debug` into `logger.info`.

kiike:
> Would that be sufficient? i.e. we can just rename those `logger.debug` into `logger.info`.

I think that would be a bit better! Right now the program is pretty verbose, so for more important stuff like this (where papis saves documents) I think it's a good change!

[Commited](https://github.com/papis/papis/pull/770/commits/cf9a3549419ec2a2bbb47fb905e702c6b6a843b4)!

alexfikl:
> I think that would be a bit better! Right now the program is pretty verbose, so for more important stuff like this (where papis saves documents) I think it's a good change!
> 
> [Commited](https://github.com/papis/papis/pull/770/commits/cf9a3549419ec2a2bbb47fb905e702c6b6a843b4)!

This is good to go from my side, so just let me know when you think it's ready to merge :rocket: 

Also, feel free to open some issues for any renaming problems that you came across! 
Especially that file renaming thing needs fixing in a more general way so that all importers keep the name they downloaded unless the user explicitly asks for them to be renamed.

kiike:
> This is good to go from my side, so just let me know when you think it's ready to merge 🚀

For the two more important functionalities, ie. downloading metadata and the files, it's functional and good to go, being able to be improved incrementally.

> Also, feel free to open some issues for any renaming problems that you came across! Especially that file renaming thing needs fixing in a more general way so that all importers keep the name they downloaded unless the user explicitly asks for them to be renamed.

That's a good idea. Since I am using papis now, I'll probably come across something that I can improve!

alexfikl:
Thanks for working on this! :rocket:

PR #773, opened 2024-03-11
zenodo: remove markdownify

alexfikl:
@kiike I noticed that that `markdownify` usage was still in there, so this removes it and uses `bs4` to remove the HTML instead (one less dependency).

Are Zenodo descriptions always in markdown format? Would this be acceptable?


kiike:
Hi @alexfikl! Apparently (i used Zenodo occasionally) they are formatted in rich text that is stored in HTML in the JSON exports we are getting.

I tried using `get_text` from bs4 but IMHO it looks worse than the HTML -> Markdown when looking at the yaml. Sure, bold text or italics gets converted to plain text which is fine, but bullet lists or links get lost.

But go ahead with the change if you think it's better than the optional dependency :smiley:!

PR #775, opened 2024-03-11
Add zenodo entrypoint

alexfikl:
@kiike Tried to play around a bit with the Zenodo importer and noticed it didn't have an entrypoint set. How did you test it?

Also found a little bug: the filename from Zenodo could be a relative path e.g. `username/filename.zip`, in which case just writing the file wouldn't work because the folder does not exist. I modified it to just use the basename instead.

### DIFF
--- a/papis/downloaders/__init__.py
+++ b/papis/downloaders/__init__.py
@@ -1,4 +1,5 @@
 import re
+import os
 import tempfile
 from typing import List, Optional, Any, Sequence, Type, Dict, Union, TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


kiike:
> @kiike Tried to play around a bit with the Zenodo importer and noticed it didn't have an entrypoint set. How did you test it?

Thanks for trying it out! OK, so I just realised something stupid: I had the `pyproject.toml` lying around, unchecked in git because I had to copy it from your branch so that VS Code could work better with the code. The entrypoint was there but not in setup.py. And the tests work because they import directly, i think?

So thanks for adding the entrypoint to setup.py! And sorry!

> Also found a little bug: the filename from Zenodo could be a relative path e.g. `username/filename.zip`, in which case just writing the file wouldn't work because the folder does not exist. I modified it to just use the basename instead.

The incoming filename could be a path? Interesting! I didn't find this case in my limited research. Nice catch!

alexfikl:
No worries! And thank you for the quick responses!

FYI: I just randomly tried a project I worked on and it failed the download, so I guess I'm just lucky :D
https://zenodo.org/records/8397396

kiike:
Lucky you! I tried with two projects which worked. Maybe it has to do with slashes in the filename? I can work on it!

PR #783, opened 2024-03-16
improve flake

jghauser:
After the move to `hatchling`, I had a look at the flake. I actually wanted to see if I could make use of [poetry2nix](https://github.com/nix-community/poetry2nix/) to simplify things, but as the name of the tool sort of implies, it requires `poetry` :sweat_smile:. I did still find various things to improve though:

- Depend on unstable `nixpkgs`. I think this makes more sense for a flake that's tracking `main`. It also means we don't have to update the flake whenever nix releases a new version.
- Updated the `arxiv.py` version to the latest release.
- Defined `hatchling` as a build dependency (rather than a standard dep).
- Have the version name be the git rev, rather than the same as the release (more appropriate given that it's tracking main)
- Use `pyproject = true` as that's what's [used nowadays](https://nixos.org/manual/nixpkgs/stable/#reference)

### DIFF
--- a/.github/workflows/nix-flake.yml
+++ b/.github/workflows/nix-flake.yml
@@ -23,4 +23,4 @@ jobs:
     - uses: DeterminateSystems/flake-checker-action@main
     - run: nix build
     - run: nix flake check
-    - run: nix develop --command bash -c "source .venv/bin/activate; python -m pytest -v papis tests"
+    - run: nix develop --command bash -c "pytest -v papis tests"
--- a/flake.lock
+++ b/flake.lock
@@ -18,26 +18,95 @@
         "type": "github"
       }
     },
+    "mdbook-nixdoc": {
+      "inputs": {
+        "nix-github-actions": [
+          "pyproject-nix",
+          "nix-github-actions"
+        ],
+        "nixpkgs": [
+          "pyproject-nix",
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1708395692,
+        "narHash": "sha256-smf0VmxGbjJDZqKvxxG3ZVqubgbVwAWG26wPo+BT/A0=",
+        "owner": "adisbladis",
+        "repo": "mdbook-nixdoc",
+        "rev": "d6a71b114b9221c0b4f20d31b81766d072cc26be",
+        "type": "github"
+      },
+      "original": {
+        "owner": "adisbladis",
+        "repo": "mdbook-nixdoc",
+        "type": "github"
+      }
+    },
+    "nix-github-actions": {
+      "inputs": {
+        "nixpkgs": [
+          "pyproject-nix",
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1703863825,
+        "narHash": "sha256-rXwqjtwiGKJheXB43ybM8NwWB8rO2dSRrEqes0S7F5Y=",
+        "owner": "nix-community",
+        "repo": "nix-github-actions",
+        "rev": "5163432afc817cf8bd1f031418d1869e4c9d5547",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-community",
+        "repo": "nix-github-actions",
+        "type": "github"
+      }
+    },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1707514827,
-        "narHash": "sha256-Y+wqFkvikpE1epCx57PsGw+M1hX5aY5q/xgk+ebDwxI=",
+        "lastModified": 1710451336,
+        "narHash": "sha256-pP86Pcfu3BrAvRO7R64x7hs+GaQrjFes+mEPowCfkxY=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "20f65b86b6485decb43c5498780c223571dd56ef",
+        "rev": "d691274a972b3165335d261cc4671335f5c67de9",
         "type": "github"
       },
       "original": {
         "owner": "NixOS",
-        "ref": "nixos-23.11",
+        "ref": "nixos-unstable",
         "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "pyproject-nix": {
+      "inputs": {
+        "mdbook-nixdoc": "mdbook-nixdoc",
+        "nix-github-actions": "nix-github-actions",
+        "nixpkgs": [
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1709886419,
+        "narHash": "sha256-qXZsmegy0W0hn9yxYQFin1jZB3ikmBWXyBDQx1AqEQ4=",
+        "owner": "nix-community",
+        "repo": "pyproject.nix",
+        "rev": "2841577401c77106150b98bcb3a30510cb6c652a",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-community",
+        "repo": "pyproject.nix",
         "type": "github"
       }
     },
     "root": {
       "inputs": {
         "flake-utils": "flake-utils",
-        "nixpkgs": "nixpkgs"
+        "nixpkgs": "nixpkgs",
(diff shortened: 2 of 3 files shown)


alexfikl:
Have you looked at [pyproject.nix](https://github.com/nix-community/pyproject.nix)? Just found it from a quick search..

Besides that, everything looks good to me! Feel free to merge whenever you think it's ready :grin:

jghauser:
Huh, how did that escape my notice? :sweat_smile: i'll have a look and see if things could be further improved. thanks!

jghauser:
Thanks for the pointer, `pyproject.nix` was indeed exactly what we needed. I've now changed the flake to use that, which means all the metadata from the `pyproject.toml` can be used automatically. Then, while I was at it, I added all the missing dev dependencies, so that the nix devshell is now complete.

The only problem is that `make pytest` fails for many tests (outside the container). I'm gonna have to investigate. The nix build test in our CI is also failing, though that's (additionally) because the command there uses venv whereas `pyproject.nix` uses something nix native (I think :sweat_smile:).

alexfikl:
Hm, that's a very weird assortment of errors. Can't say I'm seeing any patter as to why those are failing..

For the nix CI failures, it keeps saying `importlib.metadata.PackageNotFoundError: No package metadata was found for papis`, which makes me think that the package wasn't installed? i.e. there's a missing `pip install --editable .` or something like that?

jghauser:
Thanks for taking a look!

> One thing that may be worth trying is running pytest -v -s to get more information?

```
❯ pytest -v -s
=============================== test session starts ===============================
platform linux -- Python 3.11.8, pytest-7.4.4, pluggy-1.4.0 -- /nix/store/986kcap99fczq6dqicj905qlxvmlz9mx-python3-3.11.8-env/bin/python3.11
cachedir: .pytest_cache
rootdir: /home/julian/Documents/coding/papis
configfile: pyproject.toml
plugins: papis-0.13, cov-4.1.0
collecting 236 items / 272 errors                                                 
QGIS test script has failed.
Error message: No module named 'osgeo'
```

No idea where that is coming from, Something is definitely up haha.

> For the nix CI failures, it keeps saying `importlib.metadata.PackageNotFoundError: No package metadata was found for papis`, which makes me think that the package wasn't installed? i.e. there's a missing `pip install --editable .` or something like that?

The reason for this is that the install isn't into a venv, but rather uses nix's built-in generic devshell functionality (I think). So `source .venv bin/bin/activate` would fail, and `pytest -v papis tests` should be enough to run the tests (though some as mentioned currently fail). I also just noticed the editable install didn't work, but I managed to solve it by setting the `PYTHONPATH` env. Even so, I still get the failing tests.

alexfikl:
> ❯ pytest -v -s                                             
> QGIS test script has failed.
> Error message: No module named 'osgeo'
> ```

Hm, that's very weird. None of our dependencies at least should be pulling that in, so really not sure. 

> The reason for this is that the install isn't into a venv, but rather uses nix's built-in generic devshell functionality (I think). So `source .venv bin/bin/activate` would fail, and `pytest -v papis tests` should be enough to run the tests (though some as mentioned currently fail). I also just noticed the editable install didn't work, but I managed to solve it by setting the `PYTHONPATH` env. Even so, I still get the failing tests.

At least then, you'll have to remove the active in there:
https://github.com/papis/papis/blob/56f237a1ab46b657b44b3fcaa617ea8556530d8d/.github/workflows/nix-flake.yml#L26

Besides that, is it installed anywhere or it just does a checkout?

We're using [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html) to get the version from the `pyproject.toml` file, but that requires either and editable install or a proper install because it needs to generate the `papis-0.13.dist-info/METADATA` file in `site-packages` somewhere. For that, you'd need to point `PYTHONPATH` to wherever that `papis-0.13.dist-info` folder is, not pwd. Have you tried that?

We're calling `importlib.metadata` in `papis/__init__.py`, so this will fail pretty much instantly if it doesn't find it :\

alexfikl:
> EDIT: Was I supposed to do `pytest -v -s papis tests`? That does give me [a bunch more info](https://gist.github.com/jghauser/a66ab9678b9c30a315b20db9dd160574). Having a look now...
> 
> EDIT2: It seems my personal config file is somehow being picked up by the tests? I just moved my config file, and now everything passes. Any idea why that might be the case?

Hm, yeah, that seems to be picking up both a different version of papis and a different config for the tests. Is it somehow installing both papis 0.13 and papis `main`?

jghauser:
> At least then, you'll have to remove the activate in there:

Ok, I changed that and the tests seem to pass now. *However*, the tests still fail when I run them locally :thinking:. 

> EDIT: I meant it seems to be running the tests with a different version of Papis, so the results don't match because we changed a bunch of formatting and things.

What makes you think that? Why would that cause the tests to pass when I remove my personal config file (or when there isn't one as in CI)?

alexfikl:
> What makes you think that? Why would that cause the tests to pass when I remove my personal config file (or when there isn't one as in CI)?

It was from
```
E       AssertionError: assert 'DnaSequencingWithChain_term' == 'DNA_sequencing_Sanger'
E         - DNA_sequencing_Sanger
E         + DnaSequencingWithChain_term
```
but that's from your jinja2 formatting, right? 

I'm not sure why it's picking that up, since the `tmp_config` and `tmp_library` really should fix that. Can you run a test and check what `XDG_CONFIG_HOME` is set to? Maybe something went wrong in monkey patching those variables..

jghauser:
> but that's from your jinja2 formatting, right?

Yes, I think so.

So, I just noticed something odd: When I run `pytest tests/commands`, then I have no failures (and the same is the case with the other files in which tests otherwise fail). It's only when I run `pytest tests` that some tests fail. 

I'll check the thing about `XDG_CONFIG_HOME` tomorrow, though the above weirdness makes me think something else is up.

jghauser:
So, I just checked `XDG_CONFIG_HOME` when running `pytest`. That seems to work just file, it's set to something like `/tmp/papis-test-j4y21goo` no matter whether I run `pytest -v -s tests` (which fails) or `pytest -v -s tests/commands` (which doesn't fail).

jghauser:
Some more test results (with finally some real progress I think :smile:):

- When I use `distrobox` to create an archlinux container with an empty home folder `distrobox create -H ~/home_arch -i archlinux`, do `sudo pacman -S python-pip`, and then do `pip install -e ".[develop,docs,optional]"` and finally run `pytest tests` everything works as expected.
- However, when I then copy my `~/.config/papis` to `~/home_arch/.config`, the tests fail. So that seems to indicate it's not the fault of my `nix develop` environment.
- Finally, when I remove `~/home_arch/.config/papis/config.py` (I have both a `config` and `config.py`), then everything starts working again. So I think somehow the monkeypatching doesn't exclude `config.py` personal configs, and this has gone under the radar because most don't use a python configuration file at all.

Any idea what might be causing this?

alexfikl:
>     * Finally, when I remove `~/home_arch/.config/papis/config.py` (I have both a `config` and `config.py`), then everything starts working again. So I think somehow the monkeypatching doesn't exclude `config.py` personal configs, and this has gone under the radar because most don't use a python configuration file at all.
> 
> 
> Any idea what might be causing this?

Ha! Yeah, that's probably it.. it didn't even cross my mind! 

In those `tmp_config` fixtures we just reload the config from `/tmp/papis-whatnot`, which works great for the static configuration because any values is had before from the user's `~/.config/papis` are just cleared. However, for `config.py`, since that's just `eval`-ed into the global scope, we can't remove whatever changes it made on the first load..

I'm not quite sure what a nice way to fix this is :\

jghauser:
Ok, thanks! I'm glad we figured out what is the fault at least and that the new flake works :grinning:. Until we find some way of fixing the issue with `config.py`, I'll just use the containerised tests. I'm gonna merge this PR if that's ok with you (given that the issue is unrelated to it) and open an issue regarding the `config.py`.

alexfikl:
Sounds good!

PR #786, opened 2024-03-17
Fix tests picking up config.py

### LINKED ISSUE #784: `config.py` gets loaded when running `pytest`
As discussed in #783, when we run `pytest`, the monkeypatching doesn't get rid of options set by the user in a `config.py` (because they are `eval`-ed into the global scope). This means that whatever options are set there may cause the tests to fail.

alexfikl:
@jghauser This should hopefully fix the issues you were seeing. I added a `~/.config/papis/config.py` with just `raise RuntimeError` and it's never loaded now. 

* The first commit adds a little hack so that we don't hit this anymore, by setting `XDG_CONFIG_HOME=/tmp` before any tests run. This is under a `--tmp-xdg-config-home` flag so that we can check if any test breaks things.
* The second commit fixes our tests so that they don't call `papis.config` before we have a chance to construct a `TemporaryConfiguration`.

The second commit should make it so that the tests still work even if we don't add the `--tmp-xdg-config-home` flag to `pyproject.toml`. Let me know if you get a chance to test it!

Fixes #784.

### DIFF
--- a/papis/database/cache.py
+++ b/papis/database/cache.py
@@ -66,8 +66,8 @@ def filter_documents(
 
     """
     papis.docmatcher.DocMatcher.set_search(search)
-    papis.docmatcher.DocMatcher.parse()
     papis.docmatcher.DocMatcher.set_matcher(match_document)
+    papis.docmatcher.DocMatcher.parse()
 
     logger.debug("Filtering %d docs (search '%s').", len(documents), search)
 
--- a/papis/docmatcher.py
+++ b/papis/docmatcher.py
@@ -88,7 +88,7 @@ class DocMatcher:
     matcher: ClassVar[Optional[MatcherCallable]] = None
     #: A format string (defaulting to :confval:`match-format`) used
     #: to match the parsed search results if no document key is present.
-    match_format: ClassVar[str] = papis.config.getstring("match-format")
+    match_format: ClassVar[str] = ""
 
     @classmethod
     def return_if_match(
@@ -135,6 +135,7 @@ class DocMatcher:
             >>> DocMatcher.search
             'author:Hummel'
         """
+
         cls.search = search
 
     @classmethod
@@ -169,7 +170,10 @@ class DocMatcher:
         """
         if search is None:
             search = cls.search
+
+        cls.match_format = papis.config.getstring("match-format")
         cls.parsed_search = parse_query(search)
+
         return cls.parsed_search
 
 
--- a/papis/testing.py
+++ b/papis/testing.py
@@ -486,7 +486,7 @@ class ResourceCache:
 
 
 @pytest.fixture(autouse=True)
-def _docconf(request: SubRequest) -> Iterator[None]:
+def _doctest_tmp_config(request: SubRequest) -> Iterator[None]:
     """A fixture for doctests to ensure that they run in a clean environment.
 
     This fixture is enabled automatically for all doctests using the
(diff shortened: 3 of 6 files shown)


jghauser:
It works perfectly though I'm not quite sure I understand how the two commits interrelate :sweat_smile:. Are you saying that given the second commit, we should be able to remove `--tmp-xdg-config-home` from `pyproject.toml` and still pass the test? If so, that doesn't work. But with that flag, everything works very nicely, so this PR seems to fix the issue I had! :heart:

alexfikl:
> we should be able to remove `--tmp-xdg-config-home` from `pyproject.toml` and still pass the test? If so, that doesn't work.

Hm, that's weird. It works locally. I added some `raise Exception` in there and it never gets called, as expected..

jghauser:
If I run e.g. `pytest -v -s tests papis` without `--tmp-xdg-config-home` in `pyproject.toml`, then I see the behaviour I had earlier, namely the same bunch of tests failing (which I guess means that the `config.py` does get parsed).

alexfikl:
> If I run e.g. `pytest -v -s tests papis` without `--tmp-xdg-config-home` in `pyproject.toml`, then I see the behaviour I had earlier, namely the same bunch of tests failing (which I guess means that the `config.py` does get parsed).

I can't seem to reproduce that on my system, not quite why.. :confused:

jghauser:
Can you maybe share the `config.py` that you used to test this so that I can give it a spin?

alexfikl:
Literally just added
```
if True:
	raise ValueError("What are you doing here?")
```
so that the tests just fail if they try to access the file at all and you can see which ones potentially.

jghauser:
If I use your `config.py`, then, without `--tmp-xdg-config-home` I get the error from the `config.py`, meaning that it does seem to get picked up. If I add `--tmp-xdg-config-home`, everything works. :thinking:

I just tested with an archlinux container, and I'm seeing what your seeing as well. So it's at least as possible that there's something odd about my environment rather than yours haha. I'll try to figure out what it could be!

alexfikl:
> If I use your `config.py`, then, without `--tmp-xdg-config-home` I get the error from the `config.py`, meaning that it does seem to get picked up. If I add `--tmp-xdg-config-home`, everything works. 🤔

Are the tests (`python -m pytest -v -s tests`) or the doctests (`python -m pytest -v -s papis`) failing for you? The doctests fail for me too.. I forgot to run those :\

jghauser:
> Are the tests (`python -m pytest -v -s tests`) or the doctests (`python -m pytest -v -s papis`) failing for you? The doctests fail for me too.. I forgot to run those :\

Both :laughing:

alexfikl:
> Both 😆

Ok, ok, then there's something to investigate still! The doctests are supposed to fail without the flag. :grin:

jghauser:
Figured it out(ish). I've been running `pytest -v -s tests` rather than `python -m pytest -v -s tests`, and while the first fails, the latter works (though I do not know why)! So current situation is:
- with this PR, with `--tmp-xdg-config-home`, all tests and doctests pass
- with this PR, without `--tmp-xdg-config-home` 
  - when adding `python -m`, the tests pass and I get two errors for the doctests
  - without adding `python -m`, both tests and doctests fail
- without this PR, tests and doctests fails no matter whether I add `python -m`

Thanks so much for helping troubleshoot this!

alexfikl:
> Figured it out(ish). I've been running `pytest -v -s tests` rather than `python -m pytest -v -s tests`, and while the first fails, the latter works (though I do not know why)! So current situation is:
> 
> * with this PR, with `--tmp-xdg-config-home`, all tests and doctests pass no matter whether I add `python -m`
> 
> * with this PR, without `--tmp-xdg-config-home`
>       
>   * when adding `python -m`, the tests pass and I get two errors for the doctests
>   * without adding `python -m`, both tests and doctests fail
> 
> * without this PR, tests and doctests fails no matter whether I add `python -m`
> 
> 
> Thanks so much for helping troubleshoot this!

That makes sense.. and somehow didn't think of it again :( Sorry for all the back and forth! For some explanation of what's going on (Basically with `-m` it correctly picks up the current directory):
https://stackoverflow.com/questions/66089649/differance-between-pytest-and-python-m-pytest-tests

As far as I can tell, this all works as intended and I'm not sure if we can make it more intuitive :\ The `--tmp-xdg-config-home` flag will stay in there, so hopefully it will just work for most users.

jghauser:
Excellent, thanks for the explanation! And no need to be sorry, it's *at least* as much my own fault for not following our own instructions :wink:. Glad we could sort this out! :tada:

PR #788, opened 2024-03-18
docs: update testing docs

alexfikl:
This adds some additional warnings to the testing docs to mention the new flags that enable running the tests in a clean environment.

@jghauser Added some more context to `docs/source/testing.rst` about those flags and why they're there. Let me know if you think anything else would be helpful to mention! I also renamed them to make it clear they're papis-specific.

xref: #786 #758.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -9,6 +9,11 @@ Coding Style
 * Follow the [PEP8 style guide](https://www.python.org/dev/peps/pep-0008/)
 * Try and run tests locally before submitting a new PR.
 
+Issues
+------
+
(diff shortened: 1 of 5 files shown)


alexfikl:
> This looks good! I was just writing down some things and then realised those aren't even things you've changed 😅.

Thanks for taking a look! :grin:

> Hopefully, I'll eventually get to take a more sustained look at the whole docs...

That would be so awesome! :tada: They're a bit better than a year ago, but still a lot of work to do..

jghauser:
They're definitely in a much better state than a year ago -- and in no small part because of you!

PR #790, opened 2024-03-18
feat/hayagriva editor list

### LINKED ISSUE #789: Multiple editors squashed into a single string, which Typst doesn't accept (too many parts error)
* papis version (`$ papis --version` or commit number):

git main

# how to reproduce the issue
```
papis export --format typst
```
During my papis -> typst export for a deliverable, I realised that, for one of my references, the editor wasn't formatted properly in the resulting yaml, making Typst throw a "too many parts" error.

Expected:

Multiple editors have to be put into a list for Typst to be happy.

kiike:
Fixes #789

I made it so we always output a list for the editors and now we also support `editor_list`. Making a list will also make the yaml easier to visually browse.

### DIFF
--- a/papis/hayagriva.py
+++ b/papis/hayagriva.py
@@ -131,7 +131,9 @@ PAPIS_TO_HAYAGRIVA_KEY_CONVERSION_MAP = [
     _k("author_list", [{"key": "author", "action": lambda a: to_hayagriva_authors(a)}]),
     _k("year", [{"key": "date", "action": None}]),
     _k("date", [{"key": "date", "action": None}]),
-    _k("editor", [papis.document.EmptyKeyConversion]),
+    _k("editor", [{"key": "editor", "action": lambda a:
+                   to_hayagriva_authors(papis.document.split_authors_name([a]))}]),
+    _k("editor_list", [{"key": "editor", "action": lambda a: to_hayagriva_authors(a)}]),
     _k("publisher", [papis.document.EmptyKeyConversion]),
     _k("location", [papis.document.EmptyKeyConversion]),
     _k("venue", [{"key": "location", "action": None}]),
--- a/papis/testing.py
+++ b/papis/testing.py
@@ -126,7 +126,7 @@ PAPIS_TEST_DOCUMENTS = [
         "pages": "56-71",
         "publisher": "Scranton Publishing",
         "ref": "scott2008that",
-        "title": "That is What She Said: The Art of Office Banter",
+        "title": "That is What She Said. The Art of Office Banter",
         "type": "incollection",
         "year": "2008",
         "_test_files": 1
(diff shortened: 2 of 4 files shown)


alexfikl:
This looks good to me.

Nothing in papis uses or handles `editor_list` though, so you may see some bugs around that when exporting / importing.

kiike:
That's weird. I did get `editor_list` imports and even `serieseditor_list` when importing from zotero.

Speaking of weird, those CI failures are really strange.

kiike:
Had a quick look and the _list thing comes from https://github.com/papis/papis-zotero/blob/3ef795e06a2f206558b2d326c88670ba695d4a0c/papis_zotero/sql.py#L110. 

I can see how it might introduce maybe not errors, but surprises when exporting. Would you rather see that fragment go, along with this `_list` feature? (keeping authors_list at least)

alexfikl:
> Would you rather see that fragment go, along with this `_list` feature? (keeping authors_list at least)

I would prefer the `_list` versions, since they're easier to handle, but I have no idea what it would take to migrate everything in a consistent manner at this point. In general, it seems like there's a lot of cases where you want parts of the name and it's always easy to do a `" and ".join(author_list)` or whatever to get a single string..

For this PR, it's definitely fine to keep them both! The `BibTeX` exporter also has a bunch of special cases for the Zotero import, so it's nothing new.

kiike:
> For the errors on the CI, my best guess currently is that there's a title with a colon in it
> 
> ```
> "title": "That is What She Said: The Art of Office Banter",
> ```
> 
> and the `test_edit_run` test is very stupid and just does a sed in there and messes up the YAML. Can you try removing that?

Interesting! Why didn't it fail on windows and why did it work on my machine? I couldn't replicate it anywhere, not even in the nix flake :cry: 

Commited the removal of the colons.

kiike:
> > Interesting! Why didn't it fail on windows and why did it work on my machine? I couldn't replicate it anywhere, not even in the nix flake 😢
> > Commited the removal of the colons.
> 
> Have you checked which document it was looking at? On the CI it seemed to use `test_doc_6` but locally for me it used `test_doc_0` and the tests passed. Not sure why it got the documents in a different order though..
> 
> EDIT: #791 should also fix the issue.

That's great!

alexfikl:

kiike:
It is a very nice find! Thanks a lot for the help!

>I would prefer the _list versions, since they're easier to handle, but I have no idea what it would take to migrate everything in a consistent manner at this point. In general, it seems like there's a lot of cases where you want parts of the name and it's always easy to do a " and ".join(author_list) or whatever to get a single string..

> For this PR, it's definitely fine to keep them both! The BibTeX exporter also has a bunch of special cases for the Zotero import, so it's nothing new.

Great! Then the PR is good to go if you consider it fine. It's just a minor improvement overall. I also prefer the lists,

PR #793, opened 2024-03-19
docs(merge): resolve TODO and add examples

kiike:
I'm going through the TODOs scattered around the code, so I can learn more about the papis code.

I found a documentation TODO and I think i managed to resolve it.

It's my first doc here so let me know if it's alright. I checked it in browser and it doesn't look bad.

### DIFF
--- a/papis/commands/merge.py
+++ b/papis/commands/merge.py
@@ -7,7 +7,48 @@ pass the ``--pick`` flag to pick twice from the documents.
 Examples
 ^^^^^^^^
 
-TODO
+-   Pick the duplicate documents:
+
+    .. code:: sh
+
+        papis merge "prediction and subs"
+
+        /papers/de8d6373a8cef74b5245124ddb288d45-dejong-gerald
+        /papers/0e378c330dba4a4483660388183aa648-dejong-gerald
+
+    With the default picker and config, you can pick the two offending documents
+    with ``CTRL+T`` The selected entries will have a ``#`` in the beginning of
+    the document title.
+
+    papis will show the difference between the files and let you select which
+    ones you want to keep:
+
+    .. code:: diff
+
+        - collections: ['t3']
+        + collections: ['t3', 't3a2']
+
+    The lines are prefixed by an indication of what will happen on accepting the
+    change. papis will remove lines prefixed with a minus sign and add those
+    with a plus sign.
+
+-   Merge the documents that match the query "prediction and subs" and commit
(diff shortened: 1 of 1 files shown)


alexfikl:
Very nice! I've never actually used `papis merge`. Does it work? :grin:

kiike:
> Very nice! I've never actually used `papis merge`. Does it work? 😁

Neither did I, but I do have a lot of duplicates so stumbling upon this TODO and writing this was a good experience!

It works fine! It can merge two entries at the same time and not more, but that is alright. I had a document that was duplicated three times, so I just ran it twice.

jghauser:
Looks great, thanks! I'd maybe capitalise "Papis", though admittedly we've not been super consistent with it...

kiike:
I saw it in lowercase in `add.py`, which I was using as a reference :yum:. I'll capitalize it here and make a different pull request to uppercase all other references.

kiike:
Oh, even the same file has it in lowercase, just above examples :laughing: 

So, is uppercase the "right way"?

jghauser:
> So, is uppercase the "right way"?

Good question! I've been using "Papis" and changing things when I encountered other ways of writing it (hence, it's what's used in the README). What do you think @alexfikl?

kiike:
> > So, is uppercase the "right way"?
> 
> Good question! I've been using "Papis" and changing things when I encountered other ways of writing it (hence, it's what's used in the README). What do you think @alexfikl?
> 
> EDIT: In any case, at the beginning of sentences, I think lower case looks a bit odd.

Not if it's a command, IMHO.

Anyway I searched old versions of the code and it looks like the project name is Papis, not papis.

https://github.com/papis/papis/commit/96ce070197ab7af7c043c0aa0a0c1bfc10310df8

kiike:
Capitalised Papis in https://github.com/papis/papis/pull/793/commits/158d0416ae5e01e3404787955094bce591074911, advise if I should revert it. Thanks for giving feedback so quickly!

jghauser:
I think consistency is what matters most here and if we can have history help us decide, all the better! And thanks for digging up this old commit, interesting to see how far Papis has come :blush:

alexfikl:
> Good question! I've been using "Papis" and changing things when I encountered other ways of writing it (hence, it's what's used in the README). What do you think @alexfikl?

I've been doing the same (possibly inspired by some of your documentation changes), so definitely :+1: for having it capitalized.

PR #794, opened 2024-03-20
docs: capitalize Papis

kiike:
After the comments in #793, I am hunting Papis written in lowercase in docstrings and comments.

This will track the progress, and also let me spot false positives here too.

### DIFF
--- a/papis/commands/__init__.py
+++ b/papis/commands/__init__.py
@@ -86,7 +86,7 @@ class AliasedGroup(click.core.Group):
 class Script(NamedTuple):
     """A papis command plugin or script.
 
-    These plugins are made available through the main papis command-line as
+    These plugins are made available through the main Papis command-line as
     subcommands.
     """
 
@@ -99,7 +99,7 @@ class Script(NamedTuple):
 
 
 def get_external_scripts() -> Dict[str, Script]:
-    """Get a mapping of all external scripts that should be registered with papis.
+    """Get a mapping of all external scripts that should be registered with Papis.
 
     An external script is an executable that can be found in the
     :func:`papis.config.get_scripts_folder` folder or in the user's PATH.
@@ -130,7 +130,7 @@ def get_external_scripts() -> Dict[str, Script]:
 
 
 def get_scripts() -> Dict[str, Script]:
-    """Get a mapping of commands that should be registered with papis.
+    """Get a mapping of commands that should be registered with Papis.
 
     This finds all the commands that are registered as entry points in the
     namespace ``"papis.command"``.
@@ -156,7 +156,7 @@ def get_scripts() -> Dict[str, Script]:
 
 
 def get_all_scripts() -> Dict[str, Script]:
-    """Get a mapping of all commands that should be registered with papis.
+    """Get a mapping of all commands that should be registered with Papis.
 
     This includes the results from :func:`get_external_scripts` and
     :func:`get_scripts`. Entrypoint-based scripts take priority, so if an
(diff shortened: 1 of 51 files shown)


papis/commands/__init__.py
```diff
@@ -86,7 +86,7 @@ def get_command(self,
 class Script(NamedTuple):
     """A papis command plugin or script.
 
-    These plugins are made available through the main papis command-line as
```
jghauser, line 89: I think this might be a case where "`papis`" is more appropriate as it seems to be talking about the command (maybe it should also be 'command' instead of 'command-line'?).

papis/commands/add.py
```diff
@@ -1,5 +1,5 @@
 """
-The ``add`` command is one of the central commands of the papis command-line
+The ``add`` command is one of the central commands of the Papis command-line
```
jghauser, line 2: Same as above here

papis/commands/__init__.py
kiike: I thought about that exact case. I didn't know if it was referring to `papis` as the command. So formatting and removing will make it clearer.

kiike:
> Nice! I actually had the same idea after yesterday's conversion -- great to see you got there first! I left two small comments though admittedly they're not clear-cut cases

Thanks for checking it! There are some more I haven't committed yet. But here they're easy to see in context and we can help clarify the uses if needed.

Cheers!

papis/commands/__init__.py
jghauser: Yeah, I agree it's not a clear case. Best to make it clear :grinning:

kiike:
Ok, so I applied a "rule":

- have `papis` in lowercase only with that format when I think it's referring to the program
- papis lowercase without format shouldn't be here anymore.
- Papis uppercase when it's generic or referring to the project, ie. the Sci-Hub plugin is not distributed with Papis, or let's import a Papis document or Papis library.

I am reviewing to see if I could apply it consistently.

jghauser:
> Ok, so I applied a "rule":
>     * have `papis` in lowercase only with that format when I think it's referring to the program
>     * papis lowercase without format shouldn't be here anymore.
>     * Papis uppercase when it's generic or referring to the project, ie. the Sci-Hub plugin is not distributed with Papis, or let's import a Papis document or Papis library.

I think that's exactly right! :+1:

alexfikl:
> Ok, so I applied a "rule":
>
> I am reviewing to see if I could apply it consistently.

That sounds very reasonable! Can you also add a little blurb about the rules to the `HACKING.md` file so we can point at it when we need to? Maybe under a `Guidelines for Documentation`.

kiike:
Oh definitely! Good idea! I was trying to have codespell handle this but I couldn´t get it to.

> > Ok, so I applied a "rule":
> > I am reviewing to see if I could apply it consistently.
> 
> That sounds very reasonable! Can you also add a little blurb about the rules to the `HACKING.md` file so we can point at it when we need to? Maybe under a `Guidelines for Documentation`.

kiike:
Let me know if you have any feedback! I couldn't find more uses of lowercase Papis that go against the new "guidelines" :smile:. Thanks!

alexfikl:
Gave it a skim and read the guidelines section and it all looked great. I'm happy to merge this when you think it's done!

jghauser:
Same here, gave it a quick read and it looks all good!

kiike:
Awesome! :heart: It's ready to merge then!

Issue #801, opened 2024-03-27
build(windows): Make an installer and Winget Package


kiike:
Following #795, and preparing for #799  I think it would be nice to have Papis available for download with Winget. Then the program will be just a `winget` away, updates included!

Requirements

- [x] MSI installer (in #802 )
- [ ] Entry at [microsoft/winget-pkgs](https://github.com/microsoft/winget-pkgs) (not started)

PR #802, opened 2024-03-27
build(windows): add an installer

kiike:
After creating a Windows executable for papis, I think having an installer is the obvious next step. It didn't take much effort to create besides skimming through WiX examples. Right now the installer is not customizable at all, when executed, if you don't have the program installed, welcomes you and after clicking next, installs the program to your `AppLocalData\Programs\papis`, which, for me is `C:\enric\AppData\Local\Programs\papis\`, then adds that location to PATH and says goodbye with info about the PATH thing.

The installer file is created during the windows GitHub Action. For that,  a couple support steps were needed. Also, editing the version in the MSI because it cannot contain letters apparently.

I've tried the program on my Windows 10 system and it works fine. Repairing and removing the program too. And most importantly, upgrading.

I used the default components of WiX and it fills me with nostalgia:

image

I think it can help for #799!

Fixes first item of #801.

EDIT: The whole point of having the MSI is actually not seeing it at all when installing from winget. Or, well, if somebody uses the MSI in an interactive way, different strokes and all.


alexfikl (approved):
I added one small question, but otherwise this looks good to me! I'm fine with merging it when you think it's ready. 

I also tested this very briefly on a windows laptop and it seems to work. However, the `papis.exe` pyinstaller generates seems to be very slow.. it takes a good second or two to do `papis --help`. Did you see the same slowdown?

.github/workflows/windows.yml
```diff
@@ -32,17 +32,41 @@ jobs:
         sed -i \
           "s/^version =\".*\"/version =\"${version}\"/" \
           pyproject.toml
-    
+        msi_version=$(
+            git describe --tags --dirty \
+            | sed -e 's/^v//' -e 's/-\([^-]*-g\)/.\1/;s/-g.*//g'
+          )
+        sed -i "s/Version=\".*\"/Version=\"${msi_version}\"/" \
+          scripts/windows/papis.wxs
```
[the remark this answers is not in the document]
kiike: The windows installer sdk doesn't allow letters in versions.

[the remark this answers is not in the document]
alexfikl: Ah! I also stumbled upon some version strings and actually `X.Y.devZ+HASHID` is valid according to the PEP (missed it before :\)
https://peps.python.org/pep-0440/#local-version-identifiers
so maybe we can switch to something like what you initially suggested
```bash
git describe --tags --dirty | awk -F '-' '{ printf "%s.dev%s+%s\n", substr($1, 2), $2, $3 }'
```
?

kiike:
> I added one small question, but otherwise this looks good to me! I'm fine with merging it when you think it's ready.
> 
> I also tested this very briefly on a windows laptop and it seems to work. However, the `papis.exe` pyinstaller generates seems to be very slow.. it takes a good second or two to do `papis --help`. Did you see the same slowdown?

I did, although not as strongly as you. That is due to the one-file bundle. On execution, pyinstaller extracts the runtime to a temporary folder, from which the files are actually loaded.  The alternative, which I am trying out, is the "one-dir" bundle. It is much faster due to no extraction needed; the runtime is installed in a known location. The problem then lies on the installer which is what I am working on as we speak.

alexfikl:
> > I added one small question, but otherwise this looks good to me! I'm fine with merging it when you think it's ready.
> > I also tested this very briefly on a windows laptop and it seems to work. However, the `papis.exe` pyinstaller generates seems to be very slow.. it takes a good second or two to do `papis --help`. Did you see the same slowdown?
> 
> I did, although not as strongly as you. That is due to the one-file bundle. On execution, pyinstaller extracts the runtime to a temporary folder, from which the files are actually loaded. The alternative, which I am trying out, is the "one-dir" bundle. It is much faster due to no extraction needed; the runtime is installed in a known location. The problem then lies on the installer which is what I am working on as we speak.
> 
> I'd like to improve this I'll tell you when it's ready 😄

That sounds awesome! 

Have you looked at something like [nuitka](https://nuitka.net/)? Although not sure if it even works for us, because we do make use of quite a few very dynamic python features..

.github/workflows/windows.yml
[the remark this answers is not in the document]
alexfikl: That's fair enough! Maybe add a little comment so we don't forget :grin:

kiike:
> I added one small question, but otherwise this looks good to me! I'm fine with merging it when you think it's ready.
> 
> I also tested this very briefly on a windows laptop and it seems to work. However, the `papis.exe` pyinstaller generates seems to be very slow.. it takes a good second or two to do `papis --help`. Did you see the same slowdown?



> > > I added one small question, but otherwise this looks good to me! I'm fine with merging it when you think it's ready.
> > > I also tested this very briefly on a windows laptop and it seems to work. However, the `papis.exe` pyinstaller generates seems to be very slow.. it takes a good second or two to do `papis --help`. Did you see the same slowdown?
> > 
> > 
> > I did, although not as strongly as you. That is due to the one-file bundle. On execution, pyinstaller extracts the runtime to a temporary folder, from which the files are actually loaded. The alternative, which I am trying out, is the "one-dir" bundle. It is much faster due to no extraction needed; the runtime is installed in a known location. The problem then lies on the installer which is what I am working on as we speak.
> > I'd like to improve this I'll tell you when it's ready 😄
> 
> That sounds awesome!
> 
> Have you looked at something like [nuitka](https://nuitka.net/)? Although not sure if it even works for us, because we do make use of quite a few very dynamic python features..

[Yes!](https://github.com/kiike/papis/actions/runs/8378554418) but it seems very popular in the malware scene and it did trigger the windows antivirus. The startup time was not better either. And the buildtime was very lengthy as you can see and I'm not sure if we might have problems with the GitHub Actions if we used that.

kiike:
I had some issues with the GitHub action while creating the manifest for WiX. The issues are resolved and now we have two artifacts:

- the standalone executable, from which you can run papis, taking a good one second to run, but does not need installation
- the installer, which installs the "onedir" version, the default build mode of pyinstaller. This is the faster method. Trying with an almost empty library:

```
$ time papis.exe list
C:\Users\enric/Documents/papers\32acf979707b0fbd72f670832ed54813-saxena-anchal-and-m

real    0m0.661s
user    0m0.015s
sys     0m0.046s

$ time papis.exe --version
papis.exe, version 0.13

real    0m0.573s
user    0m0.015s
sys     0m0.031s

```

This is mergeable.

Some comments for you, @alexfikl, I know you'll be curious:

- we now need to list all the files that need to be installed into the `papis.wxs` manifest, which are the instructions to build the MSI. It could be done by hand but it's tedious and boring, so I made a script that walks the pyinstaller artifacts dir and builds the manifest, including in the "template" manifest and writing it into `papis.wxs`. The template is `papis_template.wxs`.
- The Ids in that script reflect the directory or filename. WiX has some restrictions in IDs so with the hashes I am aiming to avoid them, and also reliably identify the directories.

With WiX 5 in theory that won't be needed, but it's not released yet.

.github/workflows/windows.yml
```diff
@@ -32,17 +40,44 @@ jobs:
...
+
+    - name: Install WiX
+      run: |
+        dotnet tool install --global wix --version 4.0.5
+        wix extension -g add WixToolset.UI.wixext/4.0.5
+    
+    - name: Render the MSI build manifest template
+      shell: bash
+      run: |
+        python scripts/windows/wix-populate_internal_artifacts.py
```
alexfikl, line 72:
```suggestion
        python scripts/windows/wix-populate-internal-artifacts.py
```
? Looks weird to me to mix dashes and underscores.

scripts/windows/wix-populate_internal_artifacts.py
```diff
@@ -0,0 +1,59 @@
...
+
+import os
+import lxml.etree as et
+import uuid
+
+
+def get_id(*path_components: str) -> str:
+    """Create an element ID based on the WiX constraints:"""
+    path = os.path.join(*path_components)
+    result = f"_{str(abs(hash(path)))}"
```
alexfikl, line 12:
```suggestion
    result = f"_{abs(hash(path))}"
```

scripts/windows/wix-populate_internal_artifacts.py
```diff
@@ -0,0 +1,59 @@
+"""Add all components (files) and directories generated by pyinstaller in
+dist/papis/_internal into the XML manifest that is used to create the MSI."""
+
+import os
+import lxml.etree as et
+import uuid
+
+
+def get_id(*path_components: str) -> str:
+    """Create an element ID based on the WiX constraints:"""
```
alexfikl, line 10: Is there a place describing the "WiX constraints"? Would be nice to link it here for future reference.

scripts/windows/wix-populate_internal_artifacts.py
```diff
@@ -0,0 +1,59 @@
...
+                "Component",
+                Guid=str(uuid.uuid4()),
+                Id=f"c{get_id(root, file)}",
+                Directory=get_id(root),
+            )
+            et.SubElement(
+                component, "File", KeyPath="yes", Source=os.path.join(root, file)
+            )
+
+with open("scripts\\windows\\papis.wxs", mode="wb") as f:
```
alexfikl, line 58:
```suggestion
with open("scripts\\windows\\papis.wxs", mode="w") as f:
```
? Why the `b` in there?

alexfikl (approved):
Your timings also look very reasonable! The machine I tested on is quite a bit older (and just an `i3`), so maybe that's why it was slower. On Linux I'm getting around `0.3s` for a `papis list`, so `0.6` isn't bad at all!

Left a few nitpicks in the python script, but the rest looks good to me! Take a look and fix whatever looks reasonable and then we can merge this :grin:

.github/workflows/windows.yml
kiike: yeah I wanted to separate wix from the other stuff. this is fine too!

scripts/windows/wix-populate_internal_artifacts.py
kiike: definitely!

scripts/windows/wix-populate_internal_artifacts.py
kiike: The XML might look like a text file, but i would get an error saying that the file writer expected writing bytes, not a string. So I complied.

scripts/windows/wix-populate_internal_artifacts.py
kiike: Sure!

kiike:
I've addressed your comments. Let's see if the build succeeds, in which case it's good to go.

alexfikl:
Nice! Thanks for working on this! In it goes :rocket:

Issue #311, opened 2020-12-24
clean_document_name odd behaviour


jghauser:
I think the `clean_document_name` function's behaviour can be a bit strict and at times also faulty. I noticed this when using the jinja2 formatter. Is there a reason that underlines aren't allowed? And capital letters?

More importantly, however, backslashes seem to lead to the whole string until the backslash to be suppressed. When I pass `Reda_2015_2014/2256(INI)-Draft-Report-on` to the function, it returns `2256-ini-draft-report-on`. This is unfortunate because now both the author and year are suppressed. 

I haven't tested it yet, but I suspect similar problems might exist for `ref_cleanup` in `bibtex.py`

EDIT: Maybe, I should explain what I'm trying to achieve. I was trying to adapt your script in in #295 to update my folder names, which is where the problem with the forward slashes appeared. It doesn't seem to happen when adding a new document. The other issues (re underlines, capital letters) happen also when adding a new document.

I also had another look at `ref_cleanup`. I'm unable to create ref tags with underscores and with only small letters (I was trying to recreate the tag style I used in zotero e.g. `dworkin_model_1967`).

jghauser:
Closing this because I can't see the error anymore. Not sure if I was doing something wrong or something has changed in papis itself.

alejandrogallo:
Hallo Julian,

thanks for your comments. I hope I can answer
all of the issues.

- Clean document name I think you're right to be
  buffled by its behaviour, this is the definition
  of the function basically

  ```python
  def clean_document_name(doc_path: str) -> str:
      import slugify
      regex_pattern = r'[^a-z0-9.]+'
      return str(slugify.slugify(
          os.path.basename(doc_path),
          word_boundary=True,
          regex_pattern=regex_pattern))
  ```

  and as you see the function take the basename of `doc_path` ,
  which is the argument to the function.
  So actually the naming
  of the function is not accurate, it should read
  `clean_document_path` I guess. That's why your string gets mutilated
  before the `/`.
  Furthermore, you also see that the only characters that are allowed
are
  alphanumerics and the dot, and they will be replaced by `-`.

- Allowing for `_` I guess we can do it for paths, yes.
  In general it's best to limit the characters that can appear in
  paths, but `_` is not a problematic one in afaik neither in WINDOWS
  nor in *nix systems.

- About having the `_` in references, as far as I know
  latex can't deal with underscores, that's why I decided
  to be strict about it, since most people will not ever want
  to have an underscore in their reference when working with *tex
[...]

PR #803, opened 2024-03-29
add slugify configuration options

choreutes:
This adds three options to configure the behaviour of slugify when cleaning document paths:

- ``slugify-lowercase``
- ``slugify-regex-pattern``
- ``slugify-separator``

Their effects mostly follow the respective arguments to ``python-slugify``. Hopefully appropriate documentation about this is added to ``default-settings.rst``.

As this is my first PR for this project any feedback is welcome and greatly appreciated!

### DIFF
--- a/doc/source/default-settings.rst
+++ b/doc/source/default-settings.rst
@@ -165,35 +165,31 @@ General settings
 
     **Note** The older (misspelled) version ``"formater"`` is deprecated.
 
-.. papis-config:: slugify-lowercase
+.. papis-config:: doc-paths-lowercase
     :type: bool
 
     This setting controls whether capital letters in a (generated or specified)
     document path should be lowercased before the path is created.
 
-    **Note:** Lowercasing happens *before* removing disallowed characters.
-    If you disable it without allowing uppercase letters they will simply
-    disappear from document paths!
-
-.. papis-config:: slugify-regex-pattern
+.. papis-config:: doc-paths-extra-chars
     :type: str
 
-    This setting defines the set of *disallowed* characters for document paths.
-    It must be a valid regular expression matching all disallowed characters.
-    The default value disallows all characters except lowercase letters and
-    numbers (and dots for file extensions).
+    By default document paths in Papis libraries can contain only a limited set
+    of characters. This is mainly to exclude characters that are invalid for
+    file paths on any operating system or possibly unprintable. Allowed
+    characters are:
 
-    Possible alternative values are:
+    * latin letters (a to z)
+    * arabic digits (0 to 9)
+    * dots (for file extensions)
+    * directory separators (usually ``/`` on UNIX-like systems and ``\\``
+      on Windows)
 
-    * ``r"[^a-zA-Z0-9.]+"``: Don't filter out uppercase letters from the path.
-    * ``r"[^a-z0-9._]+"``: Don't filter out underscores from the path.
+    This setting allows to append additional characters to this set. It expects
+    a string containing all additional valid characters. A possible value would
+    be ``"_"`` to allow underscores in document paths.
 
-    **Note:** Allowing uppercase letters here will not yield the desired result
-    unless lowercasing is also turned off. Otherwise all uppercase letters are
-    lowercased before filtering happens and no uppercase letters will appear
-    in the path anyway.
-
-.. papis-config:: slugify-separator
+.. papis-config:: doc-paths-word-separator
     :type: str
 
     This setting defines the separator between words in document paths (usually
--- a/doc/source/default-settings.rst
+++ b/doc/source/default-settings.rst
@@ -165,6 +165,41 @@ General settings
 
     **Note** The older (misspelled) version ``"formater"`` is deprecated.
 
+.. papis-config:: slugify-lowercase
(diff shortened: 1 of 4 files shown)


doc/source/default-settings.rst
```diff
@@ -165,6 +165,41 @@ General settings
 
     **Note** The older (misspelled) version ``"formater"`` is deprecated.
 
+.. papis-config:: slugify-lowercase
```
alexfikl, line 168:
```suggestion
.. papis-config:: document-name-lowercase
```
and same for the others? Suggestions welcome, but `slugify` seems very tied to using that one library to do the cleanup.

doc/source/default-settings.rst
```diff
@@ -165,6 +165,41 @@ General settings
...
+
+    Possible alternative values are:
+
+    * ``r"[^a-zA-Z0-9.]+"``: Don't filter out uppercase letters from the path.
+    * ``r"[^a-z0-9._]+"``: Don't filter out underscores from the path.
+
+    **Note:** Allowing uppercase letters here will not yield the desired result
+    unless lowercasing is also turned off. Otherwise all uppercase letters are
+    lowercased before filtering happens and no uppercase letters will appear
+    in the path anyway.
```
alexfikl, line 194: This is a bit confusing. If I understand correctly:
* If you want uppercase, set `slugify-lowercase=False` and `slugify-regex="[^a-zA-Z0-9]+"`
* If you want lowercase, set `slugify-lowercase=True` and `slugify-regex="[^a-z0-9]+"`

I would be nice to somehow ensure that these can't be inconsistent, but I don't really have any clean ideas :\

alexfikl (commented):
This is almost good to go! Can you also add a little test for some combinations of `slugify-lowercase` and `slugify-regex`?

doc/source/default-settings.rst
alexfikl: Maybe the `regex` part should just be about additional characters that are allowed? i.e. we construct the full regex ourselves:
* If `slugify-lowercase=False`: set the regex to `r"[^a-zA-Z0-9{extras}]"`
* If `slugify-lowercase=True`: set the regex to `r"[^a-z0-9{extras}]"`

where `extras = papis.config.getstring("slugify-regex")`. What do you think?

doc/source/default-settings.rst
jghauser: I'm not sure 'document-name' is appropriate. I know the function is called that, but isn't it also used for paths? Maybe something like 'filepath'? Not the prettiest word, but at least it's descriptive...

alexfikl: Yeah, I think it's pretty exclusively used for files and folders. `document-paths-lowercase`?

doc/source/default-settings.rst
choreutes: Oh yeah, that's an awesome idea! I also thought about enforcing consistency there but didn't quite know how to go about it.

doc/source/default-settings.rst
choreutes: I shortened the ``document`` to ``doc``because the option names got pretty long otherwise. I hope that is still okay with you.

choreutes:
> This is almost good to go! Can you also add a little test for some combinations of `slugify-lowercase` and `slugify-regex`?

I'll try to come up with some test cases early next week once you had the chance to have a look at my changes to the options.

doc/source/default-settings.rst
alexfikl: That's fine! Thanks for fixing it!

alexfikl:
> > This is almost good to go! Can you also add a little test for some combinations of `slugify-lowercase` and `slugify-regex`?
> 
> I'll try to come up with some test cases early next week once you had the chance to have a look at my changes to the options.

There are some tests already in there:
https://github.com/papis/papis/blob/0edea03bffecff133094a1b54a14bf2c2bcb3aa9/tests/test_utils.py#L120
Just adding some cases where the result is uppercased or some extra characters are allowed would be perfect!

doc/source/default-settings.rst
alexfikl: Perfect! Thanks!

alexfikl:
> It is not as early this week as I would have liked but I have added some ~totally new and original~ brazenly copied test cases for the new configuration options. It's not every combination of the options there is but I think they cover most of it. Is this somewhat like what you expected? Sorry anyways for delaying the new release this long!

No worries at all! Thanks for working on this! :heart: 

I'm fine with merging this as soon as the tests pass. Is everything working as expected to fix #311 for you?

jghauser:
Thanks, @choreutes! And no worries about the delay, that happens to all of us :blush:.

Yes @alexfikl, this resolves my issue in #311!

choreutes:
Yes, I guess this adds all the functionality that I wanted. And the tests seem to check out as well. So I mark this as ready for review now and then we're done here, right?

alexfikl:
> Yes, I guess this adds all the functionality that I wanted. And the tests seem to check out as well. So I mark this as ready for review now and then we're done here, right?

Yeah, this is very nice! In it goes :rocket:

PR #806, opened 2024-03-31
Update dependencies

alexfikl:
This cleans up our dependencies a bit. Mainly it removes `tqdm`, since `prompt_toolkit` has a progress bar as well that looks pretty much the same.


alexfikl:
@hseg FYI, this removes the `tqdm` dependency.

hseg:
Thanks for the heads' up -- missed it reviewing the news. Though `tqdm` is still a transitive dependency via `habanero`, but at least I can remove the `tqdm-types` dep.

PR #807, opened 2024-04-02
ci: add 3.13-dev to the tests

alexfikl:
Just running the CI with the current Python 3.13alpha4 to see what breaks.


alexfikl:
So current failures are due to:
* `lxml` wheels not being available for python 3.13
* `cgi` module being removed from the standard library

We're in a pretty good state for the next version :heart:

PR #808, opened 2024-04-02
scripts(windows): Simplify populate_wxs_artifacts

alexfikl:
Noticed in #807 that this is the only place that hard depends on `lxml` being available, so this ports it to the stdlib `xml.etree` version. On top of that, it makes a few more improvements:
* Adds command-line support using `argparse`.
* Uses `ElementTree.write` directly. This doesn't pretty-print, but it's just a temporary file, so should be ok.
* Allows passing the MSI version string to the script directly without the need to `sed` in the CI `windows.yml`.
* Uses `awk` in the CI, since it's a bit easier to understand how the versions are constructed.

cc @kiike


alexfikl:
Hm.. apparently I broke it.. :cry:

kiike:
I see it's breaking. It's probably due to namespacing, but since I am not seeing the rendered WXS file, I am not absolutely sure. If you continue having issues, is might be worth it to just build the WXS in python, then we won't have to open a file and save it, potentially having some namespaced components and some without.

Let me know if you'd rather have me working on this or if you prefer doing it yourself :smile:

alexfikl:
> I see it's breaking. It's probably due to namespacing, but since I am not seeing the rendered WXS file, I am not absolutely sure. If you continue having issues, is might be worth it to just build the WXS in python, then we won't have to open a file and save it, potentially having some namespaced components and some without.

Yeah, the builtin `xml.etree` doesn't seem to be very smart about namespacing. I'm comparing the files locally, so hopefully I'll figure this out soon :))

kiike:
Good luck!

alexfikl:
Hm, well, the build seems to pass, but the CI says that the `dist/papis/_internal` directory doesn't exist and it seems to create a much smaller MSI installer, so something is still wrong..

alexfikl:
@kiike This seems to be finally working again.. most of the issues were just me not understanding the code :(

Not sure this was worth the effort overall :\ But it would be helpful if you could take a look to check that nothing looks fishy!

scripts/windows/populate_wxs_artifacts.py
```diff
@@ -1,63 +1,123 @@
...
+        component, "Directory",
+        Name="_internal",
+        Id=get_id(root_dir)
+    )
+    component_group = ElementTree.SubElement(
+        package, "ComponentGroup",
+        Id="_internal"
+    )
+
+    package.attrib["Version"] = version
```
kiike, line 70: Nice!

.github/workflows/windows.yml
```diff
@@ -25,27 +25,28 @@ jobs:
...
+      run: |
+        make ci-install
+        pip install '.[windows]'
 
     - name: Set version in pyproject.toml, MSI and environment
       shell: bash
       run: |
-        set -x
-        version=$(
+        # NOTE: python version follows "X.YdevZ+gHASH" format
```
kiike, line 36: This is great!

kiike (approved):
- I'm happy you got it working with ElementTree, avoiding the hard dependency on lxml. You made me learn about the namespacing spell with ET.
- The hashing with md5 is pretty smart so that we have the same hashes between runs. Not that we see the WXS at all but perhaps for local debugging it might be better.
- I like the version change.
- The argument parsing is a good addition, but since the dist path is controlled by the pyinstaller spec file or the action command in case it overrides it, not sure we'll have a good use of it.

Thank you for improving this!

alexfikl:
Thanks for taking a look! 

> * The hashing with md5 is pretty smart so that we have the same hashes between runs. Not that we see the WXS at all but perhaps for local debugging it might be better.

Yeah, I mostly changed that for debugging. It made it a lot easier to check what changed in the generated files when comparing to see what broke.

> * The argument parsing is a good addition, but since the dist path is controlled by the pyinstaller spec file or the action command in case it overrides it, not sure we'll have a good use of it.

Yeah, I'll go ahead and remove that `--dist` argument. We can extend it if needed later.

PR #810, opened 2024-04-10
feat(rename): allow batch and auto-renaming from pattern

kiike:
When I develop a feature, I try to dogfood, so sometimes I end up using default values of config options such as `add-folder-name`, thus creating folders that don't end up like I like. `papis rename` can rename entries but is now pretty limited.

With this WIP pull request, I am trying to make `papis rename` a bit more powerful:

- allow selecting multiple entries with the picker
- selecting all entries matching a query, for mass renaming of folders
- optional slugify of names
- rename without confirmation
- rename using the generated folder name

It's taking form:

With slug enabled:
```
 papis rename -arsb "black hole"
[INFO] commands.rename: Renaming 'raju-2024-howdoesinformationemerge' into 'test-2024-how-does-information-emerge-from-a-black-hole'
[INFO] commands.rename: Renaming 'macedo-2024-thermodynamicsandquasinormalmodes' into 'test-2024-thermodynamics-and-quasinormal-modes-of-the-dymnikova-black-hole-in-higher-dimensions'
[INFO] commands.rename: Renaming 'ressler-2024-blackhole-diskinteractionsin' into 'test-2024-black-hole-disk-interactions-in-magnetically-arrested-active-galactic-nuclei-general-relativistic-magnetohydrodynamic-simulations-using-a-time-dependent-binary-metric'
```

Without:
```
 papis rename -arb "black hole"
[INFO] commands.rename: Renaming 'test-2024-how-does-information-emerge-from-a-black-hole' into 'Test 2024 How does information emerge from a black hole?'
[INFO] commands.rename: Renaming 'test-2024-thermodynamics-and-quasinormal-modes-of-the-dymnikova-black-hole-in-higher-dimensions' into 'Test 2024 Thermodynamics and Quasinormal Modes of the Dymnikova Black Hole in Higher Dimensions'
[INFO] commands.rename: Renaming 'test-2024-black-hole-disk-interactions-in-magnetically-arrested-active-galactic-nuclei-general-relativistic-magnetohydrodynamic-simulations-using-a-time-dependent-binary-metric' into 'Test 2024 Black Hole-Disk Interactions in Magnetically Arrested Active Galactic Nuclei: General Relativistic Magnetohydrodynamic Simulations Using A Time-Dependent, Binary Metric'
```

Picking files:
```
 papis rename -rbs "black hole"
#How does information emerge from a black hole?
  Raju, Suvrat
   (2024) []
#Thermodynamics and Quasinormal Modes of the Dymnikova Black Hole in Higher Dimensions
  Macêdo, M. H. and Furtado, J. and Alencar, G. and Landim, R. R.
   (2024) []
|Black Hole-Disk Interactions in Magnetically Arrested Active Galactic Nuclei: General Relat
| Ressler, Sean M. and Combi, Luciano and Li, Xinyu and Ripperda, Bart and Yang, Huan
|  (2024) []

[INFO] commands.rename: Renaming 'Test 2024 How does information emerge from a black hole?' into 'test-2024-how-does-information-emerge-from-a-black-hole'
[INFO] commands.rename: Renaming 'Test 2024 Thermodynamics and Quasinormal Modes of the Dymnikova Black Hole in Higher Dimensions' into 'test-2024-thermodynamics-and-quasinormal-modes-of-the-dymnikova-black-hole-in-higher-dimensions'
```
What do you think?

### DIFF
--- a/papis/commands/rename.py
+++ b/papis/commands/rename.py
@@ -56,8 +56,6 @@ Command-line Interface
     :prog: papis rename
 """
 
-
-import os
 from typing import Optional, Tuple, List
 
 import click
@@ -80,6 +78,8 @@ logger = papis.logging.get_logger(__name__)
 
 def run(document: papis.document.Document,
         new_name: str, git: bool = False) -> None:
+    import os
+
     db = papis.database.get()
     folder = document.get_main_folder()
 
--- a/papis/commands/rename.py
+++ b/papis/commands/rename.py
@@ -2,7 +2,6 @@
 The ``rename`` command is used to rename document folders based on
 a provided format or using :confval:`add-folder-name`.
 
-
 Examples
 ^^^^^^^^
 
@@ -77,7 +76,8 @@ logger = papis.logging.get_logger(__name__)
 
 
 def run(document: papis.document.Document,
-        new_name: str, git: bool = False) -> None:
+        new_folder_name: str,
+        git: bool = False) -> None:
     import os
     import shutil
 
@@ -87,24 +87,22 @@ def run(document: papis.document.Document,
     if not folder:
         raise DocumentFolderNotFound(papis.document.describe(document))
 
-    subfolder = os.path.dirname(folder)
-    new_folder_path = os.path.join(subfolder, new_name)
+    parent = os.path.dirname(folder)
+    new_folder_path = os.path.join(parent, new_folder_name)
+    logger.debug("New document folder: '%s'.", new_folder_path)
 
     if os.path.exists(new_folder_path):
         logger.warning("Path '%s' already exists.", new_folder_path)
         return
 
     if git:
-        papis.utils.run(["git", "mv", folder, new_folder_path],
-                        cwd=folder)
-        papis.git.commit(
-            new_folder_path,
-            f"Rename from '{folder}' to '{new_name}'")
+        papis.git.mv_and_commit_resource(
+            folder, new_folder_path,
+            f"Rename '{folder}' to '{new_folder_name}'")
     else:
         shutil.move(folder, new_folder_path)
 
     db.delete(document)
-    logger.debug("New document folder: '%s'.", new_folder_path)
     document.set_folder(new_folder_path)
     db.add(document)
 
--- a/papis/commands/rename.py
+++ b/papis/commands/rename.py
@@ -128,6 +128,12 @@ def cli(query: str,
         doc_folder: Tuple[str, ...],
         sort_reverse: bool) -> None:
     """Rename document folders"""
+    if not folder_name:
+        logger.warning("No folder name format specified, so no documents can be "
+                       "renamed. Set either the configuration option "
+                       "'add-folder-name' or use the '--folder-name' flag.")
+        return
+
     documents = papis.cli.handle_doc_folder_query_all_sort(query,
                                                            doc_folder,
                                                            sort_field,
--- a/papis/commands/rename.py
+++ b/papis/commands/rename.py
@@ -13,7 +13,7 @@ Examples
 
         papis rename author:"Rick Astley"
 
-- Rename picked folders that match a query and automatically ``--regenerate``
+- Rename picked folders that match a query and automatically ``--folder-name``
   the names following the pattern provided by the ``add-folder-name``
   configuration option. It will ask for confirmation before each rename. It
   doesn't slugify the names, so if the pattern results in "Rick Astley - Never
@@ -21,7 +21,7 @@ Examples
 
    .. code:: sh
 
-        papis rename --regenerate author:"Rick Astley"
+        papis rename --folder-name author:"Rick Astley"
 
 - Rename folders without asking for confirmation using ``--batch``.
 
@@ -115,9 +115,11 @@ def prepare_run(operations: List[Tuple[papis.document.Document, str]],
 
 
 @click.command("rename")
+@click.option(
+    "--folder-name",
+    help="Name for the document's folder (papis format)",
+    default=lambda: papis.config.getstring("add-folder-name"))
 @papis.cli.bool_flag("--batch", "-b", default=False, help="Batch mode, do not prompt")
-@papis.cli.bool_flag("--regenerate", "-r", default=False,
-                     help="Regenerate the folder name from the configured patttern")
 @click.help_option("--help", "-h")
 @papis.cli.all_option()
 @papis.cli.query_argument()
@@ -126,7 +128,7 @@ def prepare_run(operations: List[Tuple[papis.document.Document, str]],
 @papis.cli.doc_folder_option()
 def cli(query: str,
         git: bool,
-        regenerate: bool,
+        folder_name: str,
         _all: bool,
         batch: bool,
         sort_field: Optional[str],
@@ -143,25 +145,16 @@ def cli(query: str,
         logger.warning(papis.strings.no_documents_retrieved_message)
         return
 
-    if regenerate:
-        folder_name_pattern = papis.config.getstring("add-folder-name")
-
     renames = []
     for document in documents:
         current_name = document.get_main_folder_name()
-        if regenerate:
-            new_name = papis.format.format(folder_name_pattern, document)
-            new_name = papis.utils.clean_document_name(new_name)
+        new_name = papis.format.format(folder_name, document)
+        new_name = papis.utils.clean_document_name(new_name)
 
-            if batch:
-                logger.info("Renaming '%s' into '%s'", current_name, new_name)
-            else:
-                papis.tui.utils.confirm(f"Rename {current_name} into {new_name}?", True)
+        if batch:
+            logger.info("Renaming '%s' into '%s'", current_name, new_name)
         else:
-            new_name = papis.tui.utils.prompt(
-                "Enter new folder name:\n"
-                ">",
-                default=current_name or "")
+            papis.tui.utils.confirm(f"Rename {current_name} into {new_name}?", True)
 
         renames.append((document, new_name))
 
--- a/papis/commands/rename.py
+++ b/papis/commands/rename.py
@@ -120,7 +120,6 @@ def prepare_run(operations: List[Tuple[papis.document.Document, str]],
 @papis.cli.bool_flag("--batch", "-b", default=False, help="Batch mode, do not prompt")
 @papis.cli.bool_flag("--regenerate", "-r", default=False,
                      help="Regenerate the folder name from the configured patttern")
-@papis.cli.bool_flag("--slugify", "-s", default=False, help="Slugify the folder name")
 @click.help_option("--help", "-h")
 @papis.cli.all_option()
 @papis.cli.query_argument()
@@ -132,7 +131,6 @@ def cli(query: str,
         regenerate: bool,
         _all: bool,
         batch: bool,
-        slugify: bool,
         sort_field: Optional[str],
         doc_folder: Tuple[str, ...],
         sort_reverse: bool) -> None:
@@ -155,8 +153,7 @@ def cli(query: str,
         current_name = document.get_main_folder_name()
         if regenerate:
             new_name = papis.format.format(folder_name_pattern, document)
-            if slugify:
-                new_name = papis.utils.clean_document_name(new_name)
+            new_name = papis.utils.clean_document_name(new_name)
 
             if batch:
                 logger.info("Renaming '%s' into '%s'", current_name, new_name)
--- a/papis/commands/rename.py
+++ b/papis/commands/rename.py
@@ -55,7 +55,7 @@ Command-line Interface
     :prog: papis rename
 """
 
-from typing import Optional, Tuple, List
+from typing import Optional, Tuple
 
 import click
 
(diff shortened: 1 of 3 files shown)


kiike:
While the new features work fine during testing with my live library, they don't behave properly during automatic testing. I am trying to test via the CLI to improve coverage and parse the logs but the behaviour of `run` is very strange, where half of the operations fail.

tests/commands/test_rename.py
```diff
@@ -20,3 +19,37 @@ def test_rename_run(tmp_library: TemporaryLibrary) -> None:
     doc, = db.query_dict({"title": old_title})
     assert doc.get_main_folder_name() == new_name
     assert os.path.exists(doc.get_main_folder())
+
+
+def test_regenerate_name(tmp_library: TemporaryLibrary,
+                         tmp_config: TemporaryConfiguration,
+                         ) -> None:
```
alexfikl, line 26: Not sure what's wrong with the tests, but from a quick skim, this should just have a `TemporaryLibrary` (that inherits from `TemporaryConfiguration`). Maybe having two confuses the tests?

tests/commands/test_rename.py
```diff
@@ -20,3 +19,37 @@ def test_rename_run(tmp_library: TemporaryLibrary) -> None:
...
+    from click.testing import CliRunner
```
alexfikl, line 27: There should be a `papis.testing.PapisRunner` that wraps `CliRunner` and sets `mix_stderr=False`, so it's better to use that.

kiike: TIL! Thanks, I'll use that!

tests/commands/test_rename.py
kiike: Could be! We'll se on the next run :thinking:

kiike: So I am looping over the picked documents. That list is shuffling after each run :open_mouth:

kiike: There is something strange when running under Pytest and I'm unable to figure out what makes the list mutate after each run. It was as if the document list was being sorted within each iteration. But, as the level of abstraction is now so high, I just can't understand why.

kiike:
I went ahead and added some more tests too. It's at 98% coverage now.

papis/commands/rename.py
```diff
@@ -54,30 +109,65 @@ def run(document: papis.document.Document,
...
+                logger.info("Renaming '%s' into '%s'", current_name, new_name)
+            else:
+                papis.tui.utils.confirm(f"Rename {current_name} into {new_name}?", True)
+        else:
+            new_name = papis.tui.utils.prompt(
+                "Enter new folder name:\n"
+                ">",
+                default=current_name or "")
+
+        renames.append((document, new_name))
```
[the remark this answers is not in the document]
alexfikl: Just noticed this is handled in `run` somewhat. Not sure what the best way would be here..

alexfikl (commented):
Thank you for working on this! :heart: This is definitely a nice thing to have, since rename was very limited before, but the interface needs a bit of work.

I've left a few comments with suggestions and questions.

kiike:
> Thank you for working on this! ❤️ This is definitely a nice thing to have, since rename was very limited before, but the interface needs a bit of work.
> 
> I've left a few comments with suggestions and questions.

Thank you very much for your feedback! Let me address the changes! If you feel brave, you can check if it fails on your end too **in pytest** if you don't have `prepare_run`. Yesterday, when I tried it, the document list was being manipulated while running and I couldn't trace the reason.

kiike:
> First of all, great contribution! ❤️
> 
> I just started reviewing this but then I realised that things are changing under my nose and maybe it's not quite yet the moment for detailed feedback on the documentation. I'm happy to take a look at the docs again once things stabilise 😅.
> 
> So I just have the following comments:
> 
>     * I noticed a warning when pressing enter to accept the default name. I think that's a bit misleading as it's the expected behaviour.
> 
>     * I agree with @alexfikl that simplifying the command by removing the slugify flag is a good idea (but I just saw that you've already changed this! 🎉).
> 
>     * I also like the proposal regarding making `--regenerate` into `--folder-name` with a functionality in line with other commands
> 
>     * I think we should note somewhere that the `--batch` flag only does something in conjunction with the `--regenerate` flag (mostly a note to myself for the later docs feedback)

Hey Julien! Thanks!

> I just started reviewing this but then I realised that things are changing under my nose and maybe it's not quite yet the moment for detailed feedback on the documentation. I'm happy to take a look at the docs again once things stabilise 😅.

Yeah, sorry, I am working it as we speak. I'll leave a "ready for review" tag when done. This will be ready quickly, I am finishing the tweaks to the tests.

>     I noticed a warning when pressing enter to accept the default name. I think that's a bit misleading as it's the expected behaviour.

I'll check if I receive such warnings. But that won't happen as now there is no prompt anymore, just asking for confirmation unless `--batch` is used.

>     I agree with @alexfikl that simplifying the command by removing the slugify flag is a good idea (but I just saw that you've already changed this! 🎉).

I agree!

>     I also like the proposal regarding making --regenerate into --folder-name with a functionality in line with other commands

Sure. Not a huge fan of `--folder-name` since what you're providing is a pattern. Unless the pattern is actually a name (doesn't include variables). Then it just gets confusing.

>     I think we should note somewhere that the --batch flag only does something in conjunction with the --regenerate flag (mostly a note to myself for the later docs feedback)

The `--batch` flag can wreck your whole library if your pattern is malformed. As for if it does something with the new `--folder-name` flag, that is not actually the case anymore (when I push the changes): if you provide the flag, then it expects you provide a format following it. If you don't provide it, the pattern is pulled from `add-folder-name`. Your folders will be renamed following that format asking for confirmation by default, or without if you use `--batch`. (or I misunderstood what was required)

kiike:
@jghauser you can test it now if you want!

alexfikl:
@kiike I looked through the code a bit and moved some things around. It all looks good to me now! :tada: 
Let me know if you spot anything fishy in what I added!

> If you feel brave, you can check if it fails on your end too in pytest if you don't have prepare_run. Yesterday, when I tried it, the document list was being manipulated while running and I couldn't trace the reason.

Yeah, it fails on my system too :cry: I looked a bit, but not sure why it's doing that.. best guess is that the documents in the database are stored in a mutable list and mutable references are returned on the `query`. We can leave it as is for now!

> EDIT: I don't know if that will affect you, but I am going to test pushing a couple tweaks for the `git` section in `run`. It fails on real Windows. No `mv` command there by default, so I'll use `shutil`. 

That's a good change either way! We should be using the more platform independent version from `shutil`.

> Sure. Not a huge fan of --folder-name since what you're providing is a pattern. Unless the pattern is actually a name (doesn't include variables). Then it just gets confusing.

I'm not sure I understand why you're not a fan of the `--folder-name`?

alexfikl (approved):
The code all looks good to me now!

@jghauser If you get the chance to look over the docs :grin:

kiike:
> The code all looks good to me now!
> 
> @jghauser If you get the chance to look over the docs 😁

Yay! Thanks for reviewing it! Let's see what @jghauser thinks!

jghauser:
I was just looking through the docs and writing up some suggestions regarding the `--folder-name` option. I think the suggestions by @alexfikl and I might have ended up messing things up a bit :sweat_smile:.
Before, running the command without `--regenerate` would give you the manual renaming option with the existing name as the default. Now, the command without `--folder-name` will rename things according to the settings in `add-folder-name`, and so it's really the same the `--regenerate` option (because by default that's what it's set to). Is that expected? I preferred the earlier functionality (especially because `add-folder-name` is empty by default so that the default folder name is an empty string).

I can see two solutions:
1. Having a `--regenerate` and a `--folder-name` option and never set `--folder-name` to `add-folder-name`
2. Retain the current behaviour, but set the default folder name to the previous name if `add-folder-name` isn't defined

What do you think?

alexfikl:
> I can see two solutions:
> 
> 1. Having a `--regenerate` and a `--folder-name` option and never set `--folder-name` to `add-folder-name`
> 2. Retain the current behaviour, but set the default folder name to the previous name if `add-folder-name` isn't defined
>  
> What do you think?

Oops, I forgot about the `--regenerate` flag. I vote for the second option, maybe with a little warning that they need to set the option if it's not set or the rename won't do anything.

For comparison, in `papis add`, if it gets an empty folder from `add-folder-name`, it falls back to `MD5HASH-author-something`.

jghauser:
> Oops, I forgot about the `--regenerate` flag! I vote for the second option, maybe with a little warning that they need to set the option if it's not set or the rename won't do anything.
> 
> For comparison, in `papis add`, if it gets an empty folder from `add-folder-name`, it falls back to `MD5HASH-author-something`.

Yes that seems sensible. It would keep the amounts of flags small and is in line with how `papis add` works!

jghauser:
I've finished up with the docs (assuming the functionality just discussed). I'm just gonna post the whole thing here as that's easier (especially cause the code might still change a bit) and I'll be away for a few days over the weekend. I've added a small paragraph after the intro sentence, because I think that's useful information specific to this command (and applicable to almost all examples below).

```rst
The ``rename`` command is used to rename document folders based on user input,
a provided formatting pattern, or :confval:`add-folder-name`.

It will (except when run with ``--all``) bring up the picker with a list of
documents that match the query. In the picker, you can select one or more
documents and then initiate renaming by pressing enter. Folder names are cleaned,
so that various characters (white spaces, punctuation, capital letters, and some
others) are automatically converted.

Examples
^^^^^^^^

- Rename folders for documents whose author is "Rick Astley". You can then either
  enter a new folder name or accept Papis' suggestion. The suggested folder name
  will be generated according to :confval:`add-folder-name` or, if this option
  isn't set, the current folder name.

   .. code:: sh

        papis rename author:"Rick Astley"

- You can use ``--folder-name`` to pass in your desired name. This option
  supports Papis formatting patterns. You will be asked for confirmation
  before the folder is renamed.

   .. code:: sh

        papis rename --folder-name "{doc[author]}-never-gonna" author:"Rick Astley"

  This will give you a folder named "rick-astley-never-gonna".


- To stop Papis from asking for confirmation, use the ``--batch`` flag:

   .. code:: sh

        papis rename --batch --folder-name author:"Rick Astley"

- If you want to rename all documents without narrowing down your selection in
  the picker, you can use the ``--all`` flag. Be careful when combining this
  with ``--batch``, as you might end up renaming a lot folders without
  confirmation:

   .. code:: sh

        papis rename --all --folder-name author:"Rick Astley"

```

kiike:
> @kiike I looked through the code a bit and moved some things around. It all looks good to me now! 🎉 Let me know if you spot anything fishy in what I added!

Not absolutely sure because I am not looking closely at the `git` code, but you requested me to use `mv_and_commit_resource` in #816 but would that work in both following scenarios?

1.
   - file is tracked by git.
   - file is moved via `papis mv` (using `papis.utils.git.mv`)
   - git makes an automatic commit with a templated message, but we're commiting anyway.
   - is the repo clean?

 2.
    - file is untracked
    - file is moved via `papis mv`
    - a commit is done
    - repo is clean (haven't tested yet)
 
> Yeah, it fails on my system too 😢 I looked a bit, but not sure why it's doing that.. best guess is that the documents in the database are stored in a mutable list and mutable references are returned on the `query`. We can leave it as is for now!

Yeah, I couldn't pinpoint the source of the problem since that only happened during pytest. We could deepcopy `documents`, but I'm not sure if that will be helpful at all. Now that you are pushing to this branch, you can do it, or I can. Advise please :smile:.

> That's a good change either way! We should be using the more platform independent version from `shutil`.
Good! I'll keep an eye for more raw commands!

> I'm not sure I understand why you're not a fan of the `--folder-name`?

If it was `folder-pattern` or `folder-format` or equivalent that'd be clearer for the purpose of that flag. Especially since it says in the help `folder-name` (Papis format). Now that I think of it, another unintuitive config option is `header-format`, since that's not for a header but for an entry in the picker.

kiike:
I'm adding the docstring shared by @jghauser. Thanks for the edits @alexfikl and @jghauser! :heart:

alexfikl:
> would that work in both following scenarios?

As far as I can know, the the git code mostly assumes that it's all well set up and all the files exist and are tracked, so it's pretty brittle. On the bright side, it's all pretty well separated in `papis.git`, so we can make it smarter if needed.
 
> Yeah, I couldn't pinpoint the source of the problem since that only happened during pytest. We could deepcopy `documents`, but I'm not sure if that will be helpful at all. Now that you are pushing to this branch, you can do it, or I can. Advise please 😄.

Doing a deep copy seems very expensive, so I'm not a big fan. For now this is fine. Maybe open an issue about it so we keep track?

> If it was `folder-pattern` or `folder-format` or equivalent that'd be clearer for the purpose of that flag. Especially since it says in the help `folder-name` (Papis format). Now that I think of it, another unintuitive config option is `header-format`, since that's not for a header but for an entry in the picker.

Ah, fair enough! Completely agree on all of those points! We should probably go over those at some point and choose slightly better names, but it's tedious, so not sure it's gonna be on the top of the list :(

For now, I'm thinking consistency between commands is at least good to have.

alexfikl:
This is good to go from my end! @kiike Everything ok for you?

kiike:
> > would that work in both following scenarios?
> 
> As far as I can know, the the git code mostly assumes that it's all well set up and all the files exist and are tracked, so it's pretty brittle. On the bright side, it's all pretty well separated in `papis.git`, so we can make it smarter if needed.

Got it! I am not using git for my library but probably should. I'll give it a go sometime!

> 
> > Yeah, I couldn't pinpoint the source of the problem since that only happened during pytest. We could deepcopy `documents`, but I'm not sure if that will be helpful at all. Now that you are pushing to this branch, you can do it, or I can. Advise please 😄.
> 
> Doing a deep copy seems very expensive, so I'm not a big fan. For now this is fine. Maybe open an issue about it so we keep track?

Sure, I'll create an issue for that! Spelunking into the code to see why it's being mutated will be interesting!
 
> > If it was `folder-pattern` or `folder-format` or equivalent that'd be clearer for the purpose of that flag. Especially since it says in the help `folder-name` (Papis format). Now that I think of it, another unintuitive config option is `header-format`, since that's not for a header but for an entry in the picker.
> 
> Ah, fair enough! Completely agree on all of those points! We should probably go over those at some point and choose slightly better names, but it's tedious, so not sure it's gonna be on the top of the list :(

I agree. Naming is hard :smile: 
 
> For now, I'm thinking consistency between commands is at least good to have.

Definitely!

> This is good to go from my end, so we can get it in!
> @kiike Everything ok for you?

I think so! It has received a good makeover. I am happy with the functionality now and the features added by you too regarding git!

alexfikl:

PR #811, opened 2024-04-10
tests(hayagriva): rename script

kiike:
When I touched the Typst exporter, I noticed the test file had a wrong name: ha**r**agriva instead of ha**y**agriva. But I forgot to submit the change. This trivial PR fixes this.


alexfikl:

PR #813, opened 2024-04-12
ci: use uv

kiike:
Hi! This is a companion PR to the discussion in #812. I have been using `uv` for some months now and would like to pitch for inclusion in the CI pipelines here.

With this PR, we can significantly cut the time that it takes to install dependencies during CI. For instance, with the current setup, installing deps on Ubuntu takes around 20 seconds. With this PR it would be around 3-6 seconds. In Windows, installing the deps takes around one minute. With this PR, the time is halved too, resulting in around 20 seconds.

For this, a 3rd party action is used. I have personally reviewed the code there and it doesn't do anything sketchy.

If you want to see it in action with Papis, it's been running [on my fork](https://github.com/kiike/papis/actions/workflows/main.yml).

When testing this PR I also found a minor bug that happened on Windows during testing in CI. The venv with uv is not global, so `test_edit_run` fails because the mock `sed` program does not have the necessary `papis` dependency since it looks in the global venv. With this PR, the interpreter is assured to be the one that runs during CI (using `sys.executable`), having access to all the runtime dependencies.

### DIFF
--- a/tests/commands/test_edit.py
+++ b/tests/commands/test_edit.py
@@ -1,4 +1,5 @@
 import os
+import sys
 import pytest
 
 import papis.database
@@ -9,7 +10,7 @@ script = os.path.join(os.path.dirname(__file__), "scripts.py")
 
 
 @pytest.mark.library_setup(settings={
-    "editor": "python {} sed".format(script)
+    "editor": "{} {} sed".format(sys.executable, script)
     })
 def test_edit_run(tmp_library: TemporaryLibrary) -> None:
     import papis.config
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,6 +25,11 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
+    - name: Install uv
+      uses: yezz123/setup-uv@v4
+      with:
+        uv-venv: "venv"
+
     - name: Install dependencies
       run: make ci-install
       shell: bash
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -24,11 +24,16 @@ jobs:
       with:
         python-version: "3.12"
 
+    - name: Install uv
+      uses: yezz123/setup-uv@v4
+      with:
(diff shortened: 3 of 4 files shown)


alexfikl:
I'm personally not very interested in this, so I'll let others make the call.
* For the CI, our whole test suite is already very fast, under 2-3 minutes, so I'm not bothered with saving 15 seconds.
* Locally, I have a venv from over a year ago, so making the `pip install --editable .` faster will save me 15 seconds over the span of a couple of years.

tests/commands/test_edit.py
```diff
@@ -9,7 +10,7 @@
 
 
 @pytest.mark.library_setup(settings={
-    "editor": "python {} sed".format(script)
+    "editor": "{} {} sed".format(sys.executable, script)
```
alexfikl, line 13: This fix is definitely useful! If you want to put it in another PR, we can get it in right away.

kiike: Will do. After this, I am preparing a PR that allows for usage in vanila windows installations (no unix commands there). They will come in handy when releasing!

kiike:
> I'm personally not very interested in this, so I'll let others make the call. To motivate a bit:
> 
>     * For the CI, our whole test suite is already very fast, under 2-3 minutes, so I'm not bothered with saving 15 seconds.
> 
>     * Locally, I have a venv from over a year ago, so making the `pip install --editable .` faster will save me 15 seconds over the span of a couple of years.
> 
>     * Plus `uv` is so new..

Thanks for the feedback! No problem :smile: . This is targeted to CI to avoid wasting time before the tests run, so the tests actually start earlier instead of having to wait unnecessarily for pip install the dependencies (which is especially true for the windows runners, that use more than a minute to install the packages). This is no way would change or influence how you prepare your venv.

alexfikl:
> This is no way would change or influence how you prepare your venv.

That's true, but once it makes it into the CI scripts and whatnot, we'll have to support it properly.

PR #814, opened 2024-04-12
ci: use ruff for linting

kiike:
HI again!

Here is a companion PR for the discussion in #812.

It adds linting with ruff, which catches many more style "issues". Along getting rid of many dependencies, it provides faster linting and formatting, which can be pretty cool when used with pre-commit, for instance.

I tried to port some of the excluded rules.

WIP

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,16 +88,11 @@ Repository = "https://github.com/papis/papis"
 
 [project.optional-dependencies]
 develop = [
-    "flake8",
-    "flake8-bugbear",
-    "flake8-quotes",
-    "Flake8-pyproject",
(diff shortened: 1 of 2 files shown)


alexfikl:
I'm on board with switching to `ruff`, but probably after we make a release with the current stuff. It's definitely due :grin:

jghauser:
I'm definitely on board with using `ruff`. I've been using it on my side for a while and it would be nice to have some way of easily enforcing a nice consistent style :).

kiike:
Cool! We'll keep this PR on the backburner!

alexfikl:
Closing this in favor of #971.

PR #815, opened 2024-04-12
tests/mock_sed: use same interpreter as the one `pytest` is running

kiike:
When testing #813 I found a minor bug that happened on Windows during testing in CI or in my machine with a vanilla Windows installation.

The venv with uv is not global, so `test_edit_run` fails because the mock sed program does not have the necessary papis dependency, not being aware of the local venv.

With this PR, the interpreter is made to be the one that runs during CI (using `sys.executable`), having access to all the runtime dependencies.

### DIFF
--- a/tests/commands/test_edit.py
+++ b/tests/commands/test_edit.py
@@ -1,4 +1,5 @@
 import os
+import sys
 import pytest
 
 import papis.database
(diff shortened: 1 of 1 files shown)


alexfikl:
Nice catch! :rocket:

kiike:
Thank you!

PR #816, opened 2024-04-12
tests(windows): fix remaining "unixisms" that break some tests

kiike:
When I was developing an earlier PR, I wanted to test it on a vanilla installation of Windows, just to make sure no config could pollute the results.

Turns out, some tests fail on this kind of setup. Even with a benign but creepy Windows error popup during `tests/commands/open`.
image

Some reasons for the failing tests:

- symlinks no functional unless "developer mode is active" (I guess CI never sees this because it's running a modified build)
- typical unix commands such as `mv`, `echo` or `mv` not present in vanilla Windows.

Some details of this PR:

- It actually contains a fix that is present on a commit of #810, so I'll remove it from there.
- I want to explain the user why symlinks don't work if we get an exception.
- I'm sad to make `["git"] if git else []) + ["mv", folder, new_folder_path]` go, being so clever. I hope using `shutils` is OK.
- the add via symlink test is marked as xfail, just in case it succeeds.
- `script.py` now has `ls` too.

Looking forward to your feedback! I think this PR getting in will make it a bit more polished when a new user tries the 0.14 release on Windows, wants to contribute, and tests pass on a fresh install!

### DIFF
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -90,7 +90,6 @@ Command-line Interface
 
 import os
 import re
-import sys
 from typing import List, Any, Optional, Dict, Tuple
 
 import click
@@ -418,17 +417,7 @@ def run(paths: List[str],
 
         if link:
             in_file_abspath = os.path.abspath(in_file_path)
-            try:
-                logger.info("[SYMLINK] '%s' to '%s'.", in_file_abspath,
-                            tmp_end_filepath)
-                os.symlink(in_file_abspath, tmp_end_filepath)
-            except OSError as e:
-                missing_privilege_error = 1314
-                if sys.platform == "win32" and e.winerror == missing_privilege_error:
-                    logger.error("Failed to link due to insufficient permissions.  You "
-                                 "can try again after enabling the 'Developer mode' "
-                                 "and restarting.")
-                raise
+            papis.utils.symlink(in_file_abspath, tmp_end_filepath)
         elif move:
             logger.info("[MV] '%s' to '%s'.", in_file_path, tmp_end_filepath)
             shutil.copy(in_file_path, tmp_end_filepath)
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -90,6 +90,7 @@ Command-line Interface
 
 import os
 import re
+import sys
 from typing import List, Any, Optional, Dict, Tuple
 
 import click
@@ -417,8 +418,17 @@ def run(paths: List[str],
 
         if link:
             in_file_abspath = os.path.abspath(in_file_path)
-            logger.info("[SYMLINK] '%s' to '%s'.", in_file_abspath, tmp_end_filepath)
-            os.symlink(in_file_abspath, tmp_end_filepath)
+            try:
+                logger.info("[SYMLINK] '%s' to '%s'.", in_file_abspath,
+                            tmp_end_filepath)
+                os.symlink(in_file_abspath, tmp_end_filepath)
+            except OSError as e:
+                missing_privilege_error = 1314
+                if sys.platform == "win32" and e.winerror == missing_privilege_error:
+                    logger.error("Failed to link due to insufficient permissions.  You "
+                                 "can try again after enabling the 'Developer mode' "
+                                 "and restarting.")
+                raise
         elif move:
             logger.info("[MV] '%s' to '%s'.", in_file_path, tmp_end_filepath)
             shutil.copy(in_file_path, tmp_end_filepath)
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -417,6 +417,7 @@ def run(paths: List[str],
 
         if link:
             in_file_abspath = os.path.abspath(in_file_path)
+            logger.info("[SYMLINK] '%s' to '%s'.", in_file_abspath, tmp_end_filepath)
             papis.utils.symlink(in_file_abspath, tmp_end_filepath)
         elif move:
             logger.info("[MV] '%s' to '%s'.", in_file_path, tmp_end_filepath)
--- a/papis/commands/addto.py
+++ b/papis/commands/addto.py
@@ -98,8 +98,7 @@ def run(document: papis.document.Document,
 
         if link:
             in_file_abspath = os.path.abspath(in_file_path)
-            logger.debug("[SYMLINK] '%s' to '%s'.", in_file_abspath, out_file_path)
-            os.symlink(in_file_abspath, out_file_path)
+            papis.utils.symlink(in_file_abspath, out_file_path)
         else:
             import shutil
             logger.info("[CP] '%s' to '%s'.", local_in_file_path, out_file_path)
--- a/papis/commands/addto.py
+++ b/papis/commands/addto.py
@@ -98,6 +98,7 @@ def run(document: papis.document.Document,
 
         if link:
             in_file_abspath = os.path.abspath(in_file_path)
+            logger.info("[SYMLINK] '%s' to '%s'.", in_file_abspath, out_file_path)
             papis.utils.symlink(in_file_abspath, out_file_path)
         else:
             import shutil
--- a/papis/commands/mv.py
+++ b/papis/commands/mv.py
@@ -34,8 +34,12 @@ def run(document: papis.document.Document,
     if not folder:
         raise DocumentFolderNotFound(papis.document.describe(document))
 
-    papis.utils.run((["git"] if git else []) + ["mv", folder, new_folder_path],
-                    cwd=folder)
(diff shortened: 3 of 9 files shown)


papis/commands/add.py
```diff
@@ -417,8 +418,17 @@ def run(paths: List[str],
...
+                logger.info("[SYMLINK] '%s' to '%s'.", in_file_abspath,
+                            tmp_end_filepath)
+                os.symlink(in_file_abspath, tmp_end_filepath)
+            except OSError as e:
+                missing_privilege_error = 1314
+                if sys.platform == "win32" and e.winerror == missing_privilege_error:
+                    logger.error("Failed to link due to insufficient permissions.  You "
+                                 "can try again after enabling the 'Developer mode' "
+                                 "and restarting.")
+                raise
```
alexfikl, line 431: Does the exception not contain any useful information here? We usually do
```
logger.error(..., exc_info=exc)
```
to show the exception message or a traceback.

alexfikl, line 431: If this really needs special handling for windows, it would be better to make it a function in `papis.utils` or something so that we can reuse it in other places (`papis addto` comes to mind).

papis/commands/mv.py
```diff
@@ -34,8 +34,12 @@ def run(document: papis.document.Document,
...
         raise DocumentFolderNotFound(papis.document.describe(document))
 
-    papis.utils.run((["git"] if git else []) + ["mv", folder, new_folder_path],
-                    cwd=folder)
+    if git:
+        papis.utils.run(["git", "mv", folder, new_folder_path],
+                        cwd=folder)
+    else:
+        import shutil
+        shutil.move(folder, new_folder_path)
```
alexfikl, line 42: This should probably use the `mv_and_commit_resource` thing introduced in #810.

tests/commands/test_add.py
```diff
@@ -189,6 +190,7 @@ def test_add_set_cli(tmp_library: TemporaryLibrary) -> None:
     assert not doc.get_files()
 
 
+@pytest.mark.xfail(sys.platform == "win32", reason="Developer mode not enabled")
```
alexfikl, line 193: Just curious, but can this be checked in any way? I'm thinking we could do a `pytest.skip` instead.

tests/commands/test_edit.py
```diff
@@ -1,16 +1,22 @@
...
+
 import pytest
 
 import papis.database
 
 from papis.testing import TemporaryLibrary, PapisRunner
 
-script = os.path.join(os.path.dirname(__file__), "scripts.py")
+
+def get_mock_script(commandlet: str) -> str:
```
alexfikl, line 10:
```suggestion
def get_mock_script(name: str) -> str:
```
? I'm not sure I've seed `commandlet` used anywhere. Is that a thing? :))

tests/commands/test_open.py
```diff
@@ -45,5 +47,5 @@ def test_open_windows_cli(tmp_library: TemporaryLibrary) -> None:
 
     result = cli_runner.invoke(
         cli,
-        ["--tool", 'cmd.exe /c start ""', "--all", "Krishnamurti"])
+        ["--tool", "cmd.exe /c type", "--all", "Krishnamurti"])
```
alexfikl, line 50: Does "start" not work here? Or?

alexfikl (approved):
This is great! :tada: 

Left a few questions and nitpicks, but hopefully easy to fix.

papis/commands/add.py
alexfikl: I can't check, but does this throw just a generic `OSError`? Maybe we can better check for a `PermissionError` or something to show that message.

kiike:
It raises WinError, an exception that subclasses OSError. WinError can’t be in code that would run on Linux, since that’s undefined. I’ll probably move Windows specific stuff into its own utility function as you suggested so that we have specific WinError exceptions.

As for checking before actually running the operations, I guess that’s not impossible, but where should the check go? After installation, warning the user that functionality would be reduced due to dev mode not enabled, for instance?

Enric Morales ***@***.***>
________________________________

alexfikl:
> It raises WinError, an exception that subclasses OSError. WinError can’t be in code that would run on Linux, since that’s undefined. I’ll probably move Windows specific stuff into its own utility function as you suggested so that we have specific WinError exceptions.

According to the docs, `WindowsError` is an alias of `OSError` starting with Python 3.3, not a subclass of it. But yeah, it's not defined at all on non-Windows systems, so it can't be used.

> As for checking before actually running the operations, I guess that’s not impossible, but where should the check go? After installation, warning the user that functionality would be reduced due to dev mode not enabled, for instance?

I was just thinking for that specific test, to just add a little
```
if not is_win_dev_mode():
    pytest.skip("This test requires Developer Mode on Windows")
```
But it was mostly for curiosity. They way it is now is just fine!

kiike:
Blocked until #810 gets in.

papis/commands/add.py
```diff
@@ -417,8 +417,7 @@ def run(paths: List[str],
 
         if link:
             in_file_abspath = os.path.abspath(in_file_path)
-            logger.info("[SYMLINK] '%s' to '%s'.", in_file_abspath, tmp_end_filepath)
```
alexfikl, line 420: Can you leave this `logger.info` message?

papis/commands/addto.py
```diff
@@ -98,8 +98,7 @@ def run(document: papis.document.Document,
 
         if link:
             in_file_abspath = os.path.abspath(in_file_path)
-            logger.debug("[SYMLINK] '%s' to '%s'.", in_file_abspath, out_file_path)
```
alexfikl, line 101: Leave this logger message too and make it a `logger.info` to match `papis add`.

papis/utils.py
```diff
@@ -633,3 +633,26 @@ def is_relative_to(path: str, other: str) -> bool:
...
+
+    On Windows, Developer Mode can be enabled to allow unprivileged users to
+    allow creation of symlinks.  Failing to do so will raise an OSError exception,
+    with error code 1314.
+
+    :param source: the existing file that `destination` points to
+    :param destination: the name of the new symbolic link, pointing to `source`.
+    """
+    try:
+        logger.debug("[SYMLINK] '%s' to '%s'.", source, destination)
```
alexfikl, line 649:
```suggestion
```

papis/utils.py
```diff
@@ -633,3 +633,26 @@ def is_relative_to(path: str, other: str) -> bool:
...
+        os.symlink(source, destination)
+    except OSError as exc:
+        windows_missing_privilege_error = 1314
```
alexfikl, line 652: Maybe make this a global variable and call it `ERROR_PRIVILEGE_NOT_HELD = 1314` so that it matches the error code in the windows docs?

papis/utils.py
```diff
@@ -633,3 +633,26 @@ def is_relative_to(path: str, other: str) -> bool:
...
+        if sys.platform == "win32" and exc.winerror == windows_missing_privilege_error:
+            # https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--1300-1699-
+            logger.error("Failed to link due to insufficient permissions.  You "
+                         "can try again after enabling the 'Developer mode' "
+                         "and restarting.", exc_info=exc)
+            raise
```
alexfikl, line 658: Maybe just re-raise this with additional information?
```suggestion
            raise OSError(exc.errno, 
            			 "Failed to link due to insufficient permissions. You "
                         "can try again after enabling the 'Developer mode' "
                         "and restarting.", exc.filename, exc.winerror, exc.filename2)
```
Not sure this works, but something like that..

alexfikl (commented):
Left a few more suggestions, but this should be good to go once we can use the git stuff from #810.

kiike:
I'm checking the exception on windows, to see if it is raising with all arguments

alexfikl:
@kiike Looks good to me now! Ready to go?

kiike:
> @kiike Looks good to me now! Ready to go?

Yes! Let's do it! But if you want me to clean up the commit list so they are logically separated by what they want to do, I can do it to ease reverting/tracking when was something introduced. Just tell me or squash merge, I don't mind :smile:

alexfikl:
> > @kiike Looks good to me now! Ready to go?
> 
> Yes! Let's do it! But if you want me to clean up the commit list so they are logically separated by what they want to do, I can do it to ease reverting/tracking when was something introduced. Just tell me or squash merge, I don't mind 😄

I don't know, squashing should be fine. My "rule" is just that when there's some "fix this and that" commits from the review => squash it :grin:

Thank you for working on this! In it went :rocket:

PR #817, opened 2024-04-13
Add a pre-commit config file to ease catching style/linting issues

kiike:
As a companion PR to #812, I want to share a config file for a tool I try to use as much as possible, pre-commit. pre-commit executes hooks on staged files which aim to catch potential issues that would only arise after CI tests. When I use pre-commit, it alerts me when linting doesn't pass and rejects the commit. Many times I forgot to run `make flake8`, thinking I did because I actually ran `pytest`, and I would get an error during CI complaining about the linting.

This PR adds a file for pre-commit, along with instructions for opting to use it. The config file includes flake8 and mypy for now. There are other useful ones, such as the whitespace ones. Or the ruff one when we adopt ruff usage.

Its usage is completely opt-in. The config file file, along with the instructions, aim to help get pre-commit set up for this project. Distributing this file would be in line of what the same the [flake8 folks](https://github.com/PyCQA/flake8/blob/main/.pre-commit-config.yaml), [mypy](https://github.com/python/mypy/blob/master/.pre-commit-config.yaml), [flask](https://github.com/pallets/flask/blob/main/.pre-commit-config.yaml) or [pytest](https://github.com/pytest-dev/pytest/blob/main/.pre-commit-config.yaml) are doing.

What do you think?


.pre-commit-config.yaml
```diff
@@ -0,0 +1,34 @@
...
+        - pep8-naming
+
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: "v1.9.0"
+    hooks:
+      - id: mypy
+        args:
+          - --strict
+          - --show-column-numbers
+          - --pretty
```
alexfikl, line 26: Does this not pick up the settings in `pyproject.toml`?

alexfikl (commented):
Not a big fan of `pre-commit` because it just adds another place we need to keep up to date and make sure all the dev dependencies are listed. I'm not against it, but would prefer to wait until someone else voices some support :\

If you need this, you can very easily create a file `.git/hooks/pre-commit` with
```bash
#!/usr/bin/bash
python -m flake8 papis
python -m mypy papis
```
although the output isn't as nice, I guess.

kiike:
> Not a big fan of `pre-commit` because it just adds another place we need to keep up to date and make sure all the dev dependencies are listed. I'm not against it, but would prefer to wait until someone else voices some support :\
> 
> If you need this, you can very easily create a file `.git/hooks/pre-commit` with
> 
> ```shell
> #!/usr/bin/bash
> python -m flake8 papis
> python -m mypy papis
> ```
> 
> although the output isn't as nice, I guess.

That's OK. The output is not as nice but the end result is not the same. Thanks for the tip! I set this as draft for when somebody else wants to reply, but feel free to close this PR :smile:!

alexfikl:
I'll go ahead and close this, since nobody has shown an interest and we switched to `ruff` (so the PR would need a bunch of changes).

PR #820, opened 2024-04-14
picker: some minor improvements for marking entries

kiike:
One of the things that I found a bit unintuitive in the picker was that marked lines only had one marker on the first line. That is alright but I find it a bit more straightforward if the marker is applied to the whole entry, as many lines as that might span.

Also, while both selected and unselected margins can be configured, marked margins can not. I use the marker quite often, so i want a clearer picture of what I have marked.

This PR introduces:

- configurable marked line margin style: allows setting the style of a marked entry's margin image
- markers span the whole line range for an entry instead of just the first line. image

While the first point is straightforward, for the second feature I expanded the list comprehension into a `for` loop to make it a bit more readable. I could turn it into a list comprehension if desired.

These are super simple changes, but I think the end result looks a bit more intuitive. What do you think?

### DIFF
--- a/papis/tui/__init__.py
+++ b/papis/tui/__init__.py
@@ -15,6 +15,7 @@ def get_default_settings() -> PapisConfigType:
         "message_toolbar_style": "bg:ansiyellow fg:ansiblack",
         "options_list.selected_margin_style": "bg:ansiblack fg:ansigreen",
         "options_list.unselected_margin_style": "bg:ansiwhite",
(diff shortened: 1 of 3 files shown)


alexfikl:
I don't use marks that much, but this is very nice! In it goes :rocket:

kiike:
> I don't use marks that much, but this is very nice! In it goes 🚀

Thanks! Sometimes I use them for merging, or deleting! Maybe not everyday, but they help me a lot with my workflow :smile: Thanks @alexfikl!

PR #826, opened 2024-04-17
feat: add batch move with formatting of folder names

kiike:
The other day I was adding changes to `mv`, and I realised I never used it since I didn't understand how it was supposed to work. Also, compared to a file manager, maybe a command that can only handle one document at a time is very underpowered.

So, today I wanted to change it more or less in line with what I did with `rename`.

This PR implements:

- [x] picking multiple folders
- [x] using `--all` like in other documents
- [x] format subfolders names
- [ ] confirm before each move, and `--batch` to skip confirmation
- [x] prompting to add the changes if the library is version-controlled
- [x] documenting the changes for the docs
- [ ] adding the info to the changelog (unless this feature is rejected, of course, let me know!)

Let me show you what it does:
- before:
```
 papis -l test list -a
[INFO] database.cache: Indexing library. This might take a while...
/antoniou-2024-axialperturbationsofhairy
/dubinsky-2024-asymptoticdecayandquasinormal
/raju-2024-howdoesinformationemerge
/ray-2024-searchingforbinaryblack
/ressler-2024-blackhole-diskinteractionsin
/yuan-2024-primordialblackholeinterpretation
```
- after running `papis -l test mv -a --folder-name "{{doc.year}}"`:
```
 papis -l test list -a
[INFO] database.cache: Indexing library. This might take a while...
/2024/antoniou-2024-axialperturbationsofhairy
/2024/dubinsky-2024-asymptoticdecayandquasinormal
/2024/raju-2024-howdoesinformationemerge
/2024/ray-2024-searchingforbinaryblack
/2024/ressler-2024-blackhole-diskinteractionsin
/2024/yuan-2024-primordialblackholeinterpretation
```
- Using slashes inside the format nests the subfolders, as expected:
```
 papis -l test mv -a --folder-name "{{doc.year}}/{{doc.tags[0]}}"
[... boring output ... ]
 papis -l test list -a
[INFO] database.cache: Indexing library. This might take a while...
/2024/to_read/antoniou-2024-axialperturbationsofhairy
/2024/to_read/dubinsky-2024-asymptoticdecayandquasinormal
/2024/to_read/raju-2024-howdoesinformationemerge
/2024/for_the_next_paper/ray-2024-searchingforbinaryblack
/2024/for_the_next_paper/ressler-2024-blackhole-diskinteractionsin
/2024/for_the_next_paper/yuan-2024-primordialblackholeinterpretation
```

As for the code, I tried to make it a bit more verbose, since these operations are quite delicate. So there are some more checks such as `Document.get_main_folder()` and the like.

I am reusing the `--folder-name` flag, although it's not defaulting to `add-folder-name`. Do you find it confusing?

I am trying a new small feature: if the library is under version control, I think it might be a good idea to prompt the user if they haven't supplied the `--git` option. If you think this might be a good idea, maybe it would be cool to also have an option in the config file to always add the changes or to never ask. Let me know what you think!


alexfikl:
I've never used `papis mv`. What's the difference compared to `papis rename`?

kiike:
From what I understood, the purpose of each is a bit different: yout wan `mv` when a document folder is not in a directory structure you like and so you want to organise that. You would want `rename` to change the folder name. 

In other words, having this document:
MY_LIBRARY/foo/bar/RIckAstley-2009-NeverGonna/info.yaml
- rename would be used to act on "RIckAstley-2009-NeverGonna"
- mv for "foo/bar"

The functionality overlaps a lot, I mean, it's just `mv` under the hood. I guess you could use rename to nest a document inside other folders and the move would work but I don't know how Papis would react. Why it was originally separated, I don't know :shrug:. And whether having only mv for both operations is better or worse for UX, i don't know either. What do you think?

kiike:
I agree. I would prefer just having a single `mv` too. And possibly make `rename` handle the attached files (so the files listed in `files` inside the `info.yaml`, as I think there is nothing that can let you rename a file after importing it). But I don't know how it would work UX-wise.

For instance, with the "git-like" semantics, ie. a single command to move and rename, imagine you have two docs, so "alex-2004-foo" and "alex-2004-bar", I guess if you do `papis mv`, pick the two documents, and choose "2004" as the dest folder, it's clear what you want. I mean, you wouldn't want to collapse the two directories into a single one. So you create the 2004 folder and move the two docs there. Easy.

But if you do `papis mv`, choose one document, and  issue 2004 in the prompt, then it's not clear if you want to move it or rename. If you already have the 2004 folder, then I guess moving is what you want, mirroring what `mv` (posix program) does.

Just thinking out loud. How would you expect this all to work? I think we should invite other to the conversation.

papis/commands/mv.py
```diff
@@ -55,58 +105,92 @@ def run(document: papis.document.Document,
...
-
-    run(document, new_folder, git=git)
+                complete_while_typing=True)
+            dest_path = os.path.join(lib_dir, subfolder)
+        except Exception as exc:
+            logger.error("Failed to choose directory.", exc_info=exc)
+            return
+
+    if not git and os.path.exists(os.path.join(lib_dir, ".git")):
+        git = papis.tui.utils.confirm("Add the changes to git?")
```
alexfikl, line 160: This is new! Probably a good idea to add some checks like this in more places!

papis/commands/mv.py
```diff
@@ -55,58 +105,92 @@ def run(document: papis.document.Document,
...
+    for document in documents:
+        # If the user has passed ``--folder-name``, use that to generate the subfolder
+        if folder_name is not None:
+            subfolder = papis.format.format(folder_name, document)
+            dest_path = os.path.join(lib_dir, subfolder)
+
+        # This should never be reached, but just in case.
+        if dest_path is None:
+            logger.error("Failed to construct a folder path for the document.")
+            return
```
alexfikl, line 172:
```suggestion
        if not dest_path:
            logger.error("Failed to construct a folder path for the document: %s",
            			 papis.document.describe(document))
            continue
```
? To just skip the document? 

I'm never sure how to handle this, e.g. the update command will exit right away if it fails to update some key in some document, but most other commands will skip it and continue.

papis/commands/mv.py
```diff
@@ -55,58 +105,92 @@ def run(document: papis.document.Document,
...
+
+    if not git and os.path.exists(os.path.join(lib_dir, ".git")):
+        git = papis.tui.utils.confirm("Add the changes to git?")
+
+    moves = []
+    for document in documents:
+        # If the user has passed ``--folder-name``, use that to generate the subfolder
+        if folder_name is not None:
+            subfolder = papis.format.format(folder_name, document)
+            dest_path = os.path.join(lib_dir, subfolder)
```
alexfikl, line 167: If `folder_name` is None, shouldn't we just skip the whole thing? i.e. this check should be moved outside the loop and return.

Then here it should check `if subfolder`, i.e. if the formatting didn't end up making an empty string for whatever reason and warn about that.

papis/commands/mv.py
```diff
@@ -55,58 +105,92 @@ def run(document: papis.document.Document,
...
+            return
+
+    if not git and os.path.exists(os.path.join(lib_dir, ".git")):
+        git = papis.tui.utils.confirm("Add the changes to git?")
+
+    moves = []
+    for document in documents:
+        # If the user has passed ``--folder-name``, use that to generate the subfolder
+        if folder_name is not None:
+            subfolder = papis.format.format(folder_name, document)
```
alexfikl, line 166: This not correct, since the document keys can contain additional path separators. There's similar logic in `papis add` that we should refactor and reuse here.

For example, if the title is `Face/Off` and the folder format is `{doc[year]}/{doc[title]}` we'll get unexpected results.

papis/commands/mv.py
```diff
@@ -55,58 +105,92 @@ def run(document: papis.document.Document,
...
+        # This can happen when using format but a key doesn't exist and so the
+        # format expression evaluates to ''.
+        if dest_path.rstrip(os.path.sep) == os.path.dirname(doc_main_folder):
+            logger.warning("Skipping '%s': its source and destination are the same",
+                           doc_main_folder)
+            continue
+
+        if not os.path.exists(dest_path):
+            logger.info("Creating path '%s'.", dest_path)
+            os.makedirs(dest_path, mode=papis.config.getint("dir-umask") or 0o666)
```
alexfikl, line 188: Does this correctly check that we're not overwritting existing documents? I didn't see any checks in `run()` either.

alexfikl (changes requested):
This needs a bit more work and I left a few comments and suggestions. Probably worth leaving them alone until we decide what to do with `rename` vs `mv` though.

alexfikl:
> I agree. I would prefer just having a single `mv` too. And possibly make `rename` handle the attached files (so the files listed in `files` inside the `info.yaml`, as I think there is nothing that can let you rename a file after importing it). But I don't know how it would work UX-wise.

I think `papis update` can handle that these days.. but it doesn't rename the files on disk (I think?), so it's not exactly the complete solution.

> For instance, with the "git-like" semantics, ie. a single command to move and rename, imagine you have two docs, so "alex-2004-foo" and "alex-2004-bar", I guess if you do `papis mv`, pick the two documents, and choose "2004" as the dest folder, it's clear what you want. I mean, you wouldn't want to collapse the two directories into a single one. So you create the 2004 folder and move the two docs there. Easy.
> 
> But if you do `papis mv`, choose one document, and issue 2004 in the prompt, then it's not clear if you want to move it or rename. If you already have the 2004 folder, then I guess moving is what you want, mirroring what `mv` (posix program) does.
>
> Just thinking out loud. How would you expect this all to work? I think we should invite other to the conversation.

Maybe this can work similarly to how `papis add` works, which has both a `--subfolder` and a `--folder-name` option. Then `--subfolder` will always force a subfolder to be created, but `--folder-name` can also be `subfolder/my-doc-folder` if the subfolder already exists.

Would that make sense? Not sure if the behavior of `papis add` is the best, but at least they would work similarly. Do you think it makes more sense to have separate commands for the two cases?

papis/commands/mv.py
alexfikl: > This is new! Probably a good idea to add some checks like this in more places! (not now, just something to keep in mind)

papis/commands/mv.py
kiike: I don't think it does! It should check and skip the move, like `rename` (IIRC) does.

papis/commands/mv.py
kiike: Yes! And also special characters. I'll see what `add` does and adapt for here.

papis/commands/mv.py
kiike: Yes, I added it as a comment to ask for feedback. UX wise i think it might be a good idea if you forget to include `--git` but then maybe i'd be good to have options to skip asking, or don't ask and always commit to git? IDK.

papis/commands/mv.py
```diff
@@ -22,31 +72,31 @@
...
 
-    folder = document.get_main_folder()
-    if not folder:
+    doc_main_folder = document.get_main_folder()
+    if not doc_main_folder:
         raise DocumentFolderNotFound(papis.document.describe(document))
 
     if git:
-        papis.git.mv(folder, new_folder_path)
+        papis.git.mv(doc_main_folder, dest_path)
```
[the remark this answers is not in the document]
kiike: I think if the file is tracked it should commit, yes.

papis/commands/mv.py
```diff
@@ -1,5 +1,54 @@
...
+  organise all documents by author "Rick Astley" by year, you can use (Python and
+  Jinja2 formatting, respectively):
+
+   .. code:: sh
+
+        # Python format
+        papis mv --folder-name "{doc[year]}" --all author:"Rick Astley"
+
+        # Jinja2 format
+        papis mv --folder-name "{{doc.year}}" --all author:"Rick Astley"
```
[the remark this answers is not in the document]
kiike: Fair enough! I'll use this change then!

kiike:
Thanks you so much for the review!

> I think papis update can handle that these days.. but it doesn't rename the files on disk (I think?), so it's not exactly the complete solution.

Yes, it looks like it doesn't handle file renames, additions or deletions, as per the TODO comment there. You can change the files list but it won't do anything to the files. Plus you have to pass Python-style syntax, but with all the shell quoting needed. 

> Maybe this can work similarly to how `papis add` works, which has both a `--subfolder` and a `--folder-name` option. Then `--subfolder` will always force a subfolder to be created, but `--folder-name` can also be `subfolder/my-doc-folder` if the subfolder already exists.
> Would that make sense? Not sure if the behavior of `papis add` is the best, but at least they would work similarly. Do you think it makes more sense to have separate commands for the two cases?

To be honest, I don't know what would be better for a user. I would be perfectly comfortable having just `mv` that renames as needed. It would simplify the code and also less mental overhead (otherwise you want to (1) rename a folder and then (2) change the folder structure). But I don't know if that would be just me.

papis/commands/mv.py
alexfikl: Yeah, it needs a bit of work, so definitely for a different PR, but overall seems like a nice idea.

papis/commands/mv.py
alexfikl: I've tried to do some refactoring in #829 to extract that logic. Probably needs a bit more work, but it should be doable to use it here as well.

alexfikl:
> Plus you have to pass Python-style syntax, but with all the shell quoting needed.

I'm not sure I follow, what does "Python-style syntax" mean here? (I haven't actually used `papis update` to rename things, so I'm mostly just curious)
 
> > Maybe this can work similarly to how `papis add` works, which has both a `--subfolder` and a `--folder-name` option. Then `--subfolder` will always force a subfolder to be created, but `--folder-name` can also be `subfolder/my-doc-folder` if the subfolder already exists.
> > Would that make sense? Not sure if the behavior of `papis add` is the best, but at least they would work similarly. Do you think it makes more sense to have separate commands for the two cases?
> 
> To be honest, I don't know what would be better for a user. I would be perfectly comfortable having just `mv` that renames as needed. It would simplify the code and also less mental overhead (otherwise you want to (1) rename a folder and then (2) change the folder structure). But I don't know if that would be just me.

Hm, I don't think I was clear here, since it sounds to me like I agree with what you're saying.

What I was suggesting was a way to resolve some ambiguity that you mentioned (?) when doing a
`papis mv --folder-name "subfolder" <QUERY>`: it's not clear if you want to move the documents to that subfolder or rename the picked document.

In that case, I would suggest adding a `--subfolder` flag (mostly because `papis add` has one) so that:
* If you want to just rename folders or move them to some existing subfolder you can use `--folder-name`.
* If you want to move them to a subfolder, you can force that with `--subfolder`. 

I think that's similar to how the `-t/--target-directory` flag works in the Unix `mv` command. Does that make sense?

kiike:
> > Plus you have to pass Python-style syntax, but with all the shell quoting needed.
> 
> I'm not sure I follow, what does "Python-style syntax" mean here? (I haven't actually used `papis update` to rename things, so I'm mostly just curious)

I was just rambling about `papis update`. If you want to change the files of a document, you need to pass something like `--set files '["file1.pdf", "file2.pdf"]'` or something like that. That's what I meant with "python-style syntax".

> 
> > To be honest, I don't know what would be better for a user. I would be perfectly comfortable having just `mv` that renames as needed. It would simplify the code and also less mental overhead (otherwise you want to (1) rename a folder and then (2) change the folder structure). But I don't know if that would be just me.
> 
> Hm, I don't think I was clear here, since it sounds to me like I agree with what you're saying.
> 
> What I was suggesting was a way to resolve some ambiguity that you mentioned (?) when doing a `papis mv --folder-name "subfolder" <QUERY>`: it's not clear if you want to move the documents to that subfolder or rename the picked document.
> 
> In that case, I would suggest adding a `--subfolder` flag (mostly because `papis add` has one) so that:
> 
>     * If you want to just rename folders or move them to some existing subfolder you can use `--folder-name`.
> 
>     * If you want to move them to a subfolder, you can force that with `--subfolder`.
> 
> 
> I think that's similar to how the `-t/--target-directory` flag works in the Unix `mv` command. Does that make sense?

Yep! Having `subfolder` and `folder-name` sounds like something that removes a lot of the ambiguity. Would you like me to try to make this change that unifies both commands and aliases `rename` to `mv`? I would do this after the documentation, since I think the release might come first.

alexfikl:
> I was just rambling about `papis update`. If you want to change the files of a document, you need to pass something like `--set files '["file1.pdf", "file2.pdf"]'` or something like that. That's what I meant with "python-style syntax".

Ah, yeah, I didn't quite like that either: the command-line string is quite clunky (like in your example) and then you also need to do an `eval` on it..

Suggestions are very welcome!
 
> Yep! Having `subfolder` and `folder-name` sounds like something that removes a lot of the ambiguity. Would you like me to try to make this change that unifies both commands and aliases `rename` to `mv`? I would do this after the documentation, since I think the release might come first.

Yeah, that sounds good, we can definitely merge this with whatever fixes you have time to make from my comments above. It's already a pretty big improvement on the previous state of `papis mv`! :tada:

jghauser:
I finally managed to have a look at the conversation here!

> In other words, having this document: `MY_LIBRARY/foo/bar/RickAstley-2009-NeverGonna/info.yaml`
> 
>     * `rename` would be used to act on `RIckAstley-2009-NeverGonna`
> 
>     * `mv` for `foo/bar`

In general I think there is sense in having a distinction between `mv` and `rename` because they are different in the minds of everyone except very geeky people :smiley:. Rename means changing the file/folder name only and move means putting the file/folder into a different directory and I think that's roughly what the commands do (though I don't actually use either of them). So I'm ok with keeping both (with rename mostly being a wrapper around `mv`). It's a bit like `papis tag` is just a more convenient version of `papis update`. On the other hand, I do see the purpose of keeping the sub-command list short, so maybe getting rid of `rename` isn't so bad (also, for some reason, when I hear 'rename', I think of the attached files and not the folder)...

> > I think papis update can handle that these days.. but it doesn't rename the files on disk (I think?), so it's not exactly the complete solution.
> 
> Yes, it looks like it doesn't handle file renames, additions or deletions, as per the TODO comment there. You can change the files list but it won't do anything to the files. Plus you have to pass Python-style syntax, but with all the shell quoting needed.

Yeah, I don't think we need `papis rename` to handle files if there is `papis update` whose functionality could be extended (after all, you're gonna always want to rename the files in concert with the info in the `.yaml`).

Btw, you don't need to use `papis update --set WEIRDPYTHONSTRING`. You can use `papis update -p files FILE1 -p files FILE2` if you want to add two files (maybe with a `papis update --drop files` if you want to clear the list first) or `papis update --rename files OLDFILENAME NEWFILENAME` if you want to rename something. I mostly added the python string stuff in case someone wants to add e.g. a dictionary or something else that is otherwise not supported. If there's any core functionality that requires python syntax, we (I :sweat_smile:) should fix that. Maybe I should also add a note to the docs that this functionality is 'advanced'.

pmiam:
> Yeah, I don't think we need papis rename to handle files if there is papis update whose functionality could be extended (after all, you're gonna always want to rename the files in concert with the info in the .yaml).

This approach seems likely to resolve https://github.com/papis/papis/issues/894. I can start working to tackle some of the TODO items in `papis.update`. Any objections?

jghauser:
> > 
> 
> This approach seems likely to resolve #894. I can start working to tackle some of the TODO items in `papis.update`. Any objections?

Would be great if you could take this on! :heart: I didn't implement it at the time I re-wrote `papis update` since that was already a rather big undertaking. Some (random) considerations:
- it concerns the names of attached notes, files, and the folder itself (or folders, for users with a more hierarchical folder organisation). I think not everything needs to be dealt with in one PR though.
- for notes and files, there's also deletion (where we should probably tread extra carefully)
- we should definitely be careful about name collisions (and not overwrite without explicit 'yes do!' from the user)
- a option that does file renames automatically whenever `papis update` is called might be nice
- it would be nice (to the extend that it's possible) if the logic were setup in a way that `papis rename` and `papis mv` could re-use it (or the other way round)

pmiam:
I just had a thought about a potential interface redesign opportunity in `papis mv`

> * If you want to just rename folders or move them to some existing subfolder you can use `--folder-name`.
> 
> * If you want to move them to a subfolder, you can force that with `--subfolder`.

Despite my understanding why they're necessary/powerful, I'm not a fan of having two separate flags for this operation. I feel the `--folder-name` flag looks to the casual cli user like it should be able to do everything `--subfolder` can do. While, in fact, folder name is constrained to specify full-length document path name(s) and subfolder is meant only to prepend parent directory(ies) to existing paths.

I wonder if this is not better achieved using a skewmorphism of shell globbing syntax. i.e.

- `papis mv --subfolder parent "query"` --> `papis mv -t parent/* "query"`
  results in document paths matched by `query` to be moved under `parent` (may be a template)
- `papis mv --folder-name parent/newdoc "query"` --> `papis mv -t parent/sub/doc "query"`
  results in document paths matched by `query` e.g. `doc` to be moved and renamed to `/parent/newdoc`

I have no idea if this is feasible, but I do think it's more familiar and intuitive to command line users.

alexfikl:
Superseeded by #1104.

PR #831, opened 2024-04-23
feat: tweak document file naming

alexfikl:
This tries to be a bit smarter about adding suffixes to files in `papis add` and `papis addto`. Namely:
* Unifies the two with a new function `papis.paths.rename_document_files`.
* Makes files with different extensions not get a suffix, e.g. if one adds `file.pdf` and `file.md` it will now not rename them like `<something>.pdf` and `<something>-a.md` (with the `a` suffix).

### DIFF
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -254,7 +254,7 @@ def run(paths: List[str],
 
     import tempfile
 
-    in_documents_paths = paths
+    in_document_paths = paths
     temp_dir = tempfile.mkdtemp()
     tmp_document = papis.document.Document(folder=temp_dir, data=data)
     papis.database.get().maybe_compute_id(tmp_document)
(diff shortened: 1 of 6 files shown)


alexfikl:
I'm not convinced by the suffix logic in here. 

One thing that's not great: if the user adds more PDF files without `add-file-name`, then they'll be renamed to `<file1>.pdf`, `<file2>-a.pdf`, `<file3>-b.pdf`, which is likely not the intention.

jghauser:
Great improvement! :heart: 

> I'm not convinced by the suffix logic in here.
> 
> One thing that's not great: if the user adds more PDF files without `add-file-name`, then they'll be renamed to `<file1>.pdf`, `<file2>-a.pdf`, `<file3>-b.pdf`, which is likely not the intention.

Can you explain a bit what the problem is here? It sounds like that would be the expected behaviour? Or do you think the default should be warning of clashes and abort?

alexfikl:
> Can you explain a bit what the problem is here? It sounds like that would be the expected behaviour? Or do you think the default should be warning of clashes and abort?

I was mostly thinking that if you give `papis add` some files `["paper1.pdf", "supplement1.pdf"]`, it wouldn't be very helpful to rename them to `["paper1.pdf", "supplement1-a.pdf"]`. They're not overlapping in any way, so the suffix doesn't give any benefit.

jghauser:
Oh right, yeah that really isn't intuitive!

alexfikl:
Finally got back to this a bit. With the new version it should
* Ensure that all files are unique after renaming.
* Not add suffixes unless needed: it keeps a set of known files and adds a suffix only when encountering a repeated one (suffixes are counted by extension as well).

For a few examples of the results (ignoring any bugs, fingers crossed)

1. Without `add-file-name` set to anything (i.e. keeping the original file names)
```sh
$ papis add some-file.pdf some-other-file.pdf some-other-file.md
["some-file.pdf", "some-other-file.pdf", "some-other-file.md"]

$ papis add some-file.pdf some-file.pdf some-other-file.md
["some-file.pdf", "some-file-a.pdf", "some-other-file.md"]
```
we don't do any actual deduplication of whatever the user passed on the command line, so both those files will go in :\\

2. With some `add-file-name`, e.g. `{doc[year]}-{doc[author]}`
```sh
$ papis add some-file.pdf some-other-file.pdf some-other-file.md
["1918-einstein.pdf", "1918-einstein-a.pdf", "1918-einstein.md"]
```
and
```
$ papis addto --files some-new-file.pdf --files some-new-file.md
["1918-einstein.pdf", "1918-einstein-a.pdf", "1918-einstein.md", "1918-einstein-b.pdf", "1918-einstein-a.md"]
```

This looks pretty good to me, so unless someone notices a bug, it should go in soon-ish :grin:

tuurep:
Thanks very much! I don't know if I should be arsed to test it since your comment describes the cases so well, looks good and am excited for this :smile:

alexfikl:
> Thanks very much! I don't know if I should be arsed to test it since your comment describes the cases so well, looks good and am excited for this 😄

Thank you for taking a look! I'll go ahead and merge it then. Feel free to make issues if you find some bugs in there :grin:

PR #835, opened 2024-04-30
format: Add default value to convert_field

hseg:
0eff41a5c6deffdea5a7023511e432b25a6fba34 made `conversion` an
`Optional[str]`, but the builtin `Formatter.convert_field` expects an
`str`. To restore correctness, we need to supply a default argument.

Set the default to `'s'` (string conversion), but perhaps something else is more
appropriate here? Untested, just made `mypy` stop complaining.

### DIFF
--- a/papis/format.py
+++ b/papis/format.py
@@ -100,7 +100,7 @@ class _PythonStringFormatter(string.Formatter):
             from papis.paths import normalize_path
             return normalize_path(str(value))
 
-        return super().convert_field(value, conversion)
+        return super().convert_field(value, conversion or "s")
 
 
 class PythonFormatter(Formatter):


hseg:
Ah, oops. Must've made a mistake in my investigations into `convert_field`'s
behaviour. Probably the best solution here is to amend the package to require
`mypy>=1.10` instead (we _could_ add an explicit case for `None` conversion, but
that'd be a wart and a performance penalty).

alexfikl:
We could also add a `# type: ignore[error-code?]` in there to support older mypy? That makes it clear it's just a typing error and the code is ok. 

> Probably the best solution here is to amend the package to require `mypy>=1.10` instead

Not a big fan of that because we'll just have to update the bound with pretty much every other new mypy release. Same for the various `flake8` plugins, since they can introduce changes sometimes too (but more rarely). Do you know how other projects are handling this?

hseg:
😅 actually, misread the AUR interface -- *I* am (one of the) responsible
packagers.

alexfikl:
> I'd rather not ignore the type error, unless there's a way of gating it to older `mypy` releases -- after all, that line might one day contain a different type error.

Hm, I don't think there's any way of only ignoring the errors for some given versions.

> Not sure what you mean by "we'll just have to update the bound with every other release" -- surely this kind of breakage (mistaken typing information causing spurious type errors) is relatively rare? And in either case, aren't the bounds supposed to convey version ranges of dependencies that should work with the current version of the code?

Incompatible changes are getting rarer, I guess, so maybe it's not such a big pain to update it 2-3 times a year. 

> Then again, I'm more acculturated to Haskell norms, and in particular a SemVer approach to versioning, so I expect a "proper" build system to have good dependency metadata and to pick a working build plan using it.

I'm not sure anyone has ever accused `pip` of being proper! Joking aside, we have a lot of dependencies, so it's pretty easy to get in a conflict when setting stricter bounds. Although these are testing dependencies, so maybe not an issue at all.. 

> My intent initially was just to add this bound to the AUR package (at least, my fork of it, and prod the responsible packager), but in view of the above, why _shouldn't_ we bump the version in `pyproject.toml` (potentially with a note in the commit message clarifying the situation)? After all, this is a change that'd face only the _developers_ (and packager) of the project, not the end-users.

To my knowledge, Arch packages generally just run unit tests (i.e. `pytest` and friends), not linting checks, in the `check` function. It generally seems like asking for trouble to run `mypy` there.

Anyways, to make a decision, I'm fine with adding a bound for `mypy` to `pyproject.toml`, if that's not an issue for you as a packager (?). It essentially means that anyone installing `papis-git` from the AUR will have to do so with `--nocheck` until `mypy` gets updated in the repos.

hseg:
Fair enough, I _am_ being slightly ridiculous with enabling absolutely all tests the project. I could also just de-escalate, and pin a comment at the top of the AUR page notifying users that this particular test is broken, and to either `--nocheck`, comment it out, or upgrade `mypy`.

alexfikl:
@hseg It seems like mypy 1.10.0 is in the Arch repos now. Is this still an issue or can we close it?

hseg:
Sorry, yes - checked and mypy didn't complain.

alexfikl:
Thanks checking and the discussion here! I'll go ahead and close it then :grin:

Issue #841, opened 2024-05-02
[feature] improved document hashing and duplicate checking


pmiam:
Hello! This issue is a follow-up to my post here:
_Originally posted by @pmiam in https://github.com/papis/papis/discussions/549#discussioncomment-9260541_

My PR branch is on my fork. I've not passed all the tests yet.
https://github.com/pmiam/papis/tree/PR/redefine-papis-id-to-aid-duplicate-sniffing

The scope of my changes has grown slightly, thanks to your feedback @alexfikl and @choreutes. I saw an opportunity to enable the option to use a deterministic `papis_id` as a useful duplicate checker for the rare case in which reference files are redundant despite differing metadata.

Summary of current PR branches, Instead of a single branch I've split them up for the time being.
1. re-organized the add command's `run` function
   - add files to tmp_document asap
   - add tmp_document to database early
   - delete `get_document_hash_folder`
2. instead of hashing a bytesting of the doc dict, I reused some of the code from the old `get_document_hash_folder` to hash file contents so this should be no different from old behavior.
   - using hashlib's md5 constructor with `useforsecurity=False` for the benefit of FIPS compliant users
   - this requires bumping the minimum python version to 3.9

My current issue is that I need to redress the duplicate checking in order to pass tests. Currently duplicate checking is a try-except-else block in the add command's `run` function. 

In order to use `papis_id` in duplicate checking, the database entry must be made before duplicate checking. Currently, that same entry is trivially found by the duplicate check. There's no simple way around that with the current code.

I think duplicate checking should be (or possibly already is) a database function, but I welcome some more feedback.

alexfikl:
I haven't looked at your duplicate checking changes, but I can't say I understand why you'd want to use `papis_id` as a duplicate check. 

I suggest you start with some PR with the clearer changes, e.g. using `papis_id` instead of `get_document_hash_folder`, and then we can see if it's worth reworking anything else.

pmiam:
No duplicate checking changes yet. I'd like to discuss what/if they should be before I move forward.

Thanks for the advice. I refocused on my original goal and opened my PR. I had no idea you could make a draft PR. Seems obvious in hindsight.

Using a checksum to compare files is common in my work. Comparisons based on metadata are definitely more useful when collecting e.g. literature or videos but I also do a lot of archiving. Note, I'm brand new to Papis but I think it could help me there especially if it kept track of meaningful checksums. Of course, it will be completely opt-in.

alexfikl:
[...]

PR #842, opened 2024-05-03
organize: simplified default folder naming

pmiam:
1. deprecate `get_document_hash_folder`
2. consolidates document hashing under `compute_an_id`
3. simplifies the creation of default directory names
4. makes the papis_id value available at format time for most templates 
   - the ref template is excluded so as to not undo #484

One test needed to be updated (https://github.com/papis/papis/pull/842/commits/02fa5a6e079f72280ddcb4df16376a7a2b3d8228) to accommodate these changes.

A side effect of these changes are that the path creating functions which used to be independent of the database now need the database to know about the document before they're called. I think that sounds worse than it is, but let me know your thoughts. With the updated test, everything passes. Also, I manually tested the `use-cache = False` case and it seems to work.

Also, I also exposed a (possibly pre-existing) problem where if doc.save() is called before doc.set_folder(), the save fails due to an unset info file path.

These changes might lead to improved duplicate checking being discussed at issue #841


jghauser:
Thanks a lot for this PR, @pmiam!

A quick question: does this by default lead to folder names that are just md5 hashes? I'm not sure I'm a big fan of this as it makes the folder names not very human-readable (in fact, I think even the current default isn't great as it starts with the hash, and i think folders being organised by e.g. author family name would be more intuitive). Of course, this can be changed in the settings and doesn't present a problem as long as one uses the `papis` command, but I think we should aim for default values that are as approachable as possible.

papis/commands/add.py
```diff
@@ -247,25 +247,28 @@ def run(paths: List[str],
...
+
     import tempfile
 
     in_documents_paths = paths
     temp_dir = tempfile.mkdtemp()
-
-    for p in in_documents_paths:
-        if not os.path.exists(p):
-            raise FileNotFoundError(f"File '{p}' not found")
+    tmp_document = papis.document.Document(folder=temp_dir, data=data)
```
alexfikl, line 259: I would add the `papis_id` computation right here, i.e. 
```suggestion
    tmp_document = papis.document.Document(folder=temp_dir, data=data)
	papis.database.get().maybe_compute_id(tmp_document)
```
Then it can be used by the ref and probably don't need to move any of the other things around either? Is there any issue with doing that? Maybe complain if `data` is empty?

papis/commands/add.py
```diff
@@ -364,6 +357,16 @@ def run(paths: List[str],
         confirm = True
         has_duplicate = True
 
+    papis.database.get().add(tmp_document)
```
alexfikl, line 360: Not a big fan of adding a temporary document to the library. Does this even get cleaned up?

papis/document.py
```diff
@@ -308,6 +308,9 @@ class Document(Dict[str, Any]):
 
         A :class:`DocHtmlEscaped` instance that can be used to escape keys
         in the document for use in HTML documents.
+
+    WARNING: running doc.save() without first running doc.set_folder()
+    will fail due to an unset info file path.
```
alexfikl, line 313: This would be better added in the docs for `Document.save`. Ideally we'd also throw a nicer error in there (not sure what error it throws now?).

tests/test_paths.py
```diff
@@ -140,82 +140,51 @@ def test_get_document_file_name_format(tmp_library: TemporaryLibrary) -> None:
...
-    newnewhh = get_document_hash_folder(data, [filename])
-    assert newnewhh != newhh
-
-
 def test_get_document_folder(tmp_library: TemporaryLibrary) -> None:
     from papis.document import from_data
+    import papis.database
 
-    pdf = tmp_library.create_random_file("pdf")
+    db = papis.database.get(library_name=tmp_library.libname)
```
alexfikl, line 147:
```suggestion
    db = papis.database.get()
```
should be sufficient here? The `tmp_library.libname` is already the default library.

pmiam:
Thank you for the warm welcome!

> does this by default lead to folder names that are just md5 hashes?

It does, although it also makes it so the `papid_id` is available in templates for people like me who prefer it that way.

My personal preference is for machine-managed, flat directory trees. Frankly, that makes the most sense to me as a default, and I think it encourages use of the Papis tui or another front end, but I see your point. Especially as I was writing this, I noticed that a lot of work has been done to make human readable file names very expressive and robust to collision. I've made sure to leave that unaffected.

We could set a human friendly template as the default configuration easily.

alexfikl (commented):
I think we can further simplify the changes to `papis.commands.add`, but overall this looks like a good direction!

I agree with @jghauser that we should try for a better default folder name `HASHHASHHASH-author-author` was not the worst, but just using the MD5 hash looks very opaque. I think Zotero also stores files on disk in some `storage/PIMHYJGK` type folder and it's very hard to find things without the interface.

papis/commands/add.py
pmiam: I went with db.add(doc) because it seemed like a safer, higher-level entry point. Indeed, it is probably better to use maybe_compute_id directly, as you do above.

jghauser:
> @pmiam @jghauser Maybe the default could be something based on `time-added` and not `papis_id`? (both keys should always be available) E.g. `{doc[time-added]}-{doc[author]}`?

This is a very intriguing idea! It would avoid collisions and easily allows sorting by a very useful property (a bit of a shame that it would lead to oldest at the top by default, though i guess that cannot be avoided).

What I personally use is `SURNAME_OF_FIRST_AUTHOR-YEAR-FIRST_5_WORDS_OF_TITLE` (to be more precise, it's the `jinja2` template I mentioned [here](https://github.com/papis/papis/discussions/836#discussioncomment-9282935)), which i think looks quite nice and doesn't really lead to collisions (though I guess it could be complemented with a suffix as for file names). Not sure if that could be (easily) implemented without `jinja2` though. I do think that having at least some of the title in the directory name is useful.

alexfikl:
> (a bit of a shame that it would lead to oldest at the top by default, though i guess that cannot be avoided).

No worries, just reverse it with `ls -r` :grin:

> What I personally use is `SURNAME_OF_FIRST_AUTHOR-YEAR-FIRST_5_WORDS_OF_TITLE` (to be more precise, it's the `jinja2` template I mentioned [here](https://github.com/papis/papis/discussions/836#discussioncomment-9282935)), which i think looks quite nice and doesn't really lead to collisions (though I guess it could be complemented with a suffix as for file names). Not sure if that could be (easily) implemented without `jinja2` though. I do think that having at least some of the title in the directory name is useful.

Hm, I think we have two issues here
* First, it would be nice to have a good default for `add-folder-name`. I'm not sure I feel strongly here, so anything like `{doc[author]}-{doc[title:.5S]}` or `{doc[year]}-{doc[author]}` should work fine?
* Second, we need a fallback value when constructing a folder name for when a mostly empty document comes in. This should then just use `{doc[time-added]}` and whatever else it can find in the document to identify it?

jghauser:
> We can probably make that another PR too. In this PR, it's sufficient to move `papis_id` higher-up so that it can be used in formatting strings and deprecate `get_document_hash_folder`. 😁

Totally agree, I'm gonna respond in the issue that you opened!

papis/document.py
pmiam: It was `FileNotFoundError: [Errno 2] No such file or directory: ''`

I added a logger.error message and raised DocumentFolderNotFound.

papis/commands/add.py
pmiam: In fact, I just relocated the db.add call without realizing it would add the temporary path to the cache, I just reverted that commit.

papis/commands/add.py
pmiam: done

doc/source/default-settings.rst
```diff
@@ -413,9 +413,9 @@ Add options
 
         add-folder-name = ``{doc[year]} / {doc[author]}-{doc[title]}``
 
-    If this setting is empty, a (MD5) hash of some document contents followed
-    by the author names is used. This will generally result in unique folder
-    names. If a folder name is not unique and the document does not appear to
+    If this setting is empty, a (MD5) hash of some document contents is used.
```
alexfikl, line 416:
```suggestion
    If this setting is empty, a hash of some document contents is used.
```
? We probably shouldn't be specific about the hash here.

papis/document.py
```diff
@@ -425,6 +425,14 @@ def get_notes(self) -> List[str]:
 
     def save(self) -> None:
         """Saves the current document fields into the info file."""
+
+        folder = self.get_main_folder()
+        if not folder or not os.path.exists(folder):
+            from papis.exceptions import DocumentFolderNotFound
+            logger.error("running doc.save() without first running doc.set_folder() "
+                         "will fail due to an unset info file path.")
```
alexfikl, line 433:
```suggestion
```
? The exception should have sufficient context about what happened and the caller can write a message if they want to.

papis/paths.py
```diff
@@ -199,80 +201,27 @@ def get_document_hash_folder(
...
-
-    if file_read_limit > 0:
-        for path in paths:
-            if not os.path.exists(path):
-                continue
-
-            with open(path, "rb") as fd:
-                md5.update(fd.read(file_read_limit))
+    warn("'get_document_hash_folder' is deprecated and will be removed"
+         "Use 'papis.id.compute_an_id' instead.",
```
alexfikl, line 205:
```suggestion
    warn("'get_document_hash_folder' is deprecated and will be removed. "
         "Use 'papis.paths.get_document_folder' instead.",
```

papis/paths.py
```diff
@@ -199,80 +201,27 @@ def get_document_hash_folder(
...
+        dirname: PathLike, *,
         folder_name_format: Optional[str] = None) -> str:
     """Generate a folder name for the document at *dirname*.
 
     This function uses :confval:`add-folder-name` to generate a folder name for
     the *doc* at *dirname*. If no folder can be constructed from the format, then
-    :func:`get_document_hash_folder` is used instead. The returned folder name
-    is guaranteed to be unique and to be a subfolder of *dirname*.
+    the reference's papis_id is used instead as a subfolder of *dirname*. The
+    papis_id is guaranteed to be unique.
```
alexfikl, line 221:
```suggestion
    the document's ``papis_id` is used instead as a subfolder of *dirname*. The
    ``papis_id`` is guaranteed to be unique.
```

papis/paths.py
```diff
@@ -312,16 +261,15 @@ def get_document_folder(
...
     if out_folder_path == dirname:
         if folder_name_format:
             logger.error(
                 "Could not produce a folder path from the provided data:\n"
                 "\tdata: %s", doc)
 
-        logger.info("Constructing an automatic (hashed) folder name.")
-        out_folder_name = get_document_hash_folder(doc, in_document_paths)
-        out_folder_path = os.path.join(dirname, out_folder_name)
+        logger.info("using database (hashed) index as reference folder name.")
```
alexfikl, line 271:
```suggestion
        logger.info("Falling back to 'papis_id' as a reference folder name.")
```
?

alexfikl (approved):
Left a few small nitpicks around the logger messages mostly, but this should be good to go! :heart:

doc/source/default-settings.rst
pmiam: I reckon the whole paragraph was a little unclear. I rewrote it with the future human-readable default in mind.

papis/paths.py
pmiam: I looked up the asymmetric backticks. I assume they're a typo?

papis/document.py
pmiam: good point. I added context to the exception. I checked all the other places it is raised and I think that's a fair step. It took me some time to figure it out when I first encountered it.

papis/paths.py
alexfikl: Yep, that was a typo!

pmiam:
I'll get right on that, sorry. I've been working between celebrations today. Happy Easter!

alexfikl:
> I'll get right on that, sorry. I've been working between celebrations today.

No worries, it's fixed and I'll merge as soon as the tests pass!

> Happy Easter!

Happy Easter!

PR #846, opened 2024-05-08
defaults: do not force isbn type

alexfikl:
@jghauser This removes the expected type for `isbn` since there seem to be quite a few documents in my library where it's a list. Those documents have e.g. an ISBN for the hardcover and an ISBN for the e-book version and they we both added from Google Books (or something, not sure?).

I'm not sure if this is better fixed in the library, i.e. only one ISBN should be kept. What do you think?

### DIFF
--- a/papis/defaults.py
+++ b/papis/defaults.py
@@ -110,7 +110,6 @@ settings: Dict[str, Any] = {
                                    "author_list:list",
                                    "doi:str",
                                    "ref:str",
-                                   "isbn:str",
                                    "author:str",
                                    "journal:str",
                                    "note:str",


alexfikl:
For an example of such a document (apparently most of them are chapters in books from Crossref):
https://api.crossref.org/works/10.1007/978-3-0348-8720-5_13

jghauser:
I feel like technically, there should only be one `isbn` (namely the one of the book the chapter is actually taken from) as it's possible (?) that e.g. pagination differs between the two. Also, if a citation style uses the `isbn`, then multiple `isbn`s are likely to break things (though this is purely conjecture). [This stackexchange discussion](https://tex.stackexchange.com/questions/231212/multiple-isbns-for-a-given-entry-e-g-a-book-in-biblatex) seems to indicate at least that the situation is non-standard and on bibtex.com, [all examples are simply a string](https://www.bibtex.com/f/isbn-field/).

I'm not very knowledgeable in all this though :smile:.

alexfikl:
> I feel like technically, there should only be one `isbn` (namely the one of the book the chapter is actually taken from) as it's possible (?) that e.g. pagination differs between the two. Also, if a citation style uses the `isbn`, then multiple `isbn`s are likely to break things (though this is purely conjecture). [This stackexchange discussion](https://tex.stackexchange.com/questions/231212/multiple-isbns-for-a-given-entry-e-g-a-book-in-biblatex) seems to indicate at least that the situation is non-standard and on bibtex.com, [all examples are simply a string](https://www.bibtex.com/f/isbn-field/).
> 
> I'm not very knowledgeable in all this though 😄.

Hm, that's a good point, I didn't check how BibTeX handles things like this. Maybe a better fix for this is to update the crossref importer to only produce one `isbn` field and maybe additional `isbnprint` and `isbnonline` like in that stackoverflow post (if there actually are multiple ISBN fields).

Thanks for looking over this!

PR #847, opened 2024-05-09
crossref: handle multiple ISBNs

alexfikl:
According to the Crossref [API Format](https://github.com/CrossRef/rest-api-doc/blob/583a8dbad0a063da4aa5ec319df33130a26ef650/api_format.md), the ISBN can be a list or a string. We mostly just want a single ISBN, so this changes the logic to
* If `ISBN` is a list, just take the first element. Sometimes this is even a list with only one element.
* If `isbn-type` exists, look for an `electronic` ISBN and use that. From looking at their tests, `electronic` and `print` seem to be the only two options for the ISBN `type`.

cc @jghauser

### DIFF
--- a/papis/crossref.py
+++ b/papis/crossref.py
@@ -93,6 +93,8 @@ type_converter = {
     "standard": "techreport",
 }
 
+# NOTE: fields checked against the official API format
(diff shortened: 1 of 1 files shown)


jghauser:
This looks like the correct way to handle this! :tada:

alexfikl:
> This looks like the correct way to handle this! 🎉

Thanks for taking a look! I'll go ahead and merge this and fix my library :heart:

PR #854, opened 2024-05-18
commands.doctor: rename keys-exist to keys-missing

### LINKED ISSUE #849: Papis doctor: `keys-exist`
I just started using `papis doctor` and noticed I got a lot of `keys-exist` messages. When I checked, it was because certain keys were missing. Maybe we might want to rename that to `key-missing` or something of that sort?

alexfikl:
This renames the `keys-exist` check to `keys-missing` with a bunch of deprecation warnings.

@jghauser What do you think about this? Should we also change the default to `biblatex-required-keys` for smarter missing key reporting?

Fixes #849.

### DIFF
--- a/doc/source/default-settings.rst
+++ b/doc/source/default-settings.rst
@@ -557,9 +557,9 @@ Doctor options
 
     A list of checks that are performed by default.
 
-.. papis-config:: doctor-keys-exist-keys
+.. papis-config:: doctor-keys-missing-keys
 
-    A list of keys used by the ``keys-exist`` check. The check will show an
+    A list of keys used by the ``keys-missing`` check. The check will show an
     error if these keys are not present in a document.
 
 .. papis-config:: doctor-duplicated-keys-keys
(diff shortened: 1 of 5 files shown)


jghauser:
Yes, `keys-missing` is much better. And yeah, I think `biblatex-required-keys` might be a better default, the `keys-exist` test generates a lot of false positives.

alexfikl:
> Yes, `keys-missing` is much better. And yeah, I think `biblatex-required-keys` might be a better default, the `keys-exist` test generates a lot of false positives.

@jghauser I went ahead and changed the defaults from `files, keys-missing, duplicated-keys` to `files, biblatex-required-keys, bibtex-type, refs`. What do you think? The motivation was mostly to make sure the metadata is BibTeX friendly since we're pretty BibTeX oriented. These should also not have many (any?) false positives.

alexfikl:
Thanks for looking over this! In it goes :rocket:

jghauser:
Just ran this and I'm getting `bibtex-type` errors twice:
```
bibtex-type	type	/home/julian/Documents/library/main/snowden-2019-permanent-record
bibtex-type	type	/home/julian/Documents/library/main/snowden-2019-permanent-record
```
This document doesn't have a type at all. Not sure if this is because of this PR or something that happened before as well.

jghauser:
Oh my bad, I thought that was already merged!

PR #855, opened 2024-05-19
config: allow libraries in default-settings

alexfikl:
This should allow setting libraries in `config.py` using `register_default_settings` dynamically. Something like this should now work as expected
```
import pathlib

from papis.config import register_default_settings

basedir = pathlib.Path('/home/user/Documents')
register_default_settings({
    **{subdir.stem: {'dir': str(subdir)} for subdir in basedir.glob('[a-z]*/')},
	**{"settings": {"default-library": "lib1"}}
    })
```

xref: #851

### DIFF
--- a/papis/config.py
+++ b/papis/config.py
@@ -562,8 +562,10 @@ def get_lib_from_name(libname: str) -> papis.library.Library:
         to an existing folder that should be considered a library.
     """
     config = get_configuration()
+    default_settings = get_default_settings()
+    libs = get_libs_from_config(config)
 
-    if libname not in config:
+    if libname not in libs:
         if os.path.isdir(libname):
             logger.warning("Setting path '%s' as the main library folder.", libname)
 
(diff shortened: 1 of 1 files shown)


alexfikl:
@octvs In case you have the git version of papis available, it would be great if you get a chance to test this!

octvs:
Just gave it a try with my [original config files], however when I execute 

```sh
$ papis -l INFO --pick-lib open
```
to see which libraries have been registered, I see the library defined as
`default-library` on `[settings]` section duplicated. Which justifies the
following block from your snippet:

```py
if libname not in self:
    self.add_section(libname)
```

I will check the API doc to see whether I can refer back to already registered
libs on `ConfigParser`, which would let me include a similar check in a less
hacky way.


[original config files]: https://github.com/papis/papis/discussions/851#discussioncomment-9479742

alexfikl:
> to see which libraries have been registered, I see the library defined as `default-library` on `[settings]` section duplicated.

Oops, nice catch! That should be fixed in the latest version. 

I went with the following choice: if the library is defined in the static `config` file, it takes priority over a library with the same name defined in `config.py` using `papis.config.register_default_settings`.

octvs:
Just gave it a try, it works as expected now.

> if the library is defined in the static `config` file, it takes priority over a library with the same name defined in `config.py`

I wonder if this should be documented somewhere.

alexfikl:
> I wonder if this should be documented somewhere.

Yeah, it really should! I've opened an issue to improve the documentation for `config.py`, since at the moment it's pretty much non-existent.

Thanks for all the testing and help with this! In it goes! :rocket:

PR #856, opened 2024-05-19
pick: add nicer library picker

alexfikl:
This adds a little wrapper to pick libraries (e.g. `papis --pick-lib add`) with more color. On my dark terminal it looks something like this

Screenshot_20240519_175742

@jghauser I'm not sure if there is any other nice information we could display about the libraries. Any thoughts?


jghauser:
This looks good, I don't think any other information is needed here!

alexfikl:
> This looks good, I don't think any other information is needed here!

Thanks for taking a look! :heart:

I've added a configuration option for it called `library-header-name` to match other similar things (docs, marks, etc). Should be good to one once the CI passes.

PR #864, opened 2024-05-26
commands.serve: port away from cgi.FieldStorage

alexfikl:
This ports away from the deprecated `cgi` module in the standard library that will be removed in Python 3.13 (coming out this fall). The `read_urlencoded` function is mostly copied from the standard library and the rest of the code is a thin wrapper around that.
https://docs.python.org/3.13/whatsnew/3.13.html#pep-594-dead-batteries-and-other-module-removals

Note that this is just a stopgap until `papis serve` gets ported to `flask` or something more capable.

@alejandrogallo I've tested this a bit and it works when adding new keys to the document, updating the `info.yaml` file, adding to `notes.md`. I'm not sure that any more advanced usage works, since I don't use `papis serve`, so any additional testing is very appreciated!

### DIFF
--- a/papis/commands/serve.py
+++ b/papis/commands/serve.py
@@ -6,7 +6,6 @@ import urllib.parse
 from typing import Any, List, Optional, Tuple, Callable, Dict
 
 import functools
-import cgi
 import collections
 import tempfile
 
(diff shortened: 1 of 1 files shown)


alexfikl:
For the moment, this just conditionally uses our own `FieldStorage`, mostly to allow anyone using python 3.13 to try it out. Should be pretty safe to merge in, since it doesn't affect older versions.

Issue #874, opened 2024-06-14
`papis doctor` and linebreaks in titles (+ other improvements)


jghauser:
While investigating [a bug in papis.nvim](https://github.com/jghauser/papis.nvim/issues/36), I realised that the root cause might be a stray `\n` in one of the titles defined in an `info.yaml`. Are linebreaks allowed in titles? If not, it might be useful to add a check to `papis doctor` for this. The entry in question was added automatically, which means the linebreak must have snuck in via some importer/downloader.

Short list of possible random "string cleaning" improvements (also in comments):
* [ ] Remove newlines `\n` and `\n\r` from titles and abstracts, e.g. "My\nawesome paper".
* [ ] Remove repeated whitespace, e.g. "My awesome     paper".
* [ ] Remove any starting "abstract" suffix in the abstract text.
* [ ] Add some spaces in author initials, e.g. many sources give "J.R.R. Tolkien" which could be "J. R. R. Tolkien".

alexfikl:
I'm not sure in what scenario they would be allowed, but I imagine generally it's not a good idea (it would at least mess up BibTeX).

I agree that we should have a very basic "string cleaner" to remove some stray things like that. Other things that it could do: add dots `J C -> J. C.` (mainly for authors), add spaces `J.C. -> J. C.`, remove repeated whitespace and tabs and things.

jghauser:
Yeah that would make sense. Actually, I think any `\n` that aren't at the end of a line should be eliminated (found some more while investigating, they just happened to not be in fields that papis.nvim uses).

avonmoll:
Slightly related question: for as long as I can remember, really long lines seem to get "hard wrapped" by papis, automatically. I can't seem to find anything about this in the docs. This almost always happens with long fields like `abstractNote`. Occasionally it happens on very long titles (which I suspect is why I was having the issues with the telescope picker in papis.nvim in the first place).

My question: is this the correct behavior for papis? It seems strange.

alexfikl:
> Slightly related question: for as long as I can remember, really long lines seem to get "hard wrapped" by papis, automatically. I can't seem to find anything about this in the docs. This almost always happens with long fields like `abstractNote`.

@avonmoll Hm, haven't looked at this, but my best guess is that it has something to do with `yaml.dump`. At least for `pyyaml` there is a `width` argument that seems to be able to control the wrapping: https://pyyaml.org/wiki/PyYAMLDocumentation.

[...]

PR #881, opened 2024-06-21
commands.doctor: smarter cleanup of JATS / MML tags

### LINKED ISSUE #880: `papis doctor -t html-tags` leaves bad strings
Currently, this command simply deletes all html tags. Unfortunately, this means that if you have e.g. `<jats:title>Abstract</jats:title>This paper...` (which in my case is in almost every abstract), we end up with `AbstractThis paper...`. This isn't ideal. I'm not sure what the best approach here would be. Maybe collect more regexes for strings where the text inside the tags can be deleted? Or can we safely say that such text can always be deleted?

alexfikl:
The `html-tags` doctor check is quite limited in how it can remove HTML from a key. This adds some smarter tag removal for some common markup (at least as returned by Crossref). The main downside is that it's using `bs4` to actually parse the HTML / XML, so it's slower than before.

Fixes #880.


alexfikl:
@jghauser I had some documents with that `<jats:title>Abstract<jats:title>` thing too. Ideally we would just remove it, but I'm not sure how to do that in a robust way that doesn't also remove other useful text..

alexfikl:
Found https://github.com/DSpace/DSpace/pull/9386 that seems to deal with a similar issue. They special cases the `jats` tags and did some fancier parsing.

jghauser:
I think this is an ok stopgap measure, though ideally I'd like to find a more robust way of handling it (though I'm not quite sure how to do so in a non overkill fashion haha). When the text is "Abstract", the result isn't pretty but also not overly confusing. But it seems to be arbitrary text at times, one entry for instance has `<jats:title>The Space in Your Head</jats:title>`, which gets very confusing if it's just connected with a space to the next sentence. One thing we might want to do is delete the string within the title tags if it's "abstract" and maybe otherwise add a ": " after the string. But I'm not sure what edge cases that heuristic misses...

Btw. it is possible to re-download all the abstracts with Papis? At the moment I'm a bit hesitant to have `papis doctor` fix things because I don't know how easy it would be to fix things if it messes it up.

alexfikl:
> I think this is an ok stopgap measure, though ideally I'd like to find a more robust way of handling it (though I'm not quite sure how to do so in a non overkill fashion haha). When the text is "Abstract", the result isn't pretty but also not overly confusing. But it seems to be arbitrary text at times, one entry for instance has `<jats:title>The Space in Your Head</jats:title>`, which gets very confusing if it's just connected with a space to the next sentence. One thing we might want to do is delete the string within the title tags if it's "abstract" and maybe otherwise add a ": " after the string. But I'm not sure what edge cases that heuristic misses...

Yeah, I don't really know what the best way to go here would be. Maybe using `markdownify` like the zenodo importer does could work? It would sort of leave paragraphs and headings alone. Might need to do a dumb JATS to HTML conversion first though.

> Btw. it is possible to re-download all the abstracts with Papis? At the moment I'm a bit hesitant to have papis doctor fix things because I don't know how easy it would be to fix things if it messes it up.

Hm, I don't think so. I mostly have everything in git so I kept doing `git restore . && papis doctor --fix --all` until it looked ok :\ 

You can always use `papis update --from ...` or whatever to redownload the whole thing, but that's not very nice either. It would be cool to teach `papis update` to do something like that :grin:

jghauser:
> Yeah, I don't really know what the best way to go here would be. Maybe using `markdownify` like the zenodo importer does could work? It would sort of leave paragraphs and headings alone. Might need to do a dumb JATS to HTML conversion first though.

Yeah that would be nice. It would also allow preserving the various italics etc. Though converting from JATS to HTML is obv another can of worms. Now why could they not just have gone with HTML...

> Hm, I don't think so. I mostly have everything in git so I kept doing `git restore . && papis doctor --fix --all` until it looked ok :\
> 
> You can always use `papis update --from ...` or whatever to redownload the whole thing, but that's not very nice either. It would be cool to teach `papis update` to do something like that 😁

Yeah I dont really want to restore an old state as I might have edited other things in the meantime (like tags etc). That would indeed be a good feature to add to `papis update`. Adding it to my list :)

alexfikl:
> Yeah that would be nice. It would also allow preserving the various italics etc. Though converting from JATS to HTML is obv another can of worms. Now why could they not just have gone with HTML...

I'll try to look into that a bit over the weekend. At least to get it to work reasonably on the example from https://github.com/DSpace/DSpace/pull/9386.

alexfikl:
@jghauser I tried parsing the JATS tags a bit, but that's indeed quite a pain to do even remotely right :( Instead, I went with a dumber version that still somewhat maintains the formatting. What do you think? (there's an example abstract conversion in the test files)

Do you have any additional documents that we can add to the tests? I have a few abstract that have some weird MathML tags in them that got messed up by this, so will need to work on it some more..

jghauser:
I just realised that pandoc says it can convert JATS. maybe this is a situation in which doing this ourselves isn't the right approach and we should instead support using other software? :sweat_smile: 

It might be nice to have a doctor fix thing that simply pipes the relevant strings to pandoc and then takes its output.

Though while it works with some simpler examples I tried, it seems to choke on your example too...

alexfikl:
@jghauser This is working pretty well for me now. I had around 60 abstracts with all sorts of weird markup and it cleaned them up pretty well! Some are added to the tests too.

Can you try and see if it works for you? One hacky thing that you can do so that you don't actually modify your library is go into the `make_fixer` in `html_tags_check` and replace the write, i.e.
```python
print(text)
# doc[key] = text 	## << comment out this line
```
so that it just prints out the cleaned up text.

jghauser:
These abstracts you're processing look very nice! :tada:

Here's an example that is pretty bad and that might be useful to your testing:
```
<jats:title>The Space in Your Head</jats:title>          <jats:p>            Space,
  and events associated with places and spaces, are represented in the brain by a
  circuitry made of place cells, head direction cells, grid cells, and border cells.
  These cell types form a collective dynamic representation of our position as we
  move through the environment. How this representation is formed has remained a mystery.
  Is it acquired, or are we born with the ability to represent external space (see
  the Perspective by            <jats:bold>              <jats:related-article xmlns:xlink="http://www.w3.org/1999/xlink"
  ext-link-type="doi" issue="5985" page="1487" related-article-type="in-this-issue"
  vol="328" xlink:href="10.1126/science.1191527">                <jats:bold>Palmer
  and Lynch</jats:bold>              </jats:related-article>            </jats:bold>            )?            <jats:bold>              Langston              <jats:italic>et
  al.</jats:italic>            </jats:bold>            (p.            <jats:related-article
  xmlns:xlink="http://www.w3.org/1999/xlink" ext-link-type="doi" page="1576" related-article-type="in-this-issue"
  vol="328" xlink:href="10.1126/science.1188210">1576</jats:related-article>            )
  and            <jats:bold>              Wills              <jats:italic>et al.</jats:italic>            </jats:bold>            (p.            <jats:related-article
  xmlns:xlink="http://www.w3.org/1999/xlink" ext-link-type="doi" page="1573" related-article-type="in-this-issue"
  vol="328" xlink:href="10.1126/science.1188224">1573</jats:related-article>            )
  investigated the early development of spatial activity in the hippocampal formation
  and the entorhinal cortex of rat pups when they first began to explore their environment.
  Rudiments of place cells, head direction cells, and grid cells already existed when
  the pups made their first movements out of the nest. A neural representation of
  external space at this early time points to strong innate components for perception
  of space. These findings provide experimental support for Kant's 200-year-old concept
  of space as an a priori faculty of the mind.          </jats:p>
```

Even pandoc struggles with it (converting to markdown):
```
Space, and events associated with places and spaces, are represented in
the brain by a circuitry made of place cells, head direction cells, grid
cells, and border cells. These cell types form a collective dynamic
representation of our position as we move through the environment. How
this representation is formed has remained a mystery. Is it acquired, or
are we born with the ability to represent external space (see the
Perspective by ****Palmer and Lynch**** )? **Langston *et al.*** (p.

1576

) and **Wills *et al.*** (p.

1573

) investigated the early development of spatial activity in the
hippocampal formation and the entorhinal cortex of rat pups when they
first began to explore their environment. Rudiments of place cells, head
direction cells, and grid cells already existed when the pups made their
first movements out of the nest. A neural representation of external
space at this early time points to strong innate components for
perception of space. These findings provide experimental support for
Kant\'s 200-year-old concept of space as an a priori faculty of the
mind.
```

jghauser:
Only saw ur message after writing mine! I'll test it with my library and will let you know :)

alexfikl:
> Only saw ur message after writing mine! I'll test it with my library and will let you know :)

Awesome! Thank you very much :grin: 

I added your abstract thing to the tests and it seems to do a reasonable job at it. At least it doesn't add weird newlines in the middle.

jghauser:
I just checked my library with this branch, and the results seem to be excellent! I didn't read through all of the results, but quickly looking over everything, I couldn't spot any problems. The only thing that could maybe be improved is to remove the string "abstract" in all types of capitalisation if it's the very first word. There are some instances of "ABSTRACT" in my library, which your solution doesn't yet catch it seems.

alexfikl:
> I just checked my library with this branch, and the results seem to be excellent! I didn't read through all of the results, but quickly looking over everything, I couldn't spot any problems. 

Woohoo! :tada: Thanks for all the testing and help with this.

> The only thing that could maybe be improved is to remove the string "abstract" in all types of capitalisation if it's the very first word. There are some instances of "ABSTRACT" in my library, which your solution doesn't yet catch it seems.

This is not very related to the HTML tags stuff, so I'd rather leave it for another PR, if that's ok. Maybe add it to the list in #874?

PR #886, opened 2024-07-15
Fix importing and exporting of institutional authors

### LINKED ISSUE #885: Handling institutional authors
* papis, version 0.13:

I am having trouble dealing with institutional authors. In order for them to be cited correctly using biblatex they must be exported to bib-files as such: `{{The name of the institutions}}`. Otherwise this gets interpreted as name-surname-pair and results in something like "Name of the institution, T." when referenced.

# how to reproduce the issue
Consider the following bib file `ipcc1990.bib`
```
@report{ipcc1990,
  author={{Intergovernmental Panel on Climate Change}},
  title={Contribution of Working Group I to the First Assessment Report of the Intergovernmental Panel on Climate Change},
  year={1990},
}
```
After importing it via `papis add --from bibtex ipcc1990.bib` it looks something like this:

``` 
author: on Climate Change, Intergovernmental Panel
author_list:
- family: on Climate Change
  given: Intergovernmental Panel
files: []
papis_id: ac0ce27201d0360e806daa3924f77201
ref: ipcc1990
time-added: 2024-07-10-11:23:44
title: Contribution of Working Group I to the First Assessment Report of the Intergovernmental Panel on Climate Change
type: report
year: '1990'

```
The author gets interpreted as name-surname-pair. I could live with that and change it manually to:
```
author: {{Intergovernmental Panel on Climate Change}}
files: []
papis_id: ac0ce27201d0360e806daa3924f77201
ref: ipcc1990
time-added: 2024-07-10-11:23:44
title: Contribution of Working Group I to the First Assessment Report of the Intergovernmental Panel on Climate Change
type: report
year: '1990'
```
Unfortunately, when trying to export that I get an error:

```sh
$ papis export ipcc1990 -f bibtex
[INFO] database.cache: Indexing library. This might take a while...
[ERROR] document: Error reading info file at '~/papis/ipcc1990__ac0ce27201d0360e806daa3924f77201/info.yaml'. Please check it! (Caught exception 'ValueError: while constructing a mapping
  in...'. Use `--log DEBUG` to see traceback)
No document selected!
```

For some reason, this also deletes all metadata from the yaml file except for the id:

```sh
$ cat ~/papis/ipcc1990__ac0ce27201d0360e806daa3924f77201/info.yaml
papis_id: e1f27a34adca7f3764a44fe99b5dcf3a
```
Any idea on how to deal with this? For now, any type of dirty hack would also be much appreciated!

alexfikl:
As shown in #885, some entries only have a single name author. This improves the handling of such entries by:
* Not doing `latex_to_unicode` on the whole author field. This removed the extra "{}" around an entry and confused "splitauthor" to think there are multiple names there.
* Adding a custom `author_list_to_author` for BibTeX so that we can handle author names in a smarter way. Now where there's a missing family or given name, it just uses `{name}` (with the curly braces, so that a roundtrip gives the same result).

Fixes #885.

### DIFF
--- a/papis/bibtex.py
+++ b/papis/bibtex.py
@@ -512,6 +512,29 @@ def create_reference(doc: Dict[str, Any], force: bool = False) -> str:
     return ref_cleanup(ref)
 
 
+def author_list_to_author(author_list: List[Dict[str, Any]]) -> str:
+    if not author_list:
+        return ""
+
+    result = []
+    fmt = "{au[family]}, {au[given]}"
+
+    for author in author_list:
+        family = author.get("family")
+        given = author.get("given")
+        if family and given:
+            result.append(fmt.format(au=author))
+        elif family:
+            result.append(f"{{{family}}}")
+        elif given:
+            result.append(f"{{{given}}}")
+        else:
+            # NOTE: empty author, just skip it
+            pass
+
+    return " and ".join(result)
+
+
 def to_bibtex_multiple(documents: List[papis.document.Document]) -> Iterator[str]:
     for doc in documents:
         bib = to_bibtex(doc)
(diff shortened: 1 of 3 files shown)


alexfikl:
Zenodo failures seem to also be happening on main #887.

PR #892, opened 2024-07-26
Add a on_add_done hook

alexfikl:
This adds a hook that gets called after all the editing in `papis add`. It should allow for some final touch-ups.


alexfikl:
These things could use some testing, but otherwise this is pretty straightforward. I'll go ahead and merge it.

PR #896, opened 2024-08-08
document: Clear dictionary before loading data

### LINKED ISSUE #895: `document.load()` does not delete fields
First of all, absolutely massive fan of `papis`, big thanks to all who have worked on it.

I came across an issue while editing documents in `papis`:  If you edit a `.yaml` file via `papis edit`, and *delete* a key from the file, it is not reflected in the database.

I can provide a minimal working example if helpful.  But I believe I found the core issue - it is this line in `document.load()`, which is called in `papis.command.edit.run()` to update the document after editing.

https://github.com/papis/papis/blob/129773dd8b35a734d19206a1640b394179c224cd/papis/document.py#L454

Because this calls `self.update(data)`, which I understand to essentially be `dict.update(data)`, it does not delete keys from the dictionary that were removed in the file.

moberst:
Clear current data (if any) in the document before loading new data in document.load().  This avoids an edge case where a document is being updated, and a key is removed.

Closes #895

### DIFF
--- a/papis/document.py
+++ b/papis/document.py
@@ -451,6 +451,7 @@ class Document(Dict[str, Any]):
                 "Error reading info file at '%s'. Please check it!",
                 self.get_info_file(), exc_info=exc)
         else:
+            self.clear()
             self.update(data)
 
 


moberst:
Note: Some tests fail when using `make pytest`, but those same tests seem to be failing currently regardless (related to `acl`, see output below)

```
================================================================================= short test summary info ==================================================================================
FAILED tests/downloaders/test_acl.py::test_acl_fetch[https://aclanthology.org/N04-1001] - KeyError: 'acl'
FAILED tests/downloaders/test_acl.py::test_acl_fetch[https://aclanthology.org/2021.naacl-main.208] - KeyError: 'acl'
====================================================== 2 failed, 261 passed, 10 skipped, 1 xfailed, 8 xpassed, 30 warnings in 40.92s =======================================================
```

alexfikl:
Thank you very much! This is good to go from my end if it works for you.

> Note: Some tests fail when using `make pytest`, but those same tests seem to be failing currently regardless (related to `acl`, see output below)

Hm, the CI seems to be working, so there might be something wrong with your dev environment? It's weird that just those two are failing though..

alexfikl:
Thank you again for the contribution! In it goes :rocket:

PR #911, opened 2024-09-27
Add PyPI release job

alexfikl:
This mostly uses the tutorial from here
https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

To work for actual releases, it needs:
* A `pypi` environment on Github: https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/managing-environments-for-deployment
* Adding the Github repo as a "trusted publisher" on PyPI: https://docs.pypi.org/trusted-publishers/adding-a-publisher/

I think we can add a step here to also make a Github release at the same time, but that would require a bit of scripting to get the changelog from `CHANGELOG.md`. Should be doable though.

@alejandrogallo @jghauser

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -56,14 +56,30 @@ jobs:
     - uses: actions/checkout@v4
       with:
         fetch-depth: 0
+
     - uses: actions/setup-python@v5
       with:
         python-version: '3.12'
+
     - id: dist
-      uses: casperdcl/deploy-pypi@v2
-      with:
-        build: true
-        upload: false
+      run: |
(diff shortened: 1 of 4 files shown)


alexfikl:
This also adds a step to make a Github release now:
* The release is named `Version v0.14`.
* It contains the info from the latest release from `CHANGELOG.md`.
* It also uploads the wheel and sdist, in case anybody wants them from here instead of from PyPI

alexfikl:
All of this is a bit experimental, so we might need to do a few `0.14rc` releases to make sure it all works as expected!

PR #914, opened 2024-09-29
default: tags should be a list

apraga:
With the new tag command, add only works if it's a list. papis doctor should fix that by default.

Thanks

### DIFF
--- a/papis/defaults.py
+++ b/papis/defaults.py
@@ -119,6 +119,7 @@ settings: Dict[str, Any] = {
                                    "author:str",
                                    "journal:str",
                                    "note:str",
+                                   "tags:list",
                                    "type:str",
                                    "publisher:str",
                                    "title:str",


alexfikl:
Agreed! Definitely time to start moving to lists for tags.

@jghauser I recall we talked about keeping some support for string tags, but it doesn't sound as appealing anymore :grin: Thoughts?

alexfikl:
@apraga We'll probably merge this, but how do you feel about adding a `doctor-key-type-check-keys-extend` configuration option? The keys there would just append to the default list. That should make it easier to add another key without the trouble of copy pasting the whole thing.

For an example, that's how `ruff` works with it's `select` and `extend-select`
https://docs.astral.sh/ruff/settings/#lint_select
https://docs.astral.sh/ruff/settings/#lint_extend-select

alexfikl:
> @alexfikl That's a good idea. The only thing that bothers me is the length of theses options. What do you think of renaming
> 
>  * `doctor-key-type-check-keys` -> `doctor-check-keys`
>  * `doctor-key-type-check-separator` -> `doctor-check-keys-separator`
>  * `doctor-key-type-check-keys-extend` -> `doctor-check-keys-extend`

The settings are currently named `doctor-[check-name]-[option]`. I'm not sure how that extra `check` got in there though..

In general, these should be named in a way that avoids potential collisions. For example, `doctor-check-keys` is very generic and we have many existing candidates (pretty much every doctor check wants to know what keys it should work on :(

We can work on shortening it a bit by allowing the `doctor` keys to be in a separate section (I don't think that works at the moment), i.e.
```
[doctor]
key-type-keys = ...
key-type-separator = ...
```
Would that be ok?

apraga:
Works for me, thanks !.

jghauser:
> @jghauser I recall we talked about keeping some support for string tags, but it doesn't sound as appealing anymore 😁 Thoughts?

I totally agree that tags should be lists. We already include functionality to convert string tags to lists, which should make the change easy for those currently using strings. I don't think there's any advantage to having tags as strings, and `papis tag` only supports lists.

I also really appreciate the `keys-extend` option as I've had the whole list of defaults duplicated because I wanted tags defined as lists. Although with this change I won't need to customise anything anymore :grinning:

alexfikl:
Awesome! Thanks for bringing this up and fixing it! In it goes :rocket:

PR #915, opened 2024-09-30
doc: add a 30s quickstart

apraga:
With the new quick guide, I felt a shorter version was needed with papis explained in 30s.
The presentation is a bit opinionated as I focused on the picker
Quick guide is moved to a tutorial.
The doc compiles locally. Any feedback is appreciated :)

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -16,9 +16,16 @@ Papis: Command-line Bibliography Manager
       :link-type: ref
       :class-card: getting-started
 
-    .. grid-item-card:: :material-regular:`rocket_launch;2em` Quick Guide
+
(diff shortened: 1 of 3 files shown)


apraga:
Hi,
That’s a good point. 2 things I struggled with at the beginning was 1. How to get my existing bibliography into papis ?2. How to navigate the bibliography quickly with different actions (adding notes, open PDF or URL mostly), like what you do in Zotero. The picker is papis strong point in my opinion but it’s not obvious that you can do a lot of things in the picker (and that you don’t have to type several commands). Point 1. is super important for helping users migrate to papis in my opinion. After thinking on it, would you be open to an updating the README quicktour with- import from BibTeX- picker actions And adding the picker key bindings in the quick guide ?.

Issue #918, opened 2024-10-01
FZF picker does not open notes


apraga:
* papis version (`$ papis --version` or commit number): 9b270726e271

# how to reproduce the issue
Command and output of `papis --log INFO <command>`:
Minimal reproducible example
```
papis add --from doi 10.1145/30355
papis add --from doi 10.1090/noti963 
 echo "a" > what-is-new-in-latex-vi.-latex-on-an-ipad.-empire/notes.norg
```
Then `ctrl-q` on this entry show
```❯ papis --log INFO edit *'
[WARNING] commands.edit: No documents retrieved
```
Using the default picker works.
Configuration
```
[settings]
default-library = papers
notes-name = notes.norg
add-folder-name = ``{doc[title]}``
doctor-key-type-check-keys = ['year:int', 'month:int', 'files:list', 'notes:str', 'author_list:list', 'doi:str', 'ref:str', 'isbn:str', 'author:str', 'journal:str', 'note:str', 'type:str', 'publisher:str', 'title:str', 'shorttitle:str', 'tags:list']
database-backend = whoosh
picktool = fzf

[papers]
dir = ~/papers2
```
Thanks !

alexfikl:
That picker seems to currently just have an edit and open commands, e.g.
https://github.com/papis/papis/blob/d5244da73c015a0a0f9958361a01a6255514138e/papis/fzf.py#L72-L82
Should be pretty easy to add another one to the list.

PR #919, opened 2024-10-01
fzf: add edit_notes as an action

apraga:
FZF only has open and edit. ctrl-q now does the same thing as the normal picker (edit a notes). Solves #918.


alexfikl:
This looks good to me once the CI failure is fixed. Did you test it works as expected? :grin:

alexfikl:
> Oops, sorry for not running the checks. I'm using it right now so it should be good to go !

No worries! Thanks for working on this. In it goes :rocket:

PR #922, opened 2024-10-12
Shortcut to browse_documents in pickers

apraga:
I often have to browse multiple entrie so a shortcut is welcome.

Not tested on fzf.


alexfikl:
Looks good to me! In it goes :rocket:

PR #930, opened 2024-10-30
papis.format: redefine Jinja2Formatter to enable easy template extension in config.py

pmiam:
Changes:

1. minimally redefine `Jinja2Formatter` to expose the jinja environment as a class variable so it can be easily accessed by the user config and freely modified
2. this environment is used to create template instances directly instead of using the jinja2 Template class.
3. when a `Jinja2Formatter` is initialized the env is "shared" so all format strings should share the same env just as they did before (this claim needs more verification)
4. call configpy with `exec(fd.read(), globals())` instead of `exec(fd.read())`

The main change is (4). From reading https://docs.python.org/3/library/functions.html#exec this should be a fully back-compatible change, but I'm not sure.

This is necessary to allow any reasonably complicated custom function to be registered with the Jinja2Formatter environment (easily, at least). Executing config.py in the top level namespace ensures all functions defined within are discoverable by the formatter. This is very much simpler than any alternative.

I was able to use these changes to implement (near enough) `bibtex-generate-autokey` from emacs' `bibtex.el` -- in my opinion, the best key generator I've used.

**USE CAUTION WHEN RUNNING FOREIGN CODE OF ANY KIND**. Here's a Gist of my config.py:
https://gist.github.com/pmiam/25b4232e9990cb251167bee7868e849b

I originally wrote this at the head of alexfikl's #711 branch, but it can stand alone.

### DIFF
--- a/papis/config.py
+++ b/papis/config.py
@@ -134,7 +134,12 @@ class Configuration(configparser.ConfigParser):
         configpy = get_configpy_file()
         if os.path.exists(configpy):
             with open(configpy) as fd:
-                exec(fd.read(), globals())  # add user's defs to global name space
+                # NOTE: this includes the `globals()` so that the user config.py
+                # can add entries to the global namespace. This was motivated
+                # by adding filters to `Jinja2Formatter.env`, which may be separated
+                # into multiple functions that would not be found otherwise.
+                #   https://github.com/papis/papis/pull/930
+                exec(fd.read(), globals())
 
 
 def get_default_settings() -> PapisConfigType:
--- a/papis/config.py
+++ b/papis/config.py
@@ -134,7 +134,7 @@ class Configuration(configparser.ConfigParser):
         configpy = get_configpy_file()
         if os.path.exists(configpy):
             with open(configpy) as fd:
-                exec(fd.read(), globals())
+                exec(fd.read(), globals())  # add user's defs to global name space
 
 
 def get_default_settings() -> PapisConfigType:
--- a/papis/config.py
+++ b/papis/config.py
@@ -134,7 +134,7 @@ class Configuration(configparser.ConfigParser):
         configpy = get_configpy_file()
         if os.path.exists(configpy):
             with open(configpy) as fd:
-                exec(fd.read())
+                exec(fd.read(), globals())
 
 
 def get_default_settings() -> PapisConfigType:
--- a/papis/format.py
+++ b/papis/format.py
@@ -6,6 +6,11 @@ import papis.plugin
 import papis.document
 import papis.logging
 
+try:
+    from jinja2.environment import Environment
+except ImportError:
+    pass
+
 logger = papis.logging.get_logger(__name__)
 
 FORMATTER: Optional["Formatter"] = None
@@ -202,6 +207,8 @@ class Jinja2Formatter(Formatter):
         "{{ doc.isbn | default('ISBN-NONE', true) }}"
     """
 
+    env: Environment = Environment()
+
     def __init__(self) -> None:
         super().__init__()
 
@@ -223,15 +230,14 @@ class Jinja2Formatter(Formatter):
         if additional is None:
             additional = {}
 
-        from jinja2 import Template
-
         fmt = unescape(fmt)
         if not isinstance(doc, papis.document.Document):
             doc = papis.document.from_data(doc)
 
         doc_name = doc_key or self.default_doc_name
         try:
-            return str(Template(fmt).render(**{doc_name: doc}, **additional))
+            tmpl = self.env.from_string(fmt)
+            return str(tmpl.render(**{doc_name: doc}, **additional))
         except Exception as exc:
             if default is not None:
                 logger.warning("Could not format string '%s' for document '%s'",
--- a/papis/format.py
+++ b/papis/format.py
@@ -8,8 +8,10 @@ import papis.logging
 
 try:
     from jinja2.environment import Environment
-except ImportError:
-    pass
+except ImportError as e:
+    exc = e
+else:
+    exc = None
 
 logger = papis.logging.get_logger(__name__)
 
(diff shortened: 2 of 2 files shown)


doc/source/default-settings.rst
```diff
@@ -196,6 +196,17 @@ General settings
...
+.. papis-config:: ref-word-separator
+    :type: str
+
+    This setting defines the separator between words in ref keys (usually
+    replacing spaces or other non-letter characters). By default this is the
+    underscore ``"_"`` but it could, e.g., also be the hyphen ``"-"``.
+
+    The ref is used as the bibtex key when creating bibtex format bibliographies.
+    Therefore, characters ``" # ' ( ) , = { } %`` are not recommended for use as
+    separators because neither ``bibtex`` nor ``biber`` can process them.
```
alexfikl, line 208: Mind making this a separate PR? It's unrelated to the jinja2 stuff.

papis/config.py
```diff
@@ -134,7 +134,7 @@ def initialize(self) -> None:
         configpy = get_configpy_file()
         if os.path.exists(configpy):
             with open(configpy) as fd:
-                exec(fd.read())
+                exec(fd.read(), globals())
```
alexfikl, line 137: Why is this needed? Maybe add a comment?

papis/format.py
```diff
@@ -202,17 +213,19 @@ class Jinja2Formatter(Formatter):
...
 
+    name: ClassVar[str] = "jinja2"
+    if not exc:
+        env: Environment = Environment()  # type: ignore
+    else:
+        logger.error(
+            "The 'jinja2' formatter requires the 'jinja' library. "
+            "To use this functionality install it using e.g. "
+            "'pip install jinja2'.", exc_info=exc)
+        raise exc
```
alexfikl, line 224: I'm not sure I like this "try..except" in the class body thing. It looks very hacky.

What do you think about just making this a cached function outside the class? Something like
```
@functools.lru_cache(maxsize=None)
def jinja_environment() -> Any:
	...
```
and then use that as a sort of singleton.

papis/format.py
```diff
@@ -33,6 +40,9 @@ def unescape(fmt: str) -> str:
 class Formatter:
     """A generic formatter that works on templated strings using a document."""
 
+    #: A name for the formatter.
+    name: ClassVar[str]
```
alexfikl, line 44: These are from #711. Remove them.

pmiam:
Thanks Alex. I like your suggestions. I'll take care of everything this weekend.

doc/source/default-settings.rst
pmiam: moved to #932

papis/format.py
pmiam: Thanks for the very pretty suggestion. I think it certainly makes better use of the try-except idiom.

The only drawback is that the type annotations are slightly less precise, but I think no harm done. I wish I could say what mypy's problem with the previous arrangement was, but this has no such issue.

pmiam: to elaborate, annotating the classvar `env` with `jinja2.environment.Environment` is no longer possible because `Environment` is stuck in the singleton's scope (I think). I'm not sure if there's a good way to pass it back to the module scope.

alexfikl: It should be fine to just annotate it with `Any` in this case. I'm not too bothered by that.

pmiam: I'll not admit how bothered I was. I looked into the issue and it seems python is just not aiming to support "forward reference" of types despite their best efforts. Additionally, regardless of the parser itself, mypy isn't a fan of my using either strings or NewType to attempt a stricter annotation. Any it is.

papis/config.py
```diff
@@ -134,7 +134,7 @@ def initialize(self) -> None:
         configpy = get_configpy_file()
         if os.path.exists(configpy):
             with open(configpy) as fd:
-                exec(fd.read())
+                exec(fd.read(), globals())  # add user's defs to global name space
```
alexfikl, line 137: I meant here to explain *why* this is needed, e.g. a usecase, not what adding `globals()` does. What did you need to add to the global name space?

alexfikl (commented):
I changed how the `env` is cached (again :(, sorry!). What do you think about this?

I tried it in `config.py` like
```python
from papis.format import Jinja2Formatter

env = Jinja2Formatter.get_environment()
env.filters["and_others"] = lambda s: f"{s} and others"
```
and it seemed to work as expected.

papis/config.py
alexfikl: Hm, I should really read the description more carefully :cry: 

> This is necessary to allow any reasonably complicated custom function to be registered with the Jinja2Formatter environment (easily, at least). Executing config.py in the top level namespace ensures all functions defined within are discoverable by the formatter. This is very much simpler than any alternative.

What exactly do you mean here? This works for me as well
```python
from papis.format import Jinja2Formatter

def and_others(s: str) -> str:
    return f"{s} and others"


env = Jinja2Formatter.get_environment()
env.filters["and_others"] = and_others
```

pmiam: Haha, you've made your point, both my long and short-form explanations are inadequate. Let's think how best to make this clear.

Try the following config without globals(), it should fail with a name not found error.

```
from papis.format import Jinja2Formatter

def _my_own_fstring(s: str) -> str:
    return f"{s} and others"

def and_others(s: str) -> str:
    return _my_own_fstring(s)

env = Jinja2Formatter.get_environment()
env.filters["and_others"] = and_others
```

alexfikl: Ah, I see! :facepalm:

alexfikl:
@pmiam I added a comment that made sense to me, so hopefully that works.

Otherwise, this looks good to go from my side! Let me know if everything is good from your side and we'll merge it in :grin:

pmiam:
> I added a comment that made sense to me, so hopefully that works.

The new comment is very clear. I also approve of the get_environment changes, it is much more clear what is going on. I did some manual tests using my new workflow. Merge looks good to go.

alexfikl:
> The new comment is very clear. I also approve of the get_environment changes, it is much more clear what is going on. I did some manual tests using my new workflow. Merge looks good to go.

Awesome! Thanks for working on this and sorry for the back and forth.

In it goes :rocket:

PR #932, opened 2024-11-03
papis.config: add configuration ref-word-separator

pmiam:
Changes:

1. introduced a configuration option `ref-word-separator` (default "_")
2. I just think enforcing a word separator in refs is too opinionated ("-" looks better in the latex context)

I'm open to alternative ways of doing this, but I reckon other approaches would require more invasive revision of tests.

### DIFF
--- a/doc/source/default-settings.rst
+++ b/doc/source/default-settings.rst
@@ -203,7 +203,7 @@ General settings
     replacing spaces or other non-letter characters). By default this is the
     underscore ``"_"`` but it could, e.g., also be the hyphen ``"-"``.
 
-    The ref is used as the bibtex key when creating bibtex format bibliographies.
+    The ``ref`` is used as the citation key when creating BibTeX format bibliographies.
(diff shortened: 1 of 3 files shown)


doc/source/default-settings.rst
```diff
@@ -196,6 +196,17 @@ General settings
...
     hyphen ``"-"`` but it could, e.g., also be the underscore ``"_"``.
 
+.. papis-config:: ref-word-separator
+    :type: str
+
+    This setting defines the separator between words in ref keys (usually
+    replacing spaces or other non-letter characters). By default this is the
+    underscore ``"_"`` but it could, e.g., also be the hyphen ``"-"``.
+
+    The ref is used as the bibtex key when creating bibtex format bibliographies.
```
alexfikl, line 206:
```suggestion
    The ``ref`` is used as the citation key when creating BibTeX format bibliographies.
```

alexfikl (approved):
Added a little suggestion, but otherwise this looks good to go!

PR #941, opened 2024-11-08
ci: parse version with python instead of awk

### LINKED ISSUE #940: CI: Windows build has flaky version parsing
It currently does something like this
```
git describe --tags --dirty \
	| awk -F '-' '{ printf "%s.dev%s+%s\n", substr($1, 2), $2, $3 }'
```
because it expects a git output like `v0.13-533-gab7488cd`. 

* [ ] This breaks if the current commit is the last tag (for which we just get `v0.13`).
* [ ] Also breaks if the version is something like `0.14rc1`, which is allowed by Python but not for MSI versions.

It would be nicer to make this more robust. Example failed workflow
https://github.com/papis/papis/actions/runs/11747502229/job/32729554402

cc @kiike

kiike:
This PR addresses the flakiness for the MSI version parsing. I think using this could do the trick.

The version for pre-releases would be in the format of MAJOR.MINOR.MICRO-LABEL.ID such as 0.14.0-rc.3 (like it was an hour ago), whereas for release versions or versions built during CI would be MAJOR.MINOR.MICRO.COUNT such as 0.14.0.0 for cut releases or 0.14.0.42 where 42 corresponds with the 42 in v0.14-**42**-dev000000 like in [this action run](https://github.com/papis/papis/actions/runs/11749836574/job/32736871128?pr=941#step:10:14).

Fixes #940.

Should I change the MSI artifact filenames too so that the filename coincides with the MSI version too?

cc @alexfikl

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -38,15 +38,9 @@ jobs:
       shell: bash
       run: |
         # NOTE: python version follows "X.YdevZ+gHASH" format
-        py_version=$(
-            git describe --tags --dirty \
-            | awk -F '-' '{ printf "%s.dev%s+%s\n", substr($1, 2), $2, $3 }'
-        )
+        py_version=$(python tools/windows/get_version.py python)
         # NOTE: MSI version cannot have letters and is only "X.Y.Z"
-        msi_version=$(
-            git describe --tags --dirty \
-            | awk -F '-' '{ printf "%s.%s", substr($1, 2), $2 }'
-        )
+        msi_version=$(python tools/windows/get_version.py msi)
         sed -i "s/^version =\".*\"/version =\"${py_version}\"/" pyproject.toml
 
         echo "PAPIS_VERSION=${py_version}" >> "$GITHUB_ENV"
--- a/tools/windows/get_version.py
+++ b/tools/windows/get_version.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python
+
+from packaging.version import Version
+import subprocess
+from typing import Literal
+
+import pytest
+import click
+
+
+Mode = Literal["python", "msi"]
+
+
+def get_version(description: str, mode: Mode) -> str:
+    # When the command returns something like v0.14
+    if "-" in description:
+        raw_version, count, commit_hash = description.split("-")
+    # Otherwise, like v0.14-1-f00b4r
+    else:
+        raw_version = description
+        count = "0"
+        commit_hash = ""
+
+    version = Version(raw_version)
+
+    if mode == "python":
+        # `git describe --tags` returns v.0.14
+        if count == "0":
+            return raw_version
+        # `git describe --tags` returns v0.14-1-f00b4r
+        else:
+            return f"{raw_version}+dev{count}+g{commit_hash}"
+
+    # MSI versions can be either X.Y.Z.W or X.Y.Z-LABEL.W
+    elif mode == "msi":
+        if version.is_prerelease and version.pre is not None:
+            prerelease_id = f"{version.pre[0]}.{version.pre[1]}"
+            # This yields, for example, 0.14.0-rc.1
+            return f"{version.major}.{version.minor}.{version.micro}-{prerelease_id}"
+        else:
+            # 0.14.0.X where X is the commit count after last the tagged commit
+            return f"{version.major}.{version.minor}.{version.micro}.{count}"
+
+
+@pytest.mark.parametrize(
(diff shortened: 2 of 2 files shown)


alexfikl:
Thanks for taking care of this so quickly! :heart:

I think we also need to also update the parsing for the `py_version` variable. On the current commit 980ce4c5, the result of `git describe` is just `v0.14` and so `py_version=0.14.dev+` which I don't think is valid?

tools/windows/get_version.py
```diff
@@ -0,0 +1,75 @@
...
+import subprocess
+from typing import Literal
+
+import pytest
+import click
+
+Mode = Literal["python", "msi"]
+
+
+def get_version(description: str, mode: Mode) -> str:
```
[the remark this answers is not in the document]
kiike: An else branch would solve this but it would be unreachable as click would protect against passing an unavailable mode?

.github/workflows/windows.yml
```diff
@@ -38,15 +38,9 @@ jobs:
       shell: bash
       run: |
         # NOTE: python version follows "X.YdevZ+gHASH" format
-        py_version=$(
-            git describe --tags --dirty \
-            | awk -F '-' '{ printf "%s.dev%s+%s\n", substr($1, 2), $2, $3 }'
-        )
+        py_version=$(python tools/windows/get_version.py python)
```
kiike, line 41: @alexfikl: taking care of setting the python version now

tools/windows/get_version.py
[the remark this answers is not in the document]
alexfikl: Yeah, that's fine. My usual workaround for this is
```
else:
	raise AssertionError
```

tools/windows/get_version.py
```diff
@@ -0,0 +1,76 @@
...
+    # When the command returns something like v0.14
+    if "-" in description:
+        raw_version, count, commit_hash = description.split("-")
+    # Otherwise, like v0.14-1-f00b4r
+    else:
```
alexfikl, line 19:
```suggestion
    if "-" in description:
    	# Expected version like v0.14-1-f00b4r
        raw_version, count, commit_hash = description.split("-")
    else:
    	# Otherwise version like v0.14
```
Think these comments were reversed?

tools/windows/get_version.py
```diff
@@ -0,0 +1,76 @@
...
+
+    version = Version(raw_version)
+
+    if mode == "python":
+        # `git describe --tags` returns v.0.14
+        if count == "0":
+            return raw_version
+        # `git describe --tags` returns v0.14-1-f00b4r
+        else:
+            return f"{raw_version}+dev{count}+g{commit_hash}"
```
alexfikl, line 32:
```suggestion
            return f"{raw_version}.dev{count}+g{commit_hash}"
```

tools/windows/get_version.py
```diff
@@ -0,0 +1,76 @@
...
+        # Release candidates
+        ("v0.14rc3", "python", "v0.14rc3"),
+        ("v0.14rc3", "msi", "0.14.0-rc.3"),
+        # A commit after release
+        ("v0.14-1-g6d2f2957", "python", "v0.14+dev1+gg6d2f2957"),
+        ("v0.14-1-g6d2f2957", "msi", "0.14.0.1"),
+    ],
+)
+def test_get_version(git_description: str, mode: Mode, expected: str) -> None:
+    assert get_version(git_description, mode) == expected
```
alexfikl, line 60: This is very nice, but probably not needed here? :grin:

alexfikl (approved):
Left a few nitpicks, but otherwise this looks good to me! Thanks again.

tools/windows/get_version.py
kiike: Absolutely! Thanks!

tools/windows/get_version.py
kiike: I don't know. Let me know what you want me to remove. All the tests?

alexfikl: I just meant that they're not being run in the CI, but it's fine to keep them since I don't imagine this will see a lot of breaking changes.

alexfikl:
This looks good to merge on my side! Thanks again for taking care of it so quickly :rocket:

kiike:
Happy to help! Not using papis anymore, or bibliography reference managers for that matter, but I'll keep an eye on notifications!

kiike:
And thanks to you for your attentive reviews  :grin:

Issue #946, opened 2024-11-10
pdfjs: Try using system installation of pdfjs as well


hseg:
* papis version (`$ papis --version` or commit number): 807f65ab

# how to reproduce the issue
```
$ pacman -Q pdfjs 
pdfjs 4.8.69-1
$ ls /usr/share/pdf.js/
‘build’  ‘LICENSE’  ‘web’
$ papis --log INFO serve
[WARNING] commands.serve: 
No installation of pdfjs found.

If you want to be able to read and see PDF files from within
the Papis web applications you need to install pdfjs.

You can download the PDFJS bundle from the url

    https://github.com/mozilla/pdf.js/releases/download/v3.1.81/pdfjs-3.1.81-legacy-dist.zip

and extract it to your config directory under 'web/pdfjs', for instance
in '~/.config/papis/web/pdfjs/'.

On linux and mac you can simply run the following lines

    mkdir -p ~/.config/papis/web/pdfjs
    cd ~/.config/papis/web/pdfjs
    wget "https://github.com/mozilla/pdf.js/releases/download/v3.1.81/pdfjs-3.1.81-legacy-dist.zip" -O pdfjs.zip
    unzip pdfjs.zip
    rm pdfjs.zip
    
[INFO] commands.serve: Starting server in address 'http://localhost:8888'.
[INFO] commands.serve: Press <Ctrl-C> to exit.
[INFO] commands.serve: THIS COMMAND IS EXPERIMENTAL, expect bugs. Feedback appreciated!
```

Is there any reason we require pdfjs to be in `$XDG_CONFIG_HOME/papis/web`? I can  somewhat understand using that as a fallback for users in restricted environments (though I have a hard time imagining a user able to install `papis` but not `pdf.js`), but why aren't we picking up the system's installation of `pdf.js`?

alexfikl:
The code that looks for `pdf.js` should be here
https://github.com/papis/papis/blob/f0a5b16fcb768b25c21adf7e192415d3bb9cbb52/papis/web/pdfjs.py#L42-L47
A lot of the CSS and JS for the webapp is defined in the config file, so this should probably go in there as well, so that it's easy to point to something else. I'm not sure how to automatically find a system installation though..

@alejandrogallo has some plans to redo the webapp in `flask`, so there's a good chance a lot of this will be heavily changed for the next big release.

> As you might tell, I'm rechecking the dependency list in light of the 0.14 release.

That's very welcome! Thank you!

[...]

PR #947, opened 2024-11-13
Align code structure of epubjs.py and pdfjs.py

hseg:
Following up on
https://github.com/papis/papis/issues/946#issuecomment-2473620039, I realigned
the code for `epubjs` and `pdfjs`, in case it's useful.

### DIFF
--- a/papis/web/epubjs.py
+++ b/papis/web/epubjs.py
@@ -1,15 +1,45 @@
 import os.path
+import urllib.parse
 
 import dominate.tags as t
 
 import papis.web.html as wh
 import papis.web.static
 
-VIEWER_PATH = "epubjs-reader/reader/index.html"
 EPUBJS_URL = "https://github.com/futurepress/epubjs-reader"
+VIEWER_PATH = "epubjs-reader/reader/index.html"
 
 
-def detect_local_installation() -> bool:
+def widget(unquoted_file_path: str) -> None:
+    """
+    Widget for epub files.
+    """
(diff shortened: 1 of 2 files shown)


alexfikl:
This seems to have a bunch of whitespace/formatting changes. What are the actual changes in here?

hseg:
`epubjs.py`:
- Import `os` instead of `os.path` (though in retrospect, the reverse would be clearer)
- Quote the filename passed to the reader
- Rename `detect_local_installation` -> `detect_epubjs`
- Swap order of `EPUBJS_URL`/`VIEWER_PATH` and move `widget` to top of function list`

`pdfjs.py`:
- rename `_unquoted_file_path` -> `unquoted_file_path`
- remove line breaks from a couple of elements
- copy+adapt docstrings from `widget` and `error_message`

(btw, viewing the patch with `git show --color-moved` would've shown some of this. But then again, I should've added this in my initial message)

hseg:
There, I reversed the `os.path` -> `os` change -- now it's `pdfjs.py` that's importing `os.path` -- which is clearer in re what the import is for.

alexfikl:
> (btw, viewing the patch with git show --color-moved would've shown some of this. But then again, I should've added this in my initial message)

Haha, yeah, that's fair. I was lazily just looking at the Github diff, which isn't the best..

Anyways, thanks for explaining! That all sounds good to me, so in it goes :rocket:

PR #956, opened 2024-11-30
remove openfile from collect_importer_data

SnakeOnex:
So for some reason, in the `collect_importer_data` function in `utils.py`, when a pdf from arxiv is download it is always opened. This overwrites the `add-open` configuration and the `--open / --no-open` args for `papis add`. I removed the line.

### DIFF
--- a/papis/utils.py
+++ b/papis/utils.py
@@ -561,7 +561,6 @@ def collect_importer_data(
 
             msg = f"Use this file? (from {importer.name})"
             for f in importer.ctx.files:
-                open_file(f)
                 if batch or confirm(msg):
                     ctx.files.append(f)
 


alexfikl:
I'm not sure removing `open_file` completely is the way to go here. 

From what I recall, the point of that was to let the user see the file that was downloaded and confirm it should be added. E.g. if you don't have access to some journal, it could download a dummy HTML file (instead of the actual PDF file) and add that to the document.

What do you think about #957 instead?

SnakeOnex:
sure could be, my change is the most minimal, yours adds another flag, which arguably collides with `--open-file` a little bit, my assumption was that the `add-open` option was made for exactly this, to make sure the file is downloaded correctly. What happens now is that if you set `add-open`, the file gets opened twice, which is not ideal, so just anything that fixes this I am for. What if add-open and your confirm-files are both set to true? will this open it twice? why not just give user the option to set `add-open` if he wants to check the file was downloaded correctly?

alexfikl:
> sure could be, my change is the most minimal, yours adds another flag, which arguably collides with `--open-file` a little bit, my assumption was that the `add-open` option was made for exactly this, to make sure the file is downloaded correctly. What happens now is that if you set `add-open`, the file gets opened twice, which is not ideal, so just anything that fixes this I am for.

Hm, I agree that it's a bit confusing, so might need some additional reworking (the `--confirm` flag is not new though). 

Currently, the `--open-file` flag doesn't really do anything that useful. It just opens all the files that are added to the document, but doesn't let the user choose if they're ok. Ideally that part of the code would: open the file (if `--open-file` is used), ask for confirmation (if `--confirm` is used) and only keep the file if accepted. How does that sound?

SnakeOnex:
yeah i like that, the part that i didn't like the most was that `papis add` would open the file, so i would pass the `--no-open` flag and it would still open no matter what, combining `--confirm` and `--open` sounds like the way to go

alexfikl:
> yeah i like that, the part that i didn't like the most was that `papis add` would open the file, so i would pass the `--no-open` flag and it would still open no matter what, combining `--confirm` and `--open` sounds like the way to go

I've updated #957 to basically do that now. Can you try it out?

SnakeOnex:
#957 seems good! closing this

PR #957, opened 2024-11-30
Confirm when adding files

alexfikl:
This updates how `papis add` works when adding files:
* Files from importers are all included. Before, each importer asked for confirmation for each file.
* The `--open` flag opens files before they're added to the document for checking. Before, it would open them afterwards.
* The `--confirm` flag asks the user to add the file after it's been opened. Before, it would only ask for confirmation on downloaded files.

This way, when `--no-open` and `--no-confirm` is set, the files are just added without opening anything, as expected. Before, `collected_importer_data` would still open them up annoyingly.

xref: #956.

### DIFF
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -283,14 +283,22 @@ def run(paths: List[str],
     # rename all the given file names
     from papis.paths import symlink, rename_document_files
 
-    new_file_list = rename_document_files(tmp_document, in_document_paths)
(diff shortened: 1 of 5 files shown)


SnakeOnex:
I tried it out and seems to work as expected.

alexfikl:
> I tried it out and seems to work as expected.

Thanks a lot for helping to working through this! In it goes :rocket:

PR #971, opened 2025-03-01
Switch from flake8 to ruff

alexfikl:
This switches to ruff for all our static linting needs: it replaces `flake8` and also enables some `pylint` checks (ruff doesn't have all of them implemented yet).


alexfikl:
@f0rki @jghauser Any help with the `flake.nix` failing? It doesn't seem to find ruff, but I'm not sure how to add it.. :(

jghauser:
@alexfikl I think the problem is that our nixpkgs is pinned at too old a version so that it's set to use python 3.11. But at the time, `ruff` wasn't in the python package set (e.g. `pkgs.python3Packages.ruff` but only at the top-level `pkgs.ruff`). This has been fixed by upstream in the meantime. So, when I update the `flake.lock` with `nix flake update`, it all works nicely :). So if you got nix, try that command, otherwise I can commit a new `flake.lock` myself.

alexfikl:
> So, when I update the `flake.lock` with `nix flake update`, it all works nicely :). 

Thanks for the help! Just pushed a version with an updated `flake.lock`. Let's see if that works :-?

jghauser:
Glad it worked and happy to help! :grinning:

alexfikl:
I'll go ahead and merge this! Besides the obvious, the main push is that packaging all the flake8 plugins seems to be a bit of pain https://github.com/papis/papis/discussions/1037.

cc @hseg

hseg:
Thanks, pushed the update to AUR, cc @tuurep

The flake8 plugins were symptomatic of the fragility of the status quo, but they're not in themselves the main issue. Unless I use an AUR helper like @tuurep does (and which maybe might be a necessary addition to my workflow), I'd need to track down and download each of the following (from `~/aur/9+deps/papis-git/`):
```
papis-git-materialized/
├── check
│   ├── python-markdownify
│   ├── python-types-beautifulsoup4
│   ├── python-types-docutils
│   ├── python-types-pygments
│   ├── python-types-pyyaml
│   └── python-types-requests
├── dep
│   ├── python-arxiv
│   ├── python-doi
│   ├── python-habanero
│   └── python-isbnlib
├── make
│   └── python-sphinx-design
└── opt
    ├── python-citeproc-py
    │   └── make
    │       └── rnc2rng
    │           └── python-rply
    └── python-markdownify
```

~~TBH, my experience was that it's the deep branch on `python-arxiv` that was the more annoying of all the dependencies. So either more of these packages need to be picked up by the official Arch repos, or the projects need to do a better job of not using exotic deps.~~

~~… and now I inspect it, it's clear the `python-arxiv` `PKGBUILD` is unnecessarily adding `python-pip-audit` to its `checkdepends` -- it is a dev dependency. Brb, going to comment that on the AUR page (but this itself is indicative of the trouble relying on AUR packages).~~

EDIT: … I should know by now to check if a mess-up isn't my fault before casting stones. Turns out the mistaken addition of `python-pip-audit` is my fault. Thankfully, the packager hasn't listened to me, so everybody else's experience is better than mine.

The point still stands that this is quite a lot of AUR-based dependencies, but at least only one of them is transitive.

<detail>
<summary>The `python-arxiv` dependency tree my mistake ended me with</summary>

```
python-arxiv
└── check
    └── python-pip-audit
        └── deps
            ├── python-cyclonedx-lib
            │   └── deps
            │       ├── python-packageurl
            │       └── python-py-serializable
            ├── python-pip-api
            └── python-pip-requirements-parser
```
</detail>

alexfikl:
> … and now I inspect it, it's clear the `python-arxiv` `PKGBUILD` is unnecessarily adding `python-pip-audit` to its `checkdepends` -- it is a dev dependency. Brb, going to comment that on the AUR page (but this itself is indicative of the trouble relying on AUR packages).

Yeah, I've gone into the habit of more-or-less forking any AUR package I need and putting them in a local repo. It's not great, but at least there are less surprises.. :(

hseg:
Same here -- have ended up with local edits on 68/106 AUR packages I have installed, ranging from just bumping pkgvers when maintainers are slow to maintaining a local wishlist of changes I'd make if I were to adopt the package (some of these are minor enough and the maintainers opinionated enough that it's not worth the fight).

(though you may have missed my edit noting that the `python-arxiv` problem is of my own making)

tuurep:
Nice, seems like a great step forward, right?

For the first time, I was able to `yay -S papis-git` without the build failing.

Regards helper usage: the most trouble I've had was with a [windows 10 fonts package](https://aur.archlinux.org/packages/ttf-ms-win10-auto) with a lot of arguing in the comments whether `yay` is acceptable to use or not.

hseg:
> a great step forward

Yeah, reevaluating after taking into account that the big dependency on `python-pip-audit` was of my own making, it's definitely an improvement -- while I hope my reports on `flake8` will change policy going forward, the fact that `ruff` bundles up all these plugins in a single package makes it less fragile in this regard.

Plus, it's many fewer AUR dependencies, which is always a win.

> whether `yay` is acceptable to use or not

Reading through that comment thread, it seems that it's not so much a question of whether `yay` is an acceptable tool to use (that's a user preference), but rather whether package maintainers are under obligation of the AUR social contract to make sure their packages build under `yay`. In particular, the question is whether package maintainers need to take "it doesn't build with `yay`" into consideration, or whether they can legitimately respond "WFM in a clean `chroot`" and be done with it.

Based on eg [AW:AUR helpers], [AW:AUR], [AW:Packaging standards], I'd say the official position is that it's up to the maintainer's discretion.

And from my experience with eg #1037, I'd agree -- one difficulty as a maintainer with accepting feedback over "it doesn't build with `yay`" is that it hides *which* PKGBUILD is at fault for the breakage -- after all, you posted that issue on this bugtracker, not the `python-flake8-pyproject` or `python-pycodestyle` bugtrackers.

`yay` agrees with this, as per their README (though it's buried 90% down the page and could be made more prominent)
> … Since Yay is not officially supported by Arch Linux, support should not be sought out on the forums, AUR comments or other official channels.
> … A package may only be considered broken if it fails to build with makepkg.
> Reports should be made using makepkg and include the full output as well as any other relevant information. Never make reports using Yay or any other external tools.


[AW:AUR helpers]: https://wiki.archlinux.org/title/AUR_helpers
[AW:AUR]: https://wiki.archlinux.org/title/Arch_User_Repository
[AW:Packaging standards]: https://wiki.archlinux.org/title/Arch_packaging_standards

PR #974, opened 2025-03-01
Remove magic-nix-cache-action from ci

alexfikl:
The cache action does not seem to be available anymore according to https://determinate.systems/posts/magic-nix-cache-free-tier-eol/. I'm not sure if there's an alternative, but for now this just removes it..

cc @f0rki @jghauser

### DIFF
--- a/.github/workflows/nix-flake.yml
+++ b/.github/workflows/nix-flake.yml
@@ -19,12 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    # - uses: cachix/install-nix-action@v22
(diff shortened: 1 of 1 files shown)


jghauser:
Not really sure about this, I'm not familiar with this CI action. The post mentions using [FlakeHub Cache](https://flakehub.com/cache) and applying for a fee exemption as an opensource project, though I'm not sure whether it's worth the effort/added complexity. Removing for now seems to make sense!

alexfikl:
There's also [cache-nix-action](https://github.com/nix-community/cache-nix-action) that may work, but it requires some reading up. I'll merge this in the meantime.

PR #976, opened 2025-03-15
Add basic CSL exporting

alexfikl:
This adds some basic support for exporting citations using the CSL library [citeproc-py](https://github.com/citeproc-py/citeproc-py). It seems like development has picked up a bit over there, so it can't hurt to add it.

It still seems to have a few bugs to iron out, but hopefully they'll get to that by the time we make a new release as well. In particular the popular APA and MLA styles don't seem to be working :(

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # VERSION NEXT (TDB)
 
+## Dependency Changes
+
+- The [citeproc-py](https://github.com/citeproc-py/citeproc-py) library is now
+  required for CSL exporter support.
+
 ## Features
 
 ### Major: Improved formatter support ([#711](https://github.com/papis/papis/pull/711))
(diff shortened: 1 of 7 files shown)


alexfikl:
I'm not a big user of this CSL stuff, so I just did some very quick and dirty testing with some styles. 

If anyone has more interest in it, any testing or review is very welcome! Otherwise, I'll probably go ahead and merge it in a week or so and we can see how it goes.

alexfikl:
In it goes :rocket:

hseg:
FYI, I've been trying to package `citeproc-py` and its transitive dependencies for the last couple of hours. rnc2rng has failing tests, and rply seems to have been abandoned 5y ago. It's frustrating.

hseg:
... Correction, rnc2rng's tests are weirdly written (they need to be invoked by running them directly, instead of eg `python -m unittest discover`

alexfikl:
> FYI, I've been trying to package `citeproc-py` and its transitive dependencies for the last couple of hours. rnc2rng has failing tests, and rply seems to have been abandoned 5y ago. It's frustrating.

Yeah, `citeproc-py` itself barely got out of a slump, so it's not exactly up to date with things. I'm planning to help out a bit, but in the meantime we should make it optional for papis.

hseg:
Forgot to update - I ended up managing to package the series of dependencies, but they each had a bit of nontriviality. Please consider this next time you introduce a hard dependency.

31 mar 2025 23:32:32 Alex Fikl ***@***.***>:

alexfikl:
> Forgot to update - I ended up managing to package the series of dependencies, but they each had a bit of nontriviality. Please consider this next time you introduce a hard dependency.

Will try to pay more attention to that. However, this is not in any released version and the point is to see if it works at all until then (from comments like yours, which are very appreciated!).

hseg:
Fair enough - I _am_ using my nightly package as a way to incrementally keep on top of the changes introduced.

31 mar 2025 23:40:34 Alex Fikl ***@***.***>:

Issue #981, opened 2025-03-17
python-coveralls dependency


jghauser:
I was tidying up the project's flake.nix as a lot of packages we include there manually are now in `nixpkgs`. I came across the [`python-coveralls`](https://github.com/z4r/python-coveralls) dependency, which seems to have last been updated 6 years ago. What do we use it for? I can't seem to find any reference to it in the repo (except as dependency). If we need it, there seems to be the (maintained) [`coveralls-python`](https://github.com/TheKevJames/coveralls-python) package that does something similar?

alexfikl:
Hm, yeah, that really hasn't been used in a long time. My understanding is that that would feed into [coveralls.io](https://coveralls.io/)?

I think it's safe to remove that dependency until we look into getting some coverage statistics back. I imagine Github / Github Actions have something more builtin these days (?).

jghauser:
great, thanks! I'll push a PR with the updated flake and will remove this dependency

PR #982, opened 2025-03-17
Update flake

jghauser:
This updates the flake. We no longer build dependencies that are now in nixpkgs and I've updated those dependencies we still need to build ourselves. It also removes `python-coveralls` from wherever I found it mentioned, since we don't seem to be using anymore (see #981).

### DIFF
--- a/contrib/python-papis.scm
+++ b/contrib/python-papis.scm
@@ -242,7 +242,6 @@ an elegant DOM API.")
                       python-pylint
                       python-pytest
                       python-pytest-cov
-                      python-coveralls
                       python-lsp-server
                       python-sphinx-click
                       python-sphinx-rtd-theme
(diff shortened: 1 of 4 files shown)


alexfikl:
Looks good to me! :rocket: 

Not sure why the `usenix` downloader test is failing. They might have changed something in their webpages.. if you don't get to it, I can take a look tomorrow :\

jghauser:
Ok I'll merge this given that the errors are unrelated! Thanks for having a look :)

alexfikl:
> Ok I'll merge this given that the errors are unrelated! Thanks for having a look :)

Seems like that was just a temporary issue on their side. The tests seem to be passing now
https://github.com/papis/papis/actions/runs/13906372523

PR #983, opened 2025-03-18
Bump minimum Python version to 3.10

### LINKED ISSUE #978: Bump minimum Python version to 3.10
@jghauser @alejandrogallo What do you think? 

Python 3.8 is EOL (since October last year) and Python 3.9 is nearing its end of security releases as well, so Python 3.10 seems like a good one to jump to :grin: I'm not sure if we have any users on 3.8 only, so if anyone sees this, please yell!

Some cool new toys from jumping to 3.10:
https://docs.python.org/3/whatsnew/3.9.html
https://docs.python.org/3/whatsnew/3.10.html

Personally, I mostly appreciate the typing changes that remove a bunch of imports.
* In 3.9, can use `tuple[int, ...]` instead of `Tuple[int, ...]`
* In 3.10, can use `int | str` instead of `Union[int, str]`.

alexfikl:
This just bumps the minimum python version to 3.10 and ~does not do any code modification / upgrades to use some newer features~ fixes all the `ruff` issues.

Fixes #978.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,7 +29,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.10', '3.11', '3.12', '3.13']
       fail-fast: false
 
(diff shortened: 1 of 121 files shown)


alexfikl:
In it goes! :rocket:

PR #986, opened 2025-03-20
docs(readme): mention papis-ask instead of papis-qa

jghauser:
My papis-ask plugin is now in a reasonably ok state. I can index my whole library without errors and querying seems to work very well. @isaksamsten says [he would be archiving his repo](https://github.com/isaksamsten/papisqa/issues/1), so I'm replacing his papis-qa with my papis-ask in the README.


jghauser:
Thanks I'll change that too!

PR #987, opened 2025-03-23
Some general cleanup and refactoring for database backends

alexfikl:
I wanted to implement a backend based on sqlite and managed to amass a bunch of random cleanup stuff while looking through it. There's quite a few changes, but most of them are just moving things around.

Added some deprecations for the next next release (i.e. 0.16):
* Deprecated `papis.id.key_name()` in favour of a constant `papis.id.ID_KEY_NAME`.
* Deprecated `papis.id.has_id()` in favour of just using `ID_KEY_NAME in doc`.
* Deprecated `Database.get_lib`, `Database.get_dirs` and `Database.get_id_key` in favour of just accessing the members directly.
* Removed `Database.match` since it didn't seem to be used anywhere and was only implemented by the `papis` backend.

Some other changes:
* Made a bunch of methods "private" with a little underscore. I don't think these were ever meant to be used outside of the `Database` classes.
* Rearranged all the methods in `papis.database.cache` and `whoosh` to match the ones in `base`. I was mostly confused trying to match which abstractmethods were implemented where.
* Updated and added docs for all the methods. They were also added to the `Developer API reference` in the docs in case anyone wants to look through them in there.

I've ran the entire test suite locally with the `papis` backend and the `whoosh` backend and it all seems to work (after some fixes to the unittests).

PR #990, opened 2025-03-24
Run tests for each database

alexfikl:
This adds another job to the CI that runs all the tests with each non-default database we have to make sure they still work. The main motivation for this is that I want to add another database backend based on `sqlite` and it would be nice to stress test it a bit with this.

In #987, I found quite a few tests that were failing with the default setup for `whoosh`, so this should avoid those sneaking in in the future. This seems to still be the case, so will fix those later :(

TODO:
- [X] Note that each database backend has its own set of dedicated tests, but most tests are run with the default `papis` backend. Is it worth running these on all the OSes / Python versions?
	- Just running the tests on the latest OS and latest Python versions for now. We have enough jobs as it is :(

### DIFF
--- a/doc/source/testing.rst
+++ b/doc/source/testing.rst
@@ -10,6 +10,25 @@ tested using their testing helpers.
 We give here an overview of the pieces needed to test the various parts of the
 Papis codebase, but in general mimicking existing tests is your best choice.
 
+Environment variables
+---------------------
+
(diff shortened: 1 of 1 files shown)


alexfikl:
Turns out the remaining issue that caused fails here was just due to `default-query-string` config not switching to the database. I'll go ahead and merge this as soon as the tests pass!

PR #991, opened 2025-03-28
Document optional dependencies in PACKAGING

alexfikl:
This adds a small description for each optional dependency we have, so people can decide if they want it.

cc @hseg

### DIFF
--- a/PACKAGING.md
+++ b/PACKAGING.md
@@ -19,6 +19,19 @@ Dependencies
 ------------
 
 See `pyproject.toml` for a complete list of Python dependencies and minimum versions.
+Some optional dependencies are required for various Papis plugins
+
+* `Jinja2` is required for the `jinja` formatter, alternative to the default `python`
+  formatter.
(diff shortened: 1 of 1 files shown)


hseg:
LGTM!

hseg:
(It might be nice to advertise that "some plugins might use the semi-public internal modules papis.sphinx_ext and papis.testing, look at the extras by those names in pyproject.toml for their dependencies")

28 mar 2025 11:37:29 Alex Fikl ***@***.***>:

alexfikl:
> (It might be nice to advertise that "some plugins might use the semi-public internal modules papis.sphinx_ext and papis.testing, look at the extras by those names in pyproject.toml for their dependencies")

This was mostly meant to point more exactly to what the optional dependencies help with, not so much to mention their existence. Your comments in `pyproject.toml` should do that just fine, plus there's also various documentation for them (e.g. https://papis.readthedocs.io/en/latest/testing.html).

PR #993, opened 2025-03-30
ci: remove unused types-python-slugify from flake

jghauser:
I realised that after the recent cleanup, there's now an unused `types-python-slugify` in our flake. This PR removes it.


alexfikl:
Oops, didn't notice that was still there. Looks good to me! :rocket:

PR #985, opened 2025-03-19
Doc: Cleanup pyproject deps

hseg:
- **clean: Remove stale deps**
  Testing reveals types-slugify is no longer needed

- **doc: Document optional features**
  For this purpose, also split develop into style and test to make the components easier to document. The develop feature is left in place for ease of setup, other dev dependency components can be added to it as needed.

- **doc: Document deps for plugins using papis modules**
  This way, packagers have a single source of truth to consult.

Closes: #984

PR #994, opened 2025-03-30
Doc: Move LSP advertisement from pyproject.toml to docs

hseg:
As promised in https://github.com/papis/papis/pull/985#issuecomment-2759768177, here's an attempt to reduce the presence of `python-lsp-server` to merely an advertisement and a convenience feature in `flake.nix` (dropped from that PR since polishing this up was questionably-useful scope creep).

To recap the motivation for this: It isn't a dependency for the project, but merely a useful-to-have tool we recommend developers install. (This is unlike linters and test suites that allow us to maintain project standards -- in contrast, an LSP server is one way to configure a text editor for better Python support). In view of this, it doesn't make sense to advertise this as a component of the project (as is the implication by putting it in `pyproject.toml`).

Hopefully I've guessed the nix syntax correctly this time. Otherwise, I'll set this as a draft and withdraw from it -- I have little free time this month and do not want to commit to polishing this to a releasable state.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -43,6 +43,10 @@ The docs can be generated with
 make doc
 ```
 
+It is generally advisable to have `python-lsp-server` installed, as it enables
+your text editor to perform semantic operations on the codebase (eg Go to
+Definition, Replace All, refactorings, etc)
+
 ### Containers
 
 To quickly get things up and running, you can also use Docker with the included
--- a/flake.nix
+++ b/flake.nix
@@ -101,7 +101,6 @@
                   extras = [
                     "develop"
                     "docs"
-                    "lsp"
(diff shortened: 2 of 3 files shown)


alexfikl:
I don't use the lsp or nix, so can't say if this would be useful. Will have to wait for someone to have an opinion :grin:

hseg:
My metric for this one is less whether it's useful and more that it shouldn't break the nix workflow - afaict, `nix develop` should still open a shell with `python-lsp-server` installed, it's just that I've removed the lsp extra from pip.

30 mar 2025 18:33:34 Alex Fikl ***@***.***>:

jghauser:
Looks good to me!

Issue #950, opened 2024-11-18
Error on bibtex export


supersambo:
*papis, version 0.14

Hi, 
I am running into an issue when exporting my complete library to a bibtex file.

# how to reproduce the issue:
Command and output of `papis --log INFO export -a -f bibtex -o export.bib`:
```
Traceback (most recent call last):
  File "/home/supersambo/.local/bin/papis", line 8, in <module>
    sys.exit(run())
             ^^^^^
  File "/home/supersambo/.local/pipx/venvs/papis/lib/python3.12/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/supersambo/.local/pipx/venvs/papis/lib/python3.12/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/supersambo/.local/pipx/venvs/papis/lib/python3.12/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/supersambo/.local/pipx/venvs/papis/lib/python3.12/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/supersambo/.local/pipx/venvs/papis/lib/python3.12/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/supersambo/.local/pipx/venvs/papis/lib/python3.12/site-packages/papis/commands/export.py", line 138, in cli
    ret_string = run(documents, to_format=fmt)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/supersambo/.local/pipx/venvs/papis/lib/python3.12/site-packages/papis/commands/export.py", line 87, in run
    .plugin(document for document in documents))
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/supersambo/.local/pipx/venvs/papis/lib/python3.12/site-packages/papis/bibtex.py", line 300, in exporter
[...]

PR #996, opened 2025-04-05
doctor missing values

pmiam:
Minor doctor enhancement intended to catch empty key values. I integrated it into the missing keys check and perhaps that is questionable placement, but it is conceptually simple and I've found it to work well for me the past 5 months.

It should address issues like #950. There, Alex suggested integrating this kind of check into the key-types doctor.

### DIFF
--- a/papis/commands/doctor.py
+++ b/papis/commands/doctor.py
@@ -39,6 +39,8 @@ implemented:
   :confval:`doctor-field-type-separator` setting.
 * ``keys-missing``: checks that the keys provided by
   :confval:`doctor-keys-missing-keys` exist in the document.
+* ``values-missing``: checks that fields in the document do not have empty values
+  (e.g. *None*, empty lists, empty strings, etc).
 * ``refs``: checks that the document has a valid reference (i.e. one that would
   be accepted by BibTeX and only contains valid characters).
 * ``string-cleaner``: checks that strings do contain various undesired characters
--- a/papis/doctor.py
+++ b/papis/doctor.py
@@ -193,6 +193,42 @@ def keys_missing_check(doc: Document) -> list[Error]:
             for k in keys if k not in doc]
 
 
+VALUES_MISSING_CHECK_NAME = "values-missing"
+
+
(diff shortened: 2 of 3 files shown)


papis/commands/doctor.py
```diff
@@ -299,14 +311,28 @@ def fixer_author_list_from_author() -> None:
...
+        elif not _is_nonempty(doc[key]):
+            errors.append(Error(name=KEYS_MISSING_CHECK_NAME,
+                                path=folder,
+                                msg=f"Value contains one or more empty entries:\n{
+                                papis.document.dump({f"{key}":doc[key]})
+                                }",
+                                suggestion_cmd=f"papis edit --doc-folder {folder}",
+                                fix_action=None,
+                                payload=key,
+                                doc=doc))
```
alexfikl, line 334: I'm not sure this conceptually fits in here very well, since this check is about "keys" missing, not empty values?

Maybe worth adding very simple copy-pasted `values-missing` check instead? I think I argued a bit for adding it to the `key-type` check, since an "empty value" might be an "incorrect value type", but it just complicates the logic.

papis/commands/doctor.py
```diff
@@ -299,14 +311,28 @@ def fixer_author_list_from_author() -> None:
...
+
+        elif not _is_nonempty(doc[key]):
+            errors.append(Error(name=KEYS_MISSING_CHECK_NAME,
+                                path=folder,
+                                msg=f"Value contains one or more empty entries:\n{
+                                papis.document.dump({f"{key}":doc[key]})
+                                }",
+                                suggestion_cmd=f"papis edit --doc-folder {folder}",
+                                fix_action=None,
```
alexfikl, line 332: Add a fixer to just remove the key in this case?

alexfikl (changes requested):
This looks like a good idea to me (think I had a branch with a similar check :grin:). Could you add some tests too?

alexfikl:
@pmiam Still interested in getting something like this in? :grin:

alexfikl:
Hm, forgot we had an `empty-fields` check now, haha. Closing this.

PR #997, opened 2025-04-06
doctor unexpected keys

pmiam:
I wanted a compliment to the required keys check that would help me eliminate pointless metadata from my entries destined for bibtex export.

With this check comes some improvements to the doctor's diagnoses. Namely clearer and more actionable descriptions facilitated by a no dependency TUI list printing function.

As part of the implementation the fixer function edits the user's config file so that may be objectionable.

### DIFF
--- a/papis/commands/doctor.py
+++ b/papis/commands/doctor.py
@@ -661,13 +661,35 @@ def biblatex_expected_keys_check(doc: papis.document.Document) -> List[Error]:
     if errors:
         return [e._replace(name=BIBLATEX_REQUIRED_KEYS_CHECK_NAME) for e in errors]
 
+    def make_fixer(key: str) -> FixFn:
+        def fixer() -> None:
+            config_file = papis.config.get_config_file()
+            config = papis.config.get_configuration()
+            lib = papis.config.get_lib_name()
+            logger.info("[FIX] Adding '%s' to %s.bibtex-ignore-keys", key, lib)
+
+            if os.path.exists(config_file):
+                logger.warning("file contents (whitespace, comments) may change")
+                if not papis.tui.utils.confirm("Proceed with overwrite?"):
+                    raise Exception("fix canceled by user")
+
+            ignores = papis.config.getlist("bibtex-ignore-keys")
+            ignores.append(key)
+            config[lib]["bibtex-ignore-keys"] = str(ignores)
+
+            with open(config_file, "w") as configfile:
+                config.write(configfile)
+            logger.info("Configuration file saved at '%s'.", config_file)
+
+        return fixer
+
     return [Error(name=BIBLATEX_EXPECTED_KEYS_CHECK_NAME,
                   path=folder,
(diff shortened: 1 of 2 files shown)


papis/commands/doctor.py
```diff
@@ -633,6 +638,64 @@ def biblatex_required_keys_check(doc: papis.document.Document) -> List[Error]:
             if not any(key in doc or aliases.get(key) in doc for key in keys)]
 
 
+BIBLATEX_EXPECTED_KEYS_IGNORED = {
+    "papis_id", "author_list", "editor_list", "citations", "doc_url",
+    "files", "ref", "time-added", "type", "tags"
+} | frozenset(papis.config.getlist("bibtex-ignore-keys"))
```
alexfikl, line 644: Maybe just add most of these directly to `bibtex.bibtex_ignore_keys`?

papis/commands/doctor.py
```diff
@@ -633,6 +638,64 @@ def biblatex_required_keys_check(doc: papis.document.Document) -> List[Error]:
...
+                if not papis.tui.utils.confirm("Proceed with overwrite?"):
+                    raise Exception("fix canceled by user")
+
+            ignores = papis.config.getlist("bibtex-ignore-keys")
+            ignores.append(key)
+            config[lib]["bibtex-ignore-keys"] = str(ignores)
+
+            with open(config_file, "w") as configfile:
+                config.write(configfile)
+            logger.info("Configuration file saved at '%s'.", config_file)
```
alexfikl, line 682: This seems very convoluted and unsafe for an automated fix. I would recommend just extending the error message with a suggestion e.g. `add '{key}' to 'bibtex-ignore-keys' in your configuration file to silence this error`?

alexfikl (commented):
I know this is still a work-in-progress draft, but I left a few comments about the bigger picture to talk about.

papis/commands/doctor.py
```diff
@@ -493,7 +494,9 @@ def fixer() -> None:
     if bib_type is None:
         return [Error(name=BIBTEX_TYPE_CHECK_NAME,
                       path=folder,
-                      msg="Document does not define a type",
+                      msg=f"Document does not define type as one of:\n{
+                      papis.tui.utils.string_grid(sorted(bibtex_types))
+                      }",
```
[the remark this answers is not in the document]
alexfikl: In general, I agree that we could do a lot better with the error reporting in `papis doctor`. Should probably try to copy `ruff` or some newer tools there (e.g line numbers, error codes, better suggestions), but that's quite a bit of work..

papis/commands/doctor.py
alexfikl: One thing we could do here, is extend the `--edit` functionality to allow opening the configuration file instead if the check requests it?

alexfikl:
@pmiam Still interested in getting something like this in? :grin:

alexfikl:
I'll go ahead and close this. It looks a bit too opinionated for a built-in check.

PR #998, opened 2025-04-06
fix: misc issues in merge and split_authors_name

pmiam:
very minor miscellaneous fixes I've been sitting on for months.

### DIFF
--- a/papis/commands/merge.py
+++ b/papis/commands/merge.py
@@ -78,15 +78,20 @@ def run(keep: Document,
         files: list[str],
         keep_both: bool,
         git: bool = False) -> None:
+    import shutil
 
     files_to_move = set(files) - set(keep.get_files())
-    for f in files_to_move:
-        to_folder = keep.get_main_folder()
-        if to_folder:
-            import shutil
-            logger.info("Moving '%s' to '%s'.", f, to_folder)
+    to_folder = keep.get_main_folder()
+
+    if to_folder:
+        keep_file_names = keep.get("files", [])
+        for f in files_to_move:
+            logger.info("Copying file '%s' to '%s'.", f, to_folder)
             shutil.copy(f, to_folder)
-            keep.get("files", []).append(os.path.basename(f))
+            keep_file_names.append(os.path.basename(f))
+
+        if keep_file_names:
+            keep["files"] = keep_file_names
 
     from papis.commands.update import run as update
     update(keep, data, git=git)
--- a/papis/commands/merge.py
+++ b/papis/commands/merge.py
@@ -86,7 +86,7 @@ def run(keep: Document,
             import shutil
             logger.info("Moving '%s' to '%s'.", f, to_folder)
             shutil.copy(f, to_folder)
-            keep["files"] += [os.path.basename(f)]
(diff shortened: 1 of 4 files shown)


papis/document.py
```diff
@@ -284,7 +284,7 @@ def split_authors_name(authors: Union[str, List[str]],
         sep = separator if separator else guess_authors_separator(subauthors)
         author_list.extend([
             split_author_name(author)
-            for author in re.split(fr"\s*{sep}\s+", subauthors)
+            for author in re.split(fr"\s+{sep}\s+", subauthors)
```
alexfikl, line 287: From what I recall, this was meant to support `Name1, Name2` or something like that, where before the separator "," there's no space. What's the reasoning for changing it?

papis/commands/merge.py
```diff
@@ -93,7 +93,7 @@ def run(keep: papis.document.Document,
             import shutil
             logger.info("Moving '%s' to '%s'.", f, to_folder)
             shutil.copy(f, to_folder)
-            keep["files"] += [os.path.basename(f)]
+            keep.get("files", []).extend([os.path.basename(f)])
```
alexfikl, line 96: I don't think this works as expected. Did you test it? You want
```
keep.setdefault("files", []).append(os.path.basename(f))
```

alexfikl (changes requested):
Thanks for the fixes! Left a few comments in there :grin:

papis/document.py
alexfikl: It's also breaking the tests from what I can tell (?).

pmiam: Yes it is breaking the tests. For some reason I'm getting a bunch of errors on main too, so I didn't notice, sorry. Also, I never use a comma as separator so it didn't come up for me. 

Nonetheless, with the \s* regex, if the separator is inside a name ("and" in "Roland") it is matched and it breaks the splitting. Probably a more clever expression is needed.

papis/commands/merge.py
pmiam: I have used it successfully. I notice the basename was being placed in a list originally, so I probably kept that out of an abundance of caution and used `extend` so the merged list would not contain sublist elements. Fixed.

papis/document.py
alexfikl: Yeah, might want to use different regexes when `sep.strip().isalpha()` or not? Not sure what a nice way to handle this is..

alexfikl: Something like enforcing a word boundary with `\b` should also work?

alexfikl:
@pmiam Still interested in getting some of these fixes in? :grin:

papis/document.py
alexfikl: Went ahead and implemented something like this. It seems to pass for an author list with "Roland" as an author and "and" as a separator.

papis/commands/merge.py
alexfikl: Went ahead and fixed this. It now supports no files and if the merge results in no files added anyway, it doesn't add a "files" key to the document.

alexfikl:
Went ahead and fixes the remaining issues here.

PR #1005, opened 2025-04-22
test: Add testpaths

hseg:
While investigating something else, finally found how to get rid of this minor irritant. This enables invoking pytest with a simple `python -m pytest`, instead of requiring enumerating the test directories as `python -m pytest papis tests`, which trips me up from time to time.


alexfikl:
Thanks! In it goes :rocket:

PR #1006, opened 2025-04-23
Enable mypy everywhere

hseg:
On a personal aesthetic whim (wanting the tool invocations to be bare `python -m tool`), I looked into making `mypy` and `flake8` autodetect all the Python code we wanted them to. At which point I discovered we were annotating most of the test suites, but not actually checking them. So I enabled `mypy` everywhere, and it works pretty well.

One knock-on difficulty is that `mypy` can't distinguish between files without extension, so I needed to rename the scripts in `examples/scripts/` to contain extensions.^[If this specific change needs to be reverted, know that trying to exclude the bash script in `examples/scripts` by including everything then excluding that one specifically didn't work. Also, you'd probably need `--scripts-are-modules`] I initially thought this meant `flake8` now also catches them, but that may have already been the case (and I'm too tired now to check).

Enabling across the board mostly worked:

- **Mask errors due to compat shims**
- **Add forgotten type hints**
- **Specify config field types on lookup** by replacing `get` by `get...()` where appropriate
- **Don't redefine variables**
- **Fix implicit reexport**

In a couple cases, I needed to guide `mypy` along -- if there are ways to narrow the types on the functions it sees, that'd be preferable (see below) but in many of these cases the easier initial step is just to assert that the situation is as expected.

- **Assert proper concrete class used**
- **Manually assert nullable inputs aren't**

This process caught two errors which I fixed:

- **tui: Clarify get_line_prefix arg name**

  `blih` is an undescriptive name, I had no idea what it meant. It turns out that parameter is the wrapping width for the line, and is a non-`None` `int`. I set it to `0`, which IIUC should be OK.
- **doctor: Make fix_action non-nullable**

  There were instances where we were calling `fix_action` without being able to be sure it was non-`None`. Since PEP 505 wasn't accepted, `e?.fix_action()` isn't an option, and so I elected to add a field `can_fix` to `Error` to take the place of `fix_action is not None` and set `fix_action` to a no-op instead.

Two points I've masked and left for a future ticket, since they require more serious engineering to fix:
- Converting `bibtex` to a `dict` gives you a heterogeneous `dict`, which is bad for type inference. It'd be preferable to convert it to a `NamedTuple`, but this has knock-on effects
- `test_duplicate_new_names` reads the output log to try and get the filenames -- but it is unclear why it should be able to expect the types of the arguments it sees are all filenames. If there's a better way of obtaining this information, we might want to use it instead


hseg:
BTW, in the help I received on this, people noted we may want to use the builtin container types rather than those from `typing` once we drop `3.8` support, see https://stackoverflow.com/questions/39458193/using-list-tuple-etc-from-typing-vs-directly-referring-type-as-list-tuple-etc

papis/tui/widgets/list.py
```diff
@@ -113,7 +113,7 @@ def __init__(
     def get_line_prefix(
             self,
             line: int,
-            blih: int) -> AnyFormattedText:
+            wrap_count: int = 0) -> AnyFormattedText:
```
alexfikl, line 116: Why this change? The `GetLinePrefixCallable` definition does not seem to have a default value:
https://github.com/prompt-toolkit/python-prompt-toolkit/blob/d997aab538e434a6ca07d6bee226fd5b0628262f/src/prompt_toolkit/layout/controls.py#L60

tests/commands/test_config.py
```diff
@@ -9,9 +9,12 @@
...
 
-    assert run(["editor"]) == {"editor": papis.config.get("editor")}
-    assert run(["settings.editor"]) == {"editor": papis.config.get("editor")}
-    assert run(["papers.dir"]) == {"dir": papis.config.get("dir", section="papers")}
+    assert run(["editor"]) == {
+        "editor": papis.config.getstring("editor")}
+    assert run(["settings.editor"]) == {
+        "editor": papis.config.getstring("editor")}
+    assert run(["papers.dir"]) == {
+        "dir": papis.config.getstring("dir", section="papers")}
```
alexfikl, line 17: Why these changes? Are they necessary?

tests/commands/test_rename.py
```diff
@@ -124,6 +134,9 @@ def test_duplicate_new_names(tmp_library: TemporaryLibrary,
         # First document will rename, but all others should raise a warning
         assert len(caplog.records) == doc_count - 1
         assert all(
-            os.path.basename(record.args[0]) == "same-name"
+            # FIXME: Why do we know that args will be a tuple whose first
+            # component is a string? Is there a better way of obtaining this
+            # information?
+            os.path.basename(record.args[0]) == "same-name"       # type: ignore
```
alexfikl, line 140: The record here is just a `LogRecord` which has documented arguments:
https://docs.python.org/3/library/logging.html#logrecord-attributes
You might just need to check that `args` is non-empty instead?

tests/commands/test_rm.py
```diff
@@ -39,6 +41,7 @@ def test_rm_files_run(tmp_library: TemporaryLibrary) -> None:
     run(doc, filepath=filename, git=True)
     assert not os.path.exists(filename)
 
+    assert isinstance(db, papis.database.cache.Database)
```
alexfikl, line 44: Is this needed because it sets `db.documents = None`? This should be already done in `Database.clear`, so you can remove that line.

tests/database/test_papis.py
```diff
@@ -31,6 +31,8 @@ def test_database_reload(tmp_library: TemporaryLibrary) -> None:
     assert isinstance(db, papis.database.Database)
 
     ndocs = len(db.get_all_documents())
+
+    assert isinstance(db, papis.database.cache.Database)
```
alexfikl, line 35: Same here, just remove `db.documents = None`.

tests/database/test_papis.py
```diff
@@ -49,6 +51,7 @@ def test_database_missing(tmp_library: TemporaryLibrary) -> None:
     doc = docs[0]
     db.delete(doc)
 
+    assert isinstance(db, papis.database.cache.Database)
```
alexfikl, line 54: Same here?

tests/downloaders/test_usenix.py
```diff
@@ -43,10 +44,11 @@ def test_usenix_fetch(tmp_config: TemporaryConfiguration,
                       monkeypatch: MonkeyPatch,
                       url: str) -> None:
     cls = papis.downloaders.get_downloader_by_name("usenix")
-    assert cls is Downloader
+    assert cls is papis.downloaders.usenix.Downloader
```
alexfikl, line 47: Why is this needed? `Downloader` should already be imported from `usenix`, no?

tests/test_arxiv.py
```diff
@@ -44,13 +44,13 @@ def test_downloader_match(tmp_config: TemporaryConfiguration) -> None:
     from papis.arxiv import Downloader, ARXIV_ABS_URL
 
     down = Downloader.match("arxiv.org/sdf")
-    assert isinstance(down, Downloader)
```
alexfikl, line 47: Why is this not needed anymore? It should assert here.

alexfikl (changes requested):
Gave this a first look and left a bunch of comments. Some bigger things:
* Definitely ok with renaming the example scripts to contain the extension. I don't think that breaks anything (?).
* You seem to have done a big search-and-replace from `get` to `getstring`. Are those really necessary? It seems like many places are not.
* The `fix_action` change is not improving anything as far as I can tell, so I'm not in using that.

Everything else looks nice and definitely a good idea to keep the examples / tests clean. :grin:

EDIT: For the spelling error, feel free to just add that to the ignore list in `pyproject.toml`.

hseg:
> * The *fix_action* change is not improving anything as far as I can tell, so I'm not in using that.

My solution may have been overengineered - the alternative is to guard the call to fix_action behind a test that it isn't None. Can do that the same way I did the rest of the nullability warnings - I just thought this pattern was safer.

> * You seem to have done a big search-and-replace from *get* to *getstring*. Are those really necessary? It seems like many places are not.

They might not be necessary in some places, but in others they're needed to hint to mypy that the resulting variable is a string. The way I found these is by searching for plain gets and then looking up in the configuration docs to see if the field was a string. Except for one place whose default was None, another which explicitly was a FormatString (and which I only caught due to the test failing - that was confusing for a moment), and those keys we're renaming (and thus have logic to check for the typo), I changed all of them.
Can go through them again with a finer-toothed comb, but I'll want to accompany that with a documentation change to the configuration docs to note which keys are nullable and which keys are actually FormatString.
That the tests passed with my incorrectly narrow types is concerning - we might need to fuzz-check the tests if they're not exercising valid configuration space.

> ----------------------------------------
>
> In examples/scripts/papis-tags.py[https://github.com/papis/papis/pull/1006#discussion_r2057657245]:
>
>> @@ -57,7 +57,7 @@ def main(query: str, confirm: bool) -> None:
>      )
> 
>      # Create an empty tag list
> -    tag_list = set()
> +    tag_list: set[str] = set()
>
> I'm not sure why this passes, but I don't think it should. This is only implemented in >=3.9 in PEP 585[https://peps.python.org/pep-0585/].
>
> Is mypy checking this correctly? We might need to add something like this to *pyproject.toml*:
>
> [mypy]
> python_version = "3.8"

See for yourself - the CI for 3.8 passes this at least. I'll double-check later that it is indeed testing that.

> ----------------------------------------
>
> In papis/arxiv.py[https://github.com/papis/papis/pull/1006#discussion_r2057660742]:
>
>> @@ -58,7 +58,7 @@
>      _k("entry_id", [{"key": "url", "action": None}]),
>      _k("journal_ref", [{"key": "journal", "action": None}]),
>      _k("pdf_url", [{
> -        "key": str(papis.config.get("doc-url-key-name")),
> +        "key": str(papis.config.getstring("doc-url-key-name")),
>
> Why is this key no longer allowed to be *None*? I'm also not sure why there's a *str* wrapping this to begin with :\

This is good evidence of my search-and-replace. I'll document that it is nullable. Does this mean that with "doc-url-key-name = None" this is supposed to evaluate to "None"?

> ----------------------------------------
>
> In papis/bibtex.py[https://github.com/papis/papis/pull/1006#discussion_r2057666172]:
>
>> +# FIXME: The need to have author_list in the bibtex dict makes it have
> +# heterogeneous values -- and trying to be more specific (it's str |
> +# List[Dict[str, Any]]) creates cascading type warnings.
> +# A better solution is to instead create a NamedTuple of some sort that will
> +# hold this data
> +def bibtex_to_dict(bibtex: str) -> List[Dict[str, Any]]:
> ⬇️ Suggested change
> -# FIXME: The need to have author_list in the bibtex dict makes it have
> -# heterogeneous values -- and trying to be more specific (it's str |
> -# List[Dict[str, Any]]) creates cascading type warnings.
> -# A better solution is to instead create a NamedTuple of some sort that will
> -# hold this data
> -def bibtex_to_dict(bibtex: str) -> List[Dict[str, Any]]:
> +def bibtex_to_dict(bibtex: str) -> List[papis.document.DocumentLike]:
>
> ?
>
> The initial return type is indeed incorrect, but I don't think we can be too explicit about what the value types are unless we want to make a long list (ints are also ok for years / months / volumes / etc). *DocumentLike* seems fair because whatever this returns will likely be turned into a document soon enough anyway.
>
> About your suggestion to convert it to a *NamedTuple* or something of that type: This is generally a short lived result that gets added some importer data that is also heterogeneous, so I'm not sure that's useful. The underlying problem is that the *Document* itself is not a fixed data structure and can have arbitrary user fields. Unless we change that, this is just patchwork.. :(

Ah, ok, I'm unfamiliar with the codebase, so was unaware of that option. Will look it up.

> ----------------------------------------
>
> In papis/commands/bibtex.py[https://github.com/papis/papis/pull/1006#discussion_r2057667213]:
>
>> @@ -162,7 +162,7 @@ def cli(ctx: click.Context, no_auto_read: bool) -> None:
>      else:
>          no_auto_read = not papis.config.getboolean("auto-read", section="bibtex")
> 
> -    bibfile = papis.config.get("default-read-bibfile", section="bibtex")
> +    bibfile = papis.config.getstring("default-read-bibfile", section="bibtex")
>
> Why is this no longer allowed to be None? The if below should check for that just fine. I'm mostly worried because *getstring* will raise if the value is not a string and I don't know if all uses take that into account.

Will look.

> ----------------------------------------
>
> In papis/commands/doctor.py[https://github.com/papis/papis/pull/1006#discussion_r2057671557]:
>
>> -    fix_action: Optional[FixFn]
> +    fix_action: FixFn
> +    #: Whether fix_action can autofix the error
> +    can_fix: bool
>
> I don't like this at all and don't see how it's better than just checking if *fix_action is None*. Now you just have two variables that need to be kept in sync..
>
> Can you please revert it? From what I can tell, this is mostly an issue in the tests so it needs to be fixed there.

True, this was the first mypy error I tackled in the series, so before I landed on checking for None for the others. Didn't think to go back and revert this.

> ----------------------------------------
>
> In papis/database/__init__.py[https://github.com/papis/papis/pull/1006#discussion_r2057680983]:
>
>> @@ -1,6 +1,6 @@
>  from typing import Optional, Dict
> 
> -from .base import Database
> +from .base import Database as Database
>
> What's this for?

Mypy doesn't like implicit reexports, this is a way of pacifying it. Should've documented it.

> ----------------------------------------
>
> In papis/tui/widgets/list.py[https://github.com/papis/papis/pull/1006#discussion_r2057692432]:
>
>> -            blih: int) -> AnyFormattedText:
> +            wrap_count: int = 0) -> AnyFormattedText:
>
> Why this change? The *GetLinePrefixCallable* definition does not seem to have a default value:
> https://github.com/prompt-toolkit/python-prompt-toolkit/blob/d997aab538e434a6ca07d6bee226fd5b0628262f/src/prompt_toolkit/layout/controls.py#L60

Looking there, I entirely misread the meaning of that parameter - it's so the callers can tell get_line_prefix the width of the textbox it's rendering in. The default value here was mostly to remind myself what a sane parameter to pass here is.

> ----------------------------------------
>
> In tests/commands/test_config.py[https://github.com/papis/papis/pull/1006#discussion_r2057698117]:
>
>> -    assert run(["editor"]) == {"editor": papis.config.get("editor")}
> -    assert run(["settings.editor"]) == {"editor": papis.config.get("editor")}
> -    assert run(["papers.dir"]) == {"dir": papis.config.get("dir", section="papers")}
> +    assert run(["editor"]) == {
> +        "editor": papis.config.getstring("editor")}
> +    assert run(["settings.editor"]) == {
> +        "editor": papis.config.getstring("editor")}
> +    assert run(["papers.dir"]) == {
> +        "dir": papis.config.getstring("dir", section="papers")}
>
> Why these changes? Are they necessary?

The whitespace changes are because flake8 complained of long lines.

> ----------------------------------------
>
> In tests/commands/test_rename.py[https://github.com/papis/papis/pull/1006#discussion_r2057702436]:
>
>> -            os.path.basename(record.args[0]) == "same-name"
> +            # FIXME: Why do we know that args will be a tuple whose first
> +            # component is a string? Is there a better way of obtaining this
> +            # information?
> +            os.path.basename(record.args[0]) == "same-name"       # type: ignore
>
> The record here is just a *LogRecord* which has documented arguments:
> https://docs.python.org/3/library/logging.html#logrecord-attributes
> You might just need to check that *args* is non-empty instead?

Hm, and they're all guaranteed to be castable to str. I'll see if asserting length(record.args)>0 and that record.arts[0] is not None works.

> ----------------------------------------
>
> In tests/commands/test_rm.py[https://github.com/papis/papis/pull/1006#discussion_r2057706017]:
>
>> @@ -39,6 +41,7 @@ def test_rm_files_run(tmp_library: TemporaryLibrary) -> None:
>      run(doc, filepath=filename, git=True)
>      assert not os.path.exists(filename)
> 
> +    assert isinstance(db, papis.database.cache.Database)
>
> Is this needed because it sets *db.documents = None*? This should be already done in *Database.clear*, so you can remove that line.

Yes

> ----------------------------------------
>
> In tests/database/test_papis.py[https://github.com/papis/papis/pull/1006#discussion_r2057706901]:
>
>> @@ -31,6 +31,8 @@ def test_database_reload(tmp_library: TemporaryLibrary) -> None:
>      assert isinstance(db, papis.database.Database)
> 
>      ndocs = len(db.get_all_documents())
> +
> +    assert isinstance(db, papis.database.cache.Database)
>
> Same here, just remove *db.documents = None*.

OK

> ----------------------------------------
>
> In tests/database/test_papis.py[https://github.com/papis/papis/pull/1006#discussion_r2057707204]:
>
>> @@ -49,6 +51,7 @@ def test_database_missing(tmp_library: TemporaryLibrary) -> None:
>      doc = docs[0]
>      db.delete(doc)
> 
> +    assert isinstance(db, papis.database.cache.Database)
>
> Same here?

OK

> ----------------------------------------
>
> In tests/downloaders/test_usenix.py[https://github.com/papis/papis/pull/1006#discussion_r2057708679]:
>
>> @@ -43,10 +44,11 @@ def test_usenix_fetch(tmp_config: TemporaryConfiguration,
>                        monkeypatch: MonkeyPatch,
>                        url: str) -> None:
>      cls = papis.downloaders.get_downloader_by_name("usenix")
> -    assert cls is Downloader
> +    assert cls is papis.downloaders.usenix.Downloader
>
> Why is this needed? *Downloader* should already be imported from *usenix*, no?

This change may be unnecessary, but asserting down is a p.d.usenix.Downloader is. For some reason, mypy won't accept the hint when phrased as
`assert isinstance(down, cls)`

> ----------------------------------------
>
> In tests/test_arxiv.py[https://github.com/papis/papis/pull/1006#discussion_r2057709507]:
>
>> @@ -44,13 +44,13 @@ def test_downloader_match(tmp_config: TemporaryConfiguration) -> None:
>      from papis.arxiv import Downloader, ARXIV_ABS_URL
> 
>      down = Downloader.match("arxiv.org/sdf")
> -    assert isinstance(down, Downloader)
>
> Why is this not needed anymore? It should assert here.
>

Misread the code flow as a multi-step process to find down, instead of testing its value against multiple Downloader.match calls.

alexfikl:
> My solution may have been overengineered - the alternative is to guard the call to fix_action behind a test that it isn't None. Can do that the same way I did the rest of the nullability warnings - I just thought this pattern was safer.

My understanding was that this workaround was introduced to make `mypy` pass on `test_doctor.py` without any more changes there. Is that not right? I'm not sure in what sense it could be safer than just doing `if fix_action is not None: fix_action()`..

> That the tests passed with my incorrectly narrow types is concerning - we might need to fuzz-check the tests if they're not exercising valid configuration space.

The tests are most definitely not exercising the large part of the configuration space. I'm not sure when we'll get to something like that :(

> See for yourself - the CI for 3.8 passes this at least. I'll double-check later that it is indeed testing that.

Yeah, I noticed they're passing, but that's weird. I'll check too.

> This is good evidence of my search-and-replace. I'll document that it is nullable. Does this mean that with "doc-url-key-name = None" this is supposed to evaluate to "None"?

That stuff is passed to `keyconversion_to_data` which ignores any keys or values that are *None*. Effectively it will ignore it, so it's fine if `doc-url-key-name` is *None*.

> Mypy doesn't like implicit reexports, this is a way of pacifying it. Should've documented it.

Hm, where was it complaining about that? All the type checking on the main codebase was passing just fine, so I'm guessing it's a problem in the tests again?

> Looking there, I entirely misread the meaning of that parameter - it's so the callers can tell get_line_prefix the width of the textbox it's rendering in. The default value here was mostly to remind myself what a sane parameter to pass here is.

I agree with the name change. It seems to match what `prompt_toolkit` is using in most places as well. I'm not against the default value, but every usage place passes both parameters anyway, so it's not needed.

> The whitespace changes are because flake8 complained of long lines.

Yes, I was talking about the `get` to `getstring` changes that made that necessary. From your changes it seems like a lot of places used `get` as a `getstring` that doesn't raise, which seems fine too.

> Hm, and they're all guaranteed to be castable to str. I'll see if asserting length(record.args)>0 and that record.arts[0] is not None works.

Yeah, something like that should be just fine!

> This change may be unnecessary, but asserting down is a p.d.usenix.Downloader is. For some reason, mypy won't accept the hint when phrased as `assert isinstance(down, cls)`

Hm, there's something funky in there (might just be mypy itself..). I'll take a closer look when some of the other changes are in.

alexfikl:
@hseg Any chance I can convince you to respond on each Github comment/conversation? The formatting from email seems to be quite messed up and it's hard to keep track which conversations are resolved and which aren't :(

hseg:
Sure, I meant to go back and post there once I get back to keyboard - thought I'd respond fast first.

24 abr 2025 14:02:28 Alex Fikl ***@***.***>:

alexfikl:
> Sure, I meant to go back and post there once I get back to keyboard - thought I'd respond fast first.

No worries!

papis/arxiv.py
```diff
@@ -58,7 +58,7 @@
     _k("entry_id", [{"key": "url", "action": None}]),
     _k("journal_ref", [{"key": "journal", "action": None}]),
     _k("pdf_url", [{
-        "key": str(papis.config.get("doc-url-key-name")),
+        "key": str(papis.config.getstring("doc-url-key-name")),
```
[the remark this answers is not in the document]
hseg: This is good evidence of my search-and-replace. I'll document that it is nullable. Does this mean that with "doc-url-key-name = None" this is supposed to evaluate to "None"?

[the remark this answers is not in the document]
hseg: Actually, why is it that the other `getTYPE` functions return an `Optional[TYPE]`, whereas `getstring` (and TIL, also `getformattedstring` exists) return a non-Optional type?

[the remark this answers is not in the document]
alexfikl: No, I think that `str` there is either a bug or whoever wrote this (I'm not gonna git blame! :grin:) forgot that it can be `None`. Maybe you can just remove the `str` wrapper here? I think it's fine to leave it as `getstring`.

papis/database/__init__.py
```diff
@@ -1,6 +1,6 @@
 from typing import Optional, Dict
 
-from .base import Database
+from .base import Database as Database
```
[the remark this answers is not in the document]
hseg: `mypy` warns everywhere that references `papis.database.Database` by only importing `papis.database` that `Database` isn't explicitly exported -- yes, it is in scope, but it isn't explicit. Adding a binding for it in the module does that. Should've documented it.

papis/tui/widgets/list.py
hseg: Looking there, I entirely misread the meaning of that parameter - it's so the callers can tell get_line_prefix the width of the textbox it's rendering in. The default value here was mostly to remind myself what a sane parameter to pass here is. Correcting both.

papis/bibtex.py
```diff
@@ -406,7 +406,12 @@ def bibtexparser_entry_to_papis(entry: Dict[str, Any]) -> Dict[str, Any]:
     return result
 
 
-def bibtex_to_dict(bibtex: str) -> List[Dict[str, str]]:
+# FIXME: The need to have author_list in the bibtex dict makes it have
+# heterogeneous values -- and trying to be more specific (it's str |
+# List[Dict[str, Any]]) creates cascading type warnings.
+# A better solution is to instead create a NamedTuple of some sort that will
+# hold this data
+def bibtex_to_dict(bibtex: str) -> List[Dict[str, Any]]:
```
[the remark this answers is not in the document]
hseg: (Although, this kind of half-cooked dict juggling is precisely the kind of place where I'm very nervous of the types involved, so if we have ways of making the manipulations here more principled that'd be great)

papis/database/__init__.py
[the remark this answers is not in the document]
hseg: And yes, the cases are all in `tests/database/test_papis.py`

tests/commands/test_rename.py
hseg: Hm, and they're all guaranteed to be castable to `str`. I'll see if asserting `length(record.args)>0` and that `record.args[0] is not None` works.

hseg: ... Nope -- `record.args` has type `tuple[object, ...] | Mapping[str, object] | None`

hseg: so it's not even guaranteed to have positional arguments

hseg: OK, just shoved type assertions in there to keep `mypy` happy.

tests/database/test_papis.py
hseg: Still has the problem of `db._save_documents()` not being generic enough.

tests/commands/test_rm.py
hseg: That works.

tests/downloaders/test_usenix.py
hseg: Had some trouble spelling out the check -- in some cases, mypy was still inferring the base class. But indeed, just replacing it with `Downloader` works.

tests/test_arxiv.py
hseg: Misread the code flow as a multi-step process to find down, instead of testing its value against multiple `Downloader.match` calls.

hseg:
OK, so went through most of the reviews, leaving the following for later:

- [ ] `doc-url-key-name`, `default-read-bibfile` are according to your remarks apparently of type `Optional[str]` -- I'd like to document their correct types (if there are any other optional values with string defaults, I'd document them as well) (I'm still unconvinced they should be `Optional` -- their usage suggests they're always meant to be valid strings. Yes, the former is passed to `keyconversion_to_data`, but that doesn't mean that we need to support complex types in the config schema.
- [ ] I'd like to push back on the presumption that "a non-raising `getstring`" is a good idea -- config data is user-input data, hence unsafe, hence should be validated once, following which only the validated configuration should be used. The lack of clarity regarding the types of the configuration values is symptomatic here.
- [ ] Similarly, it's odd that `getstring` and `getformattedstring` raise, whereas `getint`, `getfloat` and `getboolean` return `None` instead. Why the inconsistency? Do we want to keep it?
- [ ] The two cases where I inserted `isinstance(db, papis.database.cache.Database)` are necessary since the code around there makes use of functionality only that class has (`_save_documents()`, `_locate_document()`). If we want to keep those tests generic I need to know how to get rid of those lines.
- [ ] The little mystery around `mypy` accepting `: set[str]` is weird, but as long as it passes CI, do we care?

The following config keys still remain with untyped `get`:
- `sort-field` (`Optional[str]`)
- `key-type-separator` (`Optional[str]`)
- `local-config-file` (`str`?)
- `header-format-file` (`Optional[str]`)
- `default-library` (`str`?)
- `editmode` (`str`?)
- `downloader-proxy` (`Optional[str]`)
- `cache-dir` (`Optional[str]`)
- `serve-default-tag-sorting` (`str`?)
- `editor` (you pushed back?) (`str`?)
- `dir` (idem) (`str`?)
- `add_file_name` (`FormattedString`?)
- bunch of instances in `test_config.py` due to exercising `TemporaryConfiguration`, reverting those I did put in until we decide to exercise all of them)

(Not counting the deprecations of `keys-exist-keys`, `key-type-check-keys`, `key-type-check-separator`, `formater`)

hseg:
Huh, 3.8 has started failing on the `set[str]` thing. Mystery solved.

tests/database/test_papis.py
alexfikl: Ah, wait, I didn't notice the file name.. this is meant to only test the `papis` database backend! Then yeah, this is perfectly fine.. sorry for the noise.

tests/database/test_papis.py
alexfikl: Yeah, didn't notice the file name. This is fine!

alexfikl:
> OK, so went through most of the reviews, leaving the following for later:
> 
>     * [ ]  `doc-url-key-name`, `default-read-bibfile` are according to your remarks apparently of type `Optional[str]` -- I'd like to document their correct types (if there are any other optional values with string defaults, I'd document them as well) (I'm still unconvinced they should be `Optional` -- their usage suggests they're always meant to be valid strings. Yes, the former is passed to `keyconversion_to_data`, but that doesn't mean that we need to support complex types in the config schema.

We don't really have a "config schema" and it seems like a lot of work to introduce one.

That being said, the alternative to using None for some of these things is using an empty string, which doesn't seem much clearer to me. At least `None` tells you it's "not set". Like you noticed, I doubt all the config options are consistent in this way, so it might need someone to take a serious look at it :-s

>     * [ ]  I'd like to push back on the presumption that "a non-raising `getstring`" is a good idea -- config data is user-input data, hence unsafe, hence should be validated once, following which only the validated configuration should be used. The lack of clarity regarding the types of the configuration values is symptomatic here.

Agreed, but `getstring` doesn't really validate anything besides "this is indeed a string", which it always is when read from the configuration file. The only way it can be `None` is:
* It picked up the value from `papis/defaults.py`
* It got set in `config.py` or something else dynamic like that.
 
>     * [ ]  Similarly, it's odd that `getstring` and `getformattedstring` raise, whereas `getint`, `getfloat` and `getboolean` return `None` instead. Why the inconsistency? Do we want to keep it?

The three are like that because they correspond to `ConfigParser.getXXX` with a fallback to `None`. I would prefer it if they all just made the fallback a keyword argument so that it can be user controller and match `dict.get`.

The configuration helpers really need some reworking :(

> Huh, 3.8 has started failing on the set[str] thing. Mystery solved.

That's reassuring :heart:

papis/database/__init__.py
```diff
@@ -1,6 +1,6 @@
 from typing import Optional, Dict
 
-from .base import Database
+from .base import Database as Database  # Explicit reexport to maky mypy happy
```
alexfikl, line 3: Can you remove this and change the places in `test_papis.py` to use `papis.database.cache.Database` instead? 

This should probably be under a `TYPE_CHECKING` guard or something like that since it's not meant to be exported.

alexfikl (approved):
Just left one extra little nitpick, but otherwise this looks good to me!

Thanks for fixing it! Making type annotations work nicely is always a nice of a pain :(

papis/database/__init__.py
hseg: OK, done. Not sure what you mean by a `TYPE_CHECKING` guard?

alexfikl: Ah, I just meant that this import is only used for type annotations, so it could be imported under a [TYPE_CHECKING guard](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) like here:
https://github.com/papis/papis/blob/241eef6371f3a37bbaef0eb327873e8b765f5168/papis/arxiv.py#L37-L38

It's fine to leave it as is though! Just thinking out loud :grin:

hseg:
> > OK, so went through most of the reviews, leaving the following for later:
> > ```
> > * [ ]  `doc-url-key-name`, `default-read-bibfile` are according to your remarks apparently of type `Optional[str]` -- I'd like to document their correct types (if there are any other optional values with string defaults, I'd document them as well) (I'm still unconvinced they should be `Optional` -- their usage suggests they're always meant to be valid strings. Yes, the former is passed to `keyconversion_to_data`, but that doesn't mean that we need to support complex types in the config schema.
> > ```
> 
> We don't really have a "config schema" and it seems like a lot of work to introduce one.
> 
> That being said, the alternative to using None for some of these things is using an empty string, which doesn't seem much clearer to me. At least `None` tells you it's "not set". Like you noticed, I doubt all the config options are consistent in this way, so it might need someone to take a serious look at it :-s

I'm not firing on all cylinders today, so I might be expressing myself less clearly than usual. I'm also probably importing my FP biases on how code "should" be structured here.

I associate `ConfigParser` to the external-facing part of the code, and so expect it eventually to produce some kind of `Config` object which only knows the fields `papis` needs -- `ConfigParser` would deal with the "merge multiple configs" and defaulting logic. In particular, the definition of `Config` would contain an enumeration of all the various fields and their types, and could itself be used to generate the Sphinx directives and look up config values by name not key (eg why are we writing `config.getstring("opentool") instead of `config.opentool`?).

Now, this may be a lot of churn, so instead we could use something like Cerberus to encapsulate the validation + normalization. But in any case, I'd submit we need a registry *somewhere* of all the fields used and their types.

> > * [ ]  I'd like to push back on the presumption that "a non-raising `getstring`" is a good idea -- config data is user-input data, hence unsafe, hence should be validated once, following which only the validated configuration should be used. The lack of clarity regarding the types of the configuration values is symptomatic here.
> 
> Agreed, but `getstring` doesn't really validate anything besides "this is indeed a string", which it always is when read from the configuration file. The only way it can be `None` is:
> 
> * It picked up the value from `papis/defaults.py`
> * It got set in `config.py` or something else dynamic like that.

Mm, no? IIUC, if the field is absent from the config file, `getstring` should also return `None`, right?

Also, if it's possible to set an inconsistent state in `config.py`, that's not a great sign.

> > * [ ]  Similarly, it's odd that `getstring` and `getformattedstring` raise, whereas `getint`, `getfloat` and `getboolean` return `None` instead. Why the inconsistency? Do we want to keep it?
> > ```
> 
> The three are like that because they correspond to `ConfigParser.getXXX` with a fallback to `None`. I would prefer it if they all just made the fallback a keyword argument so that it can be user controller and match `dict.get`.
> 
> The configuration helpers really need some reworking :(

Wait, looking at `ConfigParser`, I see its `getXXX` functions support a fallback. Unless you mean ours need to sprout a fallback parameter? In that case I'd probably agree.

alexfikl:
> I associate `ConfigParser` to the external-facing part of the code, and so expect it eventually to produce some kind of `Config` object which only knows the fields `papis` needs -- `ConfigParser` would deal with the "merge multiple configs" and defaulting logic. In particular, the definition of `Config` would contain an enumeration of all the various fields and their types, and could itself be used to generate the Sphinx directives and look up config values by name not key (eg why are we writing `config.getstring("opentool") instead of `config.opentool`?).

I understand the idea, but I'm not sure this would work with how dynamic Papis actually is. For example, you can insert new allowable configuration options from any plugin and they need to work and be accessible with any of the following
```
[settings]
myplugin-option = something

[myplugin]
option = something

[mylib]
myplugin-option = something
```

`config.getstring("option", section="myplugin")` will look through all of these and return the first one it finds. I'm not sure how `config.option` would work unless it'll just be a wrapper around the same logic. Alternatively, we could also remove a lot of this fancy flexibility..

There's probably some way to make this work given how dynamic Python is too, but at the moment it's all a dumb dict :grin:

> Mm, no? IIUC, if the field is absent from the config file, `getstring` should also return `None`, right?

If the field is not set in the configuration file, we fall back to the values in `papis/defaults.py`, so it doesn't necessarily return `None` in that case. If it's absent both in the configuration file and in all the defaults, we raise a `DefaultSettingValueMissing`.
 
> Also, if it's possible to set an inconsistent state in `config.py`, that's not a great sign.

Yeah, `config.py` is basically a "do anything" escape hatch. It just gets `eval`-ed into the main process and it can monkey-patch and change whatever it wants :grin: The only thing we can do there is go "pretty please just use the documented API calls so nothing breaks horribly".

As far as I know, that's actually used by quite a few people, so it's not going anywhere.

hseg:
Yikes. That's a shocking amount of flexibility, do we even need that much? 😰
But in fact, this precisely is why I don't think thinking in terms of the concrete INI structure is helpful - just like we don't bake into our thinking the presence of multiple config files.
So instead, I suggest - given everything can be overridden at the per-library level and the list of liberaries and plugins is fixed, perhaps think of config parsing as generating the following toml
```toml
[lib1]
# papis core settings
[lib1.plugin1]
# plugin1 registered settings
…
[lib2]
…
```
Which means we still can validate - we have a dict of configurations for each library, with the papis core schema controlled by us and the plugin schemas given when they register.

BTW, doesn't this mean a plugin could eg register itself with the name "local-config" and mess with "local-config-file"? Does this have security implications? At least, how are we preventing plugins from stepping on each other's toes?

24 abr 2025 22:05:03 Alex Fikl ***@***.***>:

alexfikl:
> Yikes. That's a shocking amount of flexibility, do we even need that much? 😰 

Probably not, but I'm generally hesitant to break too much of that, in case it's actually used :(

> BTW, doesn't this mean a plugin could eg register itself with the name "local-config" and mess with "local-config-file"? Does this have security implications? At least, how are we preventing plugins from stepping on each other's toes?

I'm mostly hoping nobody is using this anywhere where security is a consideration. 

I don't think there's currently any checking for common prefixes to some options, so they can be mistakenly overwritten by plugins. Plugins can also overwrite existing defaults using `register_default_settings`:
https://github.com/papis/papis/blob/241eef6371f3a37bbaef0eb327873e8b765f5168/papis/config.py#L194-L201

papis/database/__init__.py
alexfikl: @hseg This doesn't seem to be updated yet. Did you forget to push it?

hseg:
As you said, forgot the push.

I'm still interested in improving the documentation around configuration types, but it's more involved than I have time for in the next two weeks. Dropping my last two commits from the PR, will redo them in a future PR focusing on improving config types.

alexfikl:
Everything looks good now! Thanks for the work and patience! 

In it goes :rocket:

PR #1008, opened 2025-04-27
Add more handling for jats tags

alexfikl:
This also cleans up abstracts that start with `Synopsis:` or `Summary:`. For an example: http://dx.doi.org/10.1002/fld.3923.

### DIFF
--- a/papis/commands/doctor.py
+++ b/papis/commands/doctor.py
@@ -848,7 +848,7 @@ def html_tags_check(doc: papis.document.Document) -> List[Error]:
                     tag.replace_with(f" {tag.text.strip()} ")
 
             for tag in soup.find_all("jats:title"):
-                if tag.text.strip() == "Abstract":
+                if tag.text.strip().lower() in {"abstract", "summary", "synopsis"}:
                     tag.extract()
                 else:
                     # NOTE: these will get cleaned up in `make_fixer`


tuurep:
Hi, I've got several documents with tags like `<jats:p>` in the abstracts, didn't find help from this PR though, is this relevant here?

alexfikl:
> Hi, I've got several documents with tags like `<jats:p>` in the abstracts, didn't find help from this PR though, is this relevant here?

@tuurep This is a small addition to the `html-tags` check in `papis doctor`. I mostly use it to remove any `<jats:XXX>` and whatever other tags that look like HTML or XML or something. A command like this usually does the trick:
```bash
papis doctor --fix --explain --checks html-tags --all '.'
```
The `--fix` flag will actually autofix things for you, so you might want to remove it on first run.

I think by default the check looks in the abstract and title for tags like that. Is it not fixing things for you or what is the issue?

tuurep:
Thanks, was a dumb question on my part but I went for it anyway.. :smile: I wasn't at all familiar with how to use `papis doctor` and figured if I'd run `papis doctor <doc-with-jats-tag>`, maybe it'd work.

Your command worked perfect.

alexfikl:
> Your command worked perfect.

Awesome! :grin:

Yeah, the `html-tags` check is not on by default. From a quick look, these are the default ones
```python
"doctor-default-checks": ["files", "biblatex-required-keys", "bibtex-type", "refs"]
```

You can add more with something like this in your configuration file
```ini
doctor-default-checks-extend = ["html-tags", "html-codes"]
```
or you can just run all of them from the command line using :rocket: 
```bash
papis doctor --fix --explain --all-checks --all '.'
```

PR #1009, opened 2025-04-27
Structure configuration state

hseg:
Note: This PR is currently a WIP which I won't be touching until 2025-05-08 at the earliest, I'm parking it here for self-accountability and to provide a place to discuss it.

Following on https://github.com/papis/papis/pull/1006#issuecomment-2833433395, this PR makes the configuration state explicit, clarifying its fields and their types.

The idea is to have a `...Config` type for each registered plugin, as well as
for the papis core config, so that the global configuration state becomes a
`Dict[str,Dict[Optional[str],Config]]`.

The intent is that `config[lib][plugin][key] == (set_lib(lib); general_get(key, plugin))`, with the defaulting and cross-section search done at the time `config` is created, with plugins registering a function `ConfigParser -> ...Config` to be used to populate their own config objects. The `None` plugin references the core papis configuration.

If everything goes well, this should enable us to generate `doc/source/default-settings.rst` from the `...Config` types and eventually deprecate the use of `general_get` and friends outside the config-parsing code.

### DIFF
--- a/doc/source/default-settings.rst
+++ b/doc/source/default-settings.rst
@@ -451,7 +451,6 @@ Add options
     If this setting is ``None`` the template ``{doc[papis_id]}`` is used.
 
 .. papis-config:: add-file-name
-    :type: str
 
     Set the default file name for newly added documents, similarly to
     :confval:`add-folder-name`. If it is not set, the names of the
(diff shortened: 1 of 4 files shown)


hseg:
Alternatively (if this proves too difficult), we could use eg Cerberus to ensure the `Configuration` that results after configuration parsing indeed contains all the keys of expected types. However, I suspect this is likely to already be most of the way towards imposing at least a phase distinction between configuration parsing and using its results, so I don't know how much we'd benefit from avoiding classes here.

PR #1010, opened 2025-04-30
docs: fix grammar, small mistakes and inconsistencies

tuurep:
Hi, I have worked on this PR for a long time after noticing some particularly glaring mistakes in the docs, and then decided to read through the *entire* docs and fix all mistakes I can find :smile:

In hindsight I should have done this in smaller pieces.

My goal is that this should be safe to merge, what I've done (in order of importance:)

1. Fix formatting errors, e.g. codeblocks have paragraph text
    - And a couple logical errors in examples and such.
3. Fix grammar mistakes
    - But there are some cases where I reasoned if something is written in a needlessly confusing way, I touch it up to aid readability.
4. Improve consistency
    - Most prominently and creating a lot of diffs: some `click` short-help messages conclude with a period and others don't. I went for period at the end always, because the `-h` flag does that whether we want it or not.

PR #1011, opened 2025-05-01
Do not truncate click help messages

alexfikl:
Our hack around not truncating click help messages does not seem to work very well as shown in https://github.com/papis/papis/pull/1010#issuecomment-2845303648. This tries a different version that seems to work better.

cc @tuurep

### DIFF
--- a/papis/commands/__init__.py
+++ b/papis/commands/__init__.py
@@ -12,6 +12,31 @@ COMMAND_EXTENSION_NAME = "papis.command"
 EXTERNAL_COMMAND_REGEX = re.compile(".*papis-([^ .]+)$")
 
 
+class FullHelpCommand(click.core.Command):
+    """This is a simple wrapper around :class:`click.core.Command` that does
+    not truncate the short help messages.
+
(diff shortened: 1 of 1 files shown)


alexfikl:
Hopefully this doesn't break anything :-s

tuurep:
Nice, tested out both CLI output and the readthedocs `COMMANDS > bibtex > CLI` and looks as expected!

Other than some of them missing our newly formed 'standard' with the dot at end :smile: I can point those out. Should I perhaps add those as part of #1010 ?

alexfikl:
> Nice, tested out both CLI output and the readthedocs `COMMANDS > bibtex > CLI` and looks as expected!

Great! :tada:

> Other than some of them missing our newly formed 'standard' with the dot at end 😄 I can point those out. Should I perhaps add those as part of #1010 ?

Yeah, if you could add them in #1010 that would be great! (If they're not already there)

tuurep:
Yeah I could not find any more missing dots right now, on that branch :D if you see any, point out please

Issue #1027, opened 2025-06-04
Library picker doesn't display colors via fzf


octvs:
When the picktool is set to  fzf, running `papis --pick-lib open` shows 

```txt
<ansired>lib1</ansired> <ansiblue>['/home/user/lib1']</ansiblue>
<ansired>lib2</ansired> <ansiblue>['/home/user/lib2']</ansiblue>
```

I suspect the culprit is the type check [over here](https://github.com/papis/papis/blob/d85cd00f56bae148c2ddd212f84997c9f1d9dfac/papis/fzf.py#L142-L149).

alexfikl:
I'm not a fzf user, so not quite sure what's going on, but it looks like that's just using a default value that's meant for the default papis picker.

I would guess you'll have to set [library-header-format](https://papis.readthedocs.io/en/latest/configuration.html#confval-library-header-format) to something that fzf understands (using colorama)? You can probably use [fzf-header-format](https://papis.readthedocs.io/en/latest/configuration.html#confval-fzf-header-format) as inspiration.

octvs:
Maybe I should have elaborated more, but of course I have tried that. Setting
the default value for `fzf-header-format`:

```sh
$ grep "library-header-format" .config/papis/config
library-header-format = {c.Fore.MAGENTA}{library[name]}{c.Style.RESET_ALL} :: {c.Fore.CYAN}{library[paths]}{c.Style.RESET_ALL}
$ papis --pick-lib open 
Traceback (most recent call last):
  File "/nix/store/azxfkmvygrzjjlvv590n5963shpq63d5-python3.12-papis-0.14.1/lib/python3.12/site-packages/papis/format.py", line 165, in format
    return self.psf.format(fmt, **{doc_name: doc}, **additional)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/8w718rm43x7z73xhw9d6vh8s4snrq67h-python3-3.12.10/lib/python3.12/string.py", line 190, in format
    return self.vformat(format_string, args, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/8w718rm43x7z73xhw9d6vh8s4snrq67h-python3-3.12.10/lib/python3.12/string.py", line 194, in vformat
    result, _ = self._vformat(format_string, args, kwargs, used_args, 2)
[...]

PR #1028, opened 2025-06-04
Fix fzf pick lib colors

octvs:
Refer to #1027.


octvs:
The first commit itself solves the issue for me, however it would import
`colorama` even when `fzf` is not used. In spirit of [this], I added a second
commit which checks for `fzf`. 

I marked both as `wip`, let me know which one you think looks better @alexfikl.

[this]: https://github.com/papis/papis/commit/8cd1a2354651df3bdb5a76d014a9cb3a85d681eb#diff-d12f72a5a6c99dbfb75efcbba7c3998871dfa07830f3de8ecb14158a13608500

alexfikl:
I think it's fine to just import colorama without the check. It gets imported anyway for the logging, so it should be nicely cached (plus it's a very small library :grin:)

It breaks that abstraction a bit, but that's a bigger fish to fry..

octvs:
> I think it's fine to just import colorama without the check. It gets imported anyway for the logging, so it should be nicely cached (plus it's a very small library 😁)

Alright, I took that one off and elaborated on the commit message more.

> It breaks that abstraction a bit, but that's a bigger fish to fry..

Well, everytime I intend to modify papis src I get slightly overwhelmed. And
quit with thoughts based around not understanding the structure well enough.
Therefore, I won't be any help there :).

alexfikl:
> Well, everytime I intend to modify papis src I get slightly overwhelmed.

Haha, no worries! It took me a few years of drive-by patches to get slightly more comfortable :( 

There are a lot of moving pieces, so if you ever feel inclined to make something look less confusing (either by improving the docs or moving the code around a bit), it's very welcome! Even just complaining about it might get us moving to improve it :grin:

octvs:
Thanks for the kind & quick responses. Maybe I will have the chance in future to sit down and finally grasp what is happening. 

Until then, looks like the drive-by patches are the way for me as well :grin:

alexfikl:
In it goes :rocket:

PR #1034, opened 2025-06-23
Fixed passing --urls to `papis addto`

f0rki:
There was a bug in the new addto code that attempted to copy/symlink the https URL instead of the downloaded tempfile. Changed the return type of the `rename_document_files` function to also return the original file paths, but with all URLs replaced with either "" or a temporary path.

### DIFF
--- a/tests/commands/test_addto.py
+++ b/tests/commands/test_addto.py
@@ -104,7 +104,6 @@ def _mock_download_document(
     return create_random_file(filetype=filetype, prefix=prefix, suffix=suffix)
 
 
-# @pytest.mark.skipif(sys.platform.startswith("win"), reason="doesn't work on windows")
 def test_addto_cli_urls(tmp_library: TemporaryLibrary,
                         monkeypatch: pytest.MonkeyPatch) -> None:
     from papis.commands.addto import cli


f0rki:
Meh. sorry. it seems this broke a lot of tests... Will try to see if I can find a better way.

alexfikl:
Very quickly dropping by, but couldn't this be "fixed" by always copying downloaded files? Is that not the expected behavior? 

I'm thinking of just adding a check in `papis.commands.add` and `papis.commands.addto` for that instead of modifying `rename_document_files`. 

P.S. Can't believe we didn't catch this already though :( We should add a test!

f0rki:
> couldn't this be "fixed" by always copying downloaded files?

agree. that probably should be the behavior. However, this doesn't fix the problem here.
The error occured when copying the file, not when linking. The problem is the copy operation in `addto` doesn't know about the downloaded file path.

Maybe the `rename_document_files` function is not the best place to implement downloading?

> P.S. Can't believe we didn't catch this already though :( We should add a test!

yeah :S I see what I can do.

alexfikl:
> The error occured when copying the file, not when linking. The problem is the copy operation in `addto` doesn't know about the downloaded file path.

Ah, yes :facepalm: Then, like you hinted, we could do
```
local_file_list = download_remote_files(filepaths) # using `papis.utils.download_document`
new_file_list = rename_document_files(document, local_file_list)
```
? Having `rename_document_files` also do the download was definitely not the best idea :\ Then we can check in `addto` if `is_remote(filepaths[i])` and always copy `local_file_list[i]` to `new_file_list[i]` (or something like that).

Does that make more sense? Sorry again for the confusion and late reply :(

f0rki:
I implemented the approach and split the download and rename functions. I am not super happy with the structure of the code, but it should handle passing all kinds of URLs and files (also those that don't exist or fail to download) 

I think the best way would be to do the rename, download, and copy in a single function within a single loop. This way way we wouldn't need to pass around the bool whether it was a remote file.

f0rki:
I have no idea why the tests are so flaky. Maybe because of the PDF download?

f0rki:
> I have no idea why the tests are so flaky.

@alexfikl I tried doing `time.sleep(1)` and fsync between invoking the papis cli and checking the docs. I then tried running a local http server to make sure it's not the network connection. Any other suggestions on why sometimes the tests fail?

alexfikl:
> @alexfikl I tried doing `time.sleep(1)` and fsync between invoking the papis cli and checking the docs. I then tried running a local http server to make sure it's not the network connection. Any other suggestions on why sometimes the tests fail?

Sorry for the late reply! This looked like a bigger review and I've been putting it off :(

Can't say that I have any idea just from looking at the code / test failures.. :\ Seeing as some tests pass and some don't rather randomly, I'm inclined to think it's some rate limiting somewhere along the wire.. but not sure why none of your solutions seemed to help.

I would recommend to just make the test as simple as it was initially (without the HTTP server and whatnot) and just mark it as `xfail` for now. I assume the code actually works in normal (command-line) usage? I'll try to test it out a bit too.

papis/commands/addto.py
```diff
@@ -48,7 +48,7 @@
 
 
 def run(document: papis.document.Document,
-        filepaths: List[str],
+        filepaths_arg: List[str],
```
alexfikl: Would prefer the original name? Maybe just give more expressive names to the new variables below (`renamed_filepaths`, `new_filepaths`).

papis/commands/addto.py
```diff
@@ -57,35 +57,63 @@ def run(document: papis.document.Document,
         raise DocumentFolderNotFound(papis.document.describe(document))
 
     import shutil
-    from papis.paths import symlink, rename_document_files
+    from papis.paths import (symlink, rename_document_files,
+                             download_remote_files, is_remote_file)
+
+    # ensure all remote files are downloaded, before we start renaming.
+    # the number of
```
alexfikl:
```suggestion
    # ensure all remote files are downloaded before we start renaming
```

papis/commands/addto.py
```diff
@@ -57,35 +57,63 @@ def run(document: papis.document.Document,
...
+    local_files = download_remote_files(filepaths_arg)
+
+    # we only symlink a file it isn't a downloaded file that is stored in
+    # a temporary directory. We keep track of a boolean per file that
+    # tells us whether to link
```
alexfikl, line 68: Can this be collapsed into just one loop (i.e. into the original one below)? I was thinking of something like
```python
local_files = download_remote_files(filepaths)
renamed_local_files = rename_document_files(document, local_files)
for orig_file_path, in_file_path, out_file_name in zip(filepaths, local_files, renamed_local_files):
	if not exists(in_file_path):
		continue

	if exists(out_file_path):
		continue
	
	has_valid_files = True
	if link and not is_remote_file(orig_file_path):
		...
	else:
		...

if not has_valid_files:
	logger.warning(...)
	return
```

papis/commands/addto.py
```diff
@@ -57,35 +57,63 @@ def run(document: papis.document.Document,
...
+
         out_file_path = os.path.join(doc_folder, out_file_name)
         if os.path.exists(out_file_path):
             logger.warning("File '%s' already exists. Skipping...", out_file_path)
             continue
 
-        if link:
-            logger.info("[LN] '%s' to '%s'.", in_file_path, out_file_name)
+        if link_file:
+            logger.debug("[LN] '%s' to '%s'.", in_file_path, out_file_name)
```
alexfikl: Why make these into debug logs? It seemed generally useful to see what the files got renamed to.

papis/commands/addto.py
```diff
@@ -96,10 +124,9 @@ def run(document: papis.document.Document,
 @papis.cli.sort_option()
 @click.option("-f",
               "--files",
-              help="File fullpaths to documents.",
-              multiple=True,
-              type=click.Path(exists=True))
-@click.option("-u", "--urls", help="URLs to documents.", multiple=True)
+              help="File paths or URLs to documents.",
+              multiple=True)
+@click.option("-u", "--urls", help="(deprecated) use -f", multiple=True)
```
alexfikl, line 130: Agreed!

papis/commands/addto.py
```diff
@@ -137,4 +164,5 @@ def cli(query: str,
     try:
         run(document, files + urls, file_name=file_name, git=git, link=link)
     except Exception as exc:
-        logger.error("Failed to add files.", exc_info=exc)
+        logger.debug("Failed to add files.", exc_info=exc)
+        raise
```
alexfikl: This shouldn't raise. The full tracebacks are generally very loud and verbose, so they're hidden under the `--log DEBUG` flag.

papis/paths.py
```diff
@@ -337,15 +337,40 @@ def get_document_unique_folder(
...
+
+    new_files: List[Optional[str]] = []
+    for in_file_path in in_document_paths:
+        if is_remote_file(in_file_path):
+            local_in_file_path = download_document(in_file_path)
+            if local_in_file_path:
+                new_files.append(local_in_file_path)
+            else:
+                logger.debug("skipping failed remote path '%r'", in_file_path)
+                new_files.append(None)
```
alexfikl: Just don't append the failed downloads?

Also, I think `download_document` already has some log messages for failed downloads, so this doesn't seem necessary?

papis/paths.py
```diff
@@ -337,15 +337,40 @@ def get_document_unique_folder(
...
+
+    return new_files
+
+
 def rename_document_files(
         doc: DocumentLike,
         in_document_paths: Iterable[str], *,
-        file_name_format: Optional[Union[AnyString, Literal[False]]] = None,
-        allow_remote: bool = True,
+        file_name_format: Optional[Union[AnyString, Literal[False]]] = None
```
alexfikl: I don't know about straight up removing this, since it breaks backwards compatibility (I'm using it in some out-of-tree plugins :grin:). Maybe have a little message?
```
if allow_remote:
	warnings.warn("This is deprecated!!", DeprecationWarning)
```

f0rki:
@alexfikl what do you think of moving the copy/linking logic into the rename_documents function? I think this could simplify everything a bit, i.e., no need to make sure the multiple lists are in sync/handled correctly.

f0rki:
One test failure might be related to the way temporary files are named:
```
E           AssertionError: assert (False or False)
E            +  where False = <built-in method startswith of str object at 0x1160c5730>('single-page-test.pdf')
E            +    where <built-in method startswith of str object at 0x1160c5730> = 'tmpk6h7c2-h'.startswith
E            +  and   False = <built-in method startswith of str object at 0x1160c5730>('tmpk6h7c2_h')
E            +    where <built-in method startswith of str object at 0x1160c5730> = 'tmpk6h7c2-h'.startswith
```

`'tmpk6h7c2_h'` vs. `tmpk6h7c2-h` it seems maybe the random filename contains a `_` sometimes, which is then changed to `-` by papis, which the testcase doesn't account for.

papis/paths.py
f0rki: No, I need to append `None` here, becasue I rely on the fact that `download_remote_files` returns a list that maps 1:1 to the filepaths argument of the addto command. This is used to track whether the file was remote or not.

papis/commands/addto.py
f0rki: I tried to do this first, but the problem is that `zip(filepaths, local_files, renamed_local_files)` is not necessarily correct, because failed files are removed from some of those lists and kept in others. So if you e.g., fail to download or a file does not exist, then those lists have different lengths, which probably has very weird side effects down the road.

papis/commands/addto.py
f0rki: It seemed unnecessarily verbose to me and I thought it was maybe an oversight from when this feature came in. Maybe it is just me personally, but I don't want to see this. I just want papis to do the thing I configured it to.

papis/paths.py
f0rki: done

f0rki:
Great... It seems windows seems to ignore the prefix argument in `create_random_file` for whatever reason. 

EDIT: I am skipping the test on windows for now...

f0rki:
>  would recommend to just make the test as simple as it was initially (without the HTTP server and whatnot) and just mark it as xfail for now.

The http server now binds to a somewhat random tcp port, which seems to make the whole thing a lot less likely to fail. It seems the kernel isn't as quick to release the port as we would like to. Maybe there was also some interference between the different testcase runs (i.e., containers on the same host?).

papis/commands/addto.py
alexfikl: Fair enough, it's tricky to show just the right amount™ of information here. For the moment, this is inconsistent with `papis add` (which also has the same info logs) and a bit out of scope, so maybe leave it for another PR?

papis/commands/addto.py
alexfikl: Ah, right :facepalm: Well that's not very nice at all, but we can keep it for now.

papis/paths.py
alexfikl: Yep, missed that like in the previous comment :facepalm:

papis/paths.py
```diff
@@ -337,15 +337,40 @@ def get_document_unique_folder(
...
+    new_files: List[Optional[str]] = []
+    for in_file_path in in_document_paths:
+        if is_remote_file(in_file_path):
+            local_in_file_path = download_document(in_file_path)
+            if local_in_file_path:
+                new_files.append(local_in_file_path)
+            else:
+                new_files.append(None)
+        else:
+            new_files.append(in_file_path)
```
alexfikl:
```suggestion
        if is_remote_file(in_file_path):
            local_in_file_path = download_document(in_file_path)
        else:
        	local_in_file_path = in_file_path
        new_files.append(local_in_file_path)
```
? Seems a bit easier to follow to me.

papis/commands/addto.py
```diff
@@ -57,35 +57,63 @@ def run(document: papis.document.Document,
...
+        if os.path.exists(filepath):
+            new_filepaths.append((filepath,
+                                  link if not is_remote_file(orig_filepath) else False))
+        else:
+            logger.warning("Skipping non-existent file: %r", filepath)
 
-    new_file_list = rename_document_files(
-        document, filepaths, file_name_format=file_name
+    if not new_filepaths:
+        logger.error("No valid files provided")
```
alexfikl:
```suggestion
        logger.error("No valid files provided.")
```

papis/commands/addto.py
```diff
@@ -57,35 +57,63 @@ def run(document: papis.document.Document,
...
+    for orig_filepath, filepath in zip(filepaths, local_files):
+        # skip remote files that haven't been properly downloaded
+        if not filepath:
+            continue
+
+        if os.path.exists(filepath):
+            new_filepaths.append((filepath,
+                                  link if not is_remote_file(orig_filepath) else False))
+        else:
+            logger.warning("Skipping non-existent file: %r", filepath)
```
alexfikl:
```suggestion
            logger.warning("Skipping non-existent file: '%s'.", filepath)
```
? I think I'd prefer using `%s` here because it'll look the same when we switch to `pathlib.Path`.

papis/paths.py
```diff
@@ -369,46 +394,40 @@ def rename_document_files(
         except ValueError:
             file_name_format = None
 
+    if allow_remote is not None:
+        warn("The argument `allow_remote` to `rename_document_files` is deprecated",
+             DeprecationWarning)
```
alexfikl:
```suggestion
    if allow_remote is not None:
        warn("The argument `allow_remote` to `rename_document_files` is deprecated "
             "and will be removed in version 0.16.",
             DeprecationWarning, stacklevel=2)
    else:
        allow_remote = True
```
The handling for `allow_remote` should still stay in here even with the warning so it doesn't break callers to this function.

papis/testing.py
```diff
@@ -64,6 +64,7 @@ def create_random_file(filetype: Optional[str] = None,
     else:
         raise ValueError("Unknown file type: '{}'".format(filetype))
 
+    # TODO: there seems to be a problem with using the prefix argument on windows
```
alexfikl: Hm, that's scary. I couldn't find anything in the docs or from a quick search to mention prefix issues in Windows. What problem was there?

tests/commands/test_addto.py
```diff
@@ -1,8 +1,83 @@
...
+
+__SERVER_THREAD = None
+__SERVER_STR = None
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+
+
+@pytest.fixture(scope="session")
+def local_http_server() -> Optional[str]:
```
alexfikl: I'm still not sold at all on this server thing here.

I would recommend monkey patching `papis.downloaders.download_document` instead to just return a local temporary file with `create_random_file`. That would remove the whole remote aspect of this and just test that the URL handling is correct in `addto`, which is what we really want here. What do you think?

papis/paths.py
```diff
@@ -337,15 +337,40 @@ def get_document_unique_folder(
...
     return uri.startswith("http://") or uri.startswith("https://")
 
 
+def download_remote_files(in_document_paths: Iterable[str]) -> List[Optional[str]]:
+    """
+    Download all remote filepaths that are provided in the document list.
+
+    :param in_document_paths: a list of filename paths and URLs.
+    :returns: a list of files, where each remote file is replaced with a
+        temporary local file.
```
alexfikl:
```suggestion
    :returns: a list of files, where each remote file is replaced with a
        temporary local file. If there is an error while downloading the 
        remote file, *None* is used instead.
```
?

alexfikl (commented):
Gave this another look and left some comments! I would still like to simplify the tests a bit, but besides that it's looking good to me.

> 'tmpk6h7c2_h' vs. tmpk6h7c2-h it seems maybe the random filename contains a _ sometimes, which is then changed to - by papis, which the testcase doesn't account for.

There's a hacky workaround for this in the `test_addto_cli` test (that special `eq` function) that may help here too?

> @alexfikl what do you think of moving the copy/linking logic into the rename_documents function? I think this could simplify everything a bit, i.e., no need to make sure the multiple lists are in sync/handled correctly.

Yeah, it would be nicer to have something like that that could also be shared with `papis add`. Follow-up refactoring PR? :grin:

papis/testing.py
f0rki: The file was just named `.pdf` without any prefix or random parts.

tests/commands/test_addto.py
f0rki: OK. I think I can take a shot at doing this sometime next week. The server things seemed to me the easiest way to achieve realistic testing, i.e., also download failures. But that should be doable with monkey patching too.

papis/testing.py
f0rki: I can re-enable the test and open an issue.

tests/commands/test_addto.py
alexfikl: I can help rip it out one of these days, if that's ok with you?

> The server things seemed to me the easiest way

My main reason to be reluctant about it is that no other tests was quite as annoying (or we just `xfail`ed the few that were), so it seems like a big piece of code to maintain just for this.

> achieve realistic testing, i.e., also download failures

With all these download errors and cloudflare doing its thing, it seems increasingly unlikely that we'll have much luck in that direction :( I don't know, but I would err on the side of having a simple test for the logic in papis and leaving the thirdparty services alone.

papis/testing.py
alexfikl: Yeah, opening an issue sounds good to me! I'd like to look into this a bit too when I find the time because it's quite a scary error to have (a lot of tests use this function and pass on windows).

papis/paths.py
```diff
@@ -369,46 +392,43 @@
         except ValueError:
             file_name_format = None
 
+    if allow_remote is not None:
+        warn("The argument `allow_remote` to `rename_document_files` is deprecated "
+             "and will be removed in version 0.16.",
+             DeprecationWarning, stacklevel=2)
+    else:
+        allow_remote = True
```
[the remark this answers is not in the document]
f0rki: @alexfikl remove this again?

papis/testing.py
f0rki: seems it was a problem on my end

f0rki:
I think I addressed everything. should be good to merge now @alexfikl

alexfikl:
> I think I addressed everything. should be good to merge now @alexfikl

Looks good to me too! Thanks for all the work on this! In it goes :rocket:

Issue #988, opened 2025-03-23
Export appends to file


jghauser:
Is it on purpose that `papis export` appends to the file specified by `--out` rather than overwriting it? I usually export my whole library to `.bib` to use elsewhere and noticed other software complaining about duplicate keys.

alexfikl:
Hm, I don't recall, but both appending and overwriting have a usecase usually. Maybe we can add a flag?

tuurep:
Hi I was thinking about this one day when exporting bibs

I agree, both overwrite and append would be great to have

I'd like to point out that when it appends, it doesn't add a newline at the point where it's appended:

![Image](https://github.com/user-attachments/assets/74e0256f-1d51-4783-baab-55abb7a6b613)

but otherwise separates entries by newline

Maybe it could append with a newline, if last line of buffer isn't blank?

tuurep:
I may be interested to work on this, but can we agree on which should be the default: append or overwrite?

alexfikl:
> I may be interested to work on this, but can we agree on which should be the default: append or overwrite?

I would say the default should be overwrite so that we don't change existing behavior. You can just add a little `--append` flag then?

> Maybe it could append with a newline, if last line of buffer isn't blank?

I don't know about this.. I'm not opposed if you write up a PR, but we're probably not going to get into the business of formatting BibTeX files, so there's always going to be some inconsistency or nitpick there :\

I've been using [bibtex-tidy](https://github.com/FlamingTempura/bibtex-tidy) as a post-processing step and it works pretty nicely :D

tuurep:
I think we're getting these mixed up, as the current behavior is to append (with an ugly non-linebreak but I digress)

> don't change existing behavior

With this approach, should we then go with append default and `--overwrite` flag? 😄 

Personally, I'm going to find the overwrite behavior more useful as I want to somehow automate generating the .bib (with `bibtex-tidy` probably used, great tool btw 👍), and append would mostly serve for re-adding duplicate entries :/

> we're probably not going to get into the business of formatting BibTeX files,

Generally speaking, agreed. I'll see if I work on this how feasible it would be to add a newline in a way that I'd expect to be there on append, or if that's the responsibility of something else.

tuurep:
@jghauser I see the append by default also confused you (feels contrary to expectation), might I suggest we go overwrite by default, even if it changes current behavior?

alexfikl:
> I think we're getting these mixed up, as the current behavior is to append

[...]

PR #1036, opened 2025-07-05
Export (over)write outfile by default, add `--append` and `--batch` flags

tuurep:
- [x] implementation
    - done but append formats badly
        - may be in the scope of another issue and PR (see 2 comments below)
- [x] docs
- [x] tests (should I add any?)

Issue: #988

### DIFF
--- a/papis/commands/export.py
+++ b/papis/commands/export.py
@@ -13,17 +13,24 @@ Some examples of its usage are:
 
     papis export --format bibtex 'author : einstein'
 
-or export all of them:
+- Note that BibTeX is the default format. The ``--format`` flag can be omitted:
 
 .. code:: sh
 
-    papis export --format bibtex --all 'author : einstein'
+    papis export 'author : einstein'
 
-- Export all documents to BibTeX and save them into a ``lib.bib`` file:
+- Export all documents with author Einstein to BibTeX:
 
 .. code:: sh
 
-    papis export --all --format bibtex --out lib.bib
+    papis export --all 'author : einstein'
+
+- Export all documents in your default library to BibTeX and save them into a
+  ``lib.bib`` file:
+
+.. code:: sh
+
+    papis export --all --out lib.bib
 
 - Export a folder of one of the documents matching the word ``krebs``
   into a folder named ``interesting-document``:
--- a/papis/commands/export.py
+++ b/papis/commands/export.py
@@ -155,39 +155,31 @@ def cli(query: str,
     ret_string = run(documents, to_format=fmt)
 
     if ret_string and not folder:
-        if out is not None:
-            if append:
-                if os.path.exists(out):
-                    logger.info("Appending to '%s'.", out)
-                else:
-                    logger.info("Writing to '%s'.", out)
-
-                with open(out, "a") as fd:
-                    fd.write(ret_string)
-                return
-
-            if os.path.exists(out) and not batch:
-                if papis.tui.utils.confirm(f"File '{out}' already exists. Overwrite?"):
-                    logger.info("Overwriting '%s'.", out)
-                    with open(out, "w") as fd:
-                        fd.write(ret_string)
-                else:
-                    logger.info("Aborting.")
-                return
-
-            if os.path.exists(out):
-                # Batch write
-                logger.info("Overwriting '%s'.", out)
-            else:
-                logger.info("Writing to '%s'.", out)
-
-            with open(out, "w") as fd:
-                fd.write(ret_string)
-            return
-        else:
+        if out is None:
             logger.info("Dumping to STDOUT.")
             click.echo(ret_string)
             return
+
+        mode = "a" if append else "w"
+
+        if os.path.exists(out):
+            if append:
+                msg = f"Appending to '{out}'."
+            else:
+                if not batch:
+                    prompt = f"File '{out}' already exists. Overwrite?"
+                    if not papis.tui.utils.confirm(prompt):
+                        logger.info("Aborting.")
+                        return
+                msg = f"Overwriting '{out}'."
+        else:
+            msg = f"Writing to '{out}'."
+
+        logger.info(msg)
+        with open(out, mode) as fd:
+            fd.write(ret_string)
+
+        return
 
     import shutil
     for document in documents:
--- a/papis/commands/export.py
+++ b/papis/commands/export.py
@@ -55,6 +55,7 @@ from typing import List, Optional, Tuple
 import click
 
 import papis
+import papis.tui.utils
 import papis.utils
 import papis.document
 import papis.cli
@@ -114,6 +115,14 @@ def run(documents: List[papis.document.Document], to_format: str) -> str:
     help="Format for the document.",
     type=click.Choice(available_formats()),
     default="bibtex",)
+@papis.cli.bool_flag(
+    "-p",
+    "--append",
+    help="Append to outfile instead of overwriting.")
+@papis.cli.bool_flag(
+    "-b",
+    "--batch",
+    help="Do not prompt when overwriting a file.")
 def cli(query: str,
         doc_folder: Tuple[str, ...],
         sort_field: Optional[str],
@@ -121,6 +130,8 @@ def cli(query: str,
         folder: str,
         out: str,
         fmt: str,
+        append: bool,
+        batch: bool,
         _all: bool) -> None:
     """Export a document from a given library."""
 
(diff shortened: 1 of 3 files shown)


tuurep:
The implementation was almost ready for review but there's still this issue:

image

Appends straight from the end of the last line

Not sure if it has a relation to https://github.com/papis/papis/issues/988#issuecomment-2746178694 but we'll see soon

On https://github.com/papis/papis/issues/988#issuecomment-2746178694 I've learned that `.bib` is the only filetype that wants the double linebreak so I haven't decided what to do about it, if anything.

**Edit:**

I see this probably comes down to this problem (no newline character at the end of the `ret_string`):

https://stackoverflow.com/a/5813359

Confirmed: not introduced by this PR, this always happened but it can be confusing when I edit the outfile in my editor, it adds the newline to the end-of-file on save.

**Edit:**

Formats `bib`, `csl` and `json` have this problem, but not `typst` and `yaml`

tuurep:
Eh, I'll have to look exactly how the different plugins do it, when using only 1 document vs. many documents

Because it looks like `yaml` adds a `---` between documents when exporting multiple at once, but the append doesn't add that `---`

Seems to me like something that we can completely fix (the `bib` newlines too) but maybe not in the scope of this PR

`json` specialty:

Multi-export is like:

```
[
  {
    doc1
  },
  {
    doc2
  }
]
```

and appending produces this:

```
[
  {
    doc1
  },
  {
    doc2
  }
]
[
  {
    doc3
  }
]
```

papis/commands/export.py
```diff
@@ -155,40 +155,32 @@ def cli(query: str,
...
+                msg = f"Appending to '{out}'"
+            else:
+                if not batch:
+                    prompt = f"File '{out}' already exists. Overwrite?"
+                    if not papis.tui.utils.confirm(prompt):
+                        logger.info("Aborting.")
+                        return
+                msg = f"Overwriting '{out}'"
+        else:
+            msg = f"Writing to '{out}'"
```
tuurep: On append, if file doesn't already exist, it says "Writing" instead of "Appending"

This is intentional because to me it felt out of place the other way around, i.e. "appending" to a file that doesn't exist yet

Although *technically* that's not wrong either. Thoughts?

papis/commands/export.py
```diff
@@ -114,13 +115,23 @@ def run(documents: List[papis.document.Document], to_format: str) -> str:
     help="Format for the document.",
     type=click.Choice(available_formats()),
     default="bibtex",)
+@papis.cli.bool_flag(
+    "-p",
+    "--append",
```
tuurep, line 127: The `-p` shorthand for `--append` is something I saw in `papis tag` and `papis update`, although in those it's not a bool flag.

papis/commands/export.py
```diff
@@ -144,13 +155,30 @@ def cli(query: str,
...
+        if out is None:
             logger.info("Dumping to STDOUT.")
             click.echo(ret_string)
+            return
+
+        mode = "a" if append else "w"
+
+        if os.path.exists(out):
+            if append:
+                msg = f"Appending to '{out}'"
```
alexfikl:
```suggestion
                msg = f"Appending to '{out}'."
```
I think most logger messages have a full stop at the end, so it's good to be consistent.

alexfikl (commented):
Left a very little nitpick, but otherwise the code looks good to me!

> docs

It feels like the usage is pretty self-explanatory, but if you want to add some docs, that would be very welcome :D

> tests (should I add any?)

Always a good idea to add some tests! At least for the append side, it would be good to have a test that documents how the different formats don't add new lines correctly/consistently?

> Formats bib, csl and json have this problem, but not typst and yaml

Have you checked that all these formats work on append? I would expect
* `json` to not work properly as is, since for multiple documents it exports a list. I would expect this to append to that list, not add another list to the file? Does the result work with `json.load`?
* `yaml` adds `----` between documents, so just appending doesn't sound like it does the right thing?

This should be more obvious with some tests that write a file, append to the file an then try to read it back in.

EDIT: Just noticed that you have a comment addressing this :facepalm: Sorry for the noise!

But yes, we'll need to figure out a way to make these work as well. Maybe by extending the exporter interface to do most of the work? I can't really see a way around it..

papis/commands/export.py
alexfikl: Yeah, that should be fine, since the difference isn't particularly meaningful in that case. I would have probably just used "Writing to 'filename'" every time :))

papis/commands/export.py
alexfikl: That sounds good!

papis/commands/export.py
tuurep: Thank you, I accidentally dropped the dots while refactoring

Force-amended to the refactor commit

tuurep:
> It feels like the usage is pretty self-explanatory, but if you want to add some docs, that would be very welcome :D

Yes I went through the `export` docs now and the new flags are indeed very self-explanatory

What I would like to add is a mention that `--format bibtex` is default, i.e. you can completely omit the flag

For about a year I have been manually adding `--format bibtex` (not even the short-form `-f` :D)

This is pretty offtopic for this PR though?

> Just noticed that you have a comment addressing this [`json`/`yaml` special append] 🤦 Sorry for the noise!

Haha yeah but looks like we have the exact same idea so that's great.

About the tests, should they be added in this PR or the followup dealing with the exporters' appending behavior?

alexfikl:
> What I would like to add is a mention that `--format bibtex` is default, i.e. you can completely omit the flag
> This is pretty offtopic for this PR though?

If you want, it's definitely fine to add it from my side. It's just a little sentence somewhere in there.
 
> About the tests, should they be added in this PR or the followup dealing with the exporters' appending behavior?

Maybe add a little test with some dummy documents for the BibTeX exporter in this PR? We can do something more extensive when we tackle fixing the others.

Ignoring the tests, this looks good to me now, so let me know when it can go in :grin:

tuurep:
I'll do it today

I have to ask, does this mean I'm adding a test that we know will fail?

And let's say, https://aur.archlinux.org/packages/papis-git tries to run that test, it fails, and installation fails?

To be honest, the AUR package always hit some failing tests in the build, so I had to pass a `--mflags="--nocheck"` with `yay` when installing. I want to fix that but I'm not quite sure:

1. Why does it run the tests to begin with?
2. Is the issue the tests, or the AUR package?

By the way, do you have the same issue with that package? You've mentioned you modify the PKGBUILD, but in what way? :)

alexfikl:
> I have to ask, does this mean I'm adding a test that we know will fail?

There are two options:
1. Add a test for the current output, i.e. a missing newline in there is expected. This should just test that the append actually does what it should.
2. Add a test for the desired output, but then this should be marked with `pytest.mark.xfail` to not break the CI.

I would lean towards 1, perhaps with a little comment that this should be fixed by someone sometime :grin:

tuurep:
Got it, thanks

alexfikl:
> To be honest, the AUR package always hit some failing tests in the build, so I had to pass a `--mflags="--nocheck"` with `yay` when installing. I want to fix that but I'm not quite sure:
>     1. Why does it run the tests to begin with?
>     2. Is the issue the tests, or the AUR package?

Could open a discussion about this? I recall you mentioned it before, but we can try debugging it some more :\ We have some tests that download things from the web that could fail more randomly, so it could be a problem on our side..

tuurep:
While taking a look whether the default of `--format` can be changed, I noticed an incorrect example of `export` command in the settings docs, fixed: [3fb1624](https://github.com/papis/papis/pull/1036/commits/3fb16246b68376e6244b935d4950355323e9f3fa)

I don't mean to do any more nitpickery here :smile: 

I reworded the `export` examples section quite a bit, is this ok? [45a97c9](https://github.com/papis/papis/pull/1036/commits/45a97c9f01304825a592193049e510006b323978)

tuurep:
Here's an attempt at a test,

I have to say I don't quite know what I'm doing. Does it look like it makes much sense?

Edit: fixed lint, whitespace errors

tuurep:
> Could open a discussion about this [AUR check() issues]?

Opened: https://github.com/papis/papis/discussions/1037

Unfortunately had trouble getting to the failing tests point in the build. `flake8` vaguely seems to always be involved though?

alexfikl:
> I have to say I don't quite know what I'm doing. Does it look like it makes much sense?

Yeah, looks good to me! Was there anything else you wanted to add here? If not, this can be merged.

tuurep:
Oh ok nice.

Yeah we can merge, the tests can be expanded after #1038

Thanks!

tuurep:
Sorry, I was deleting previous merged branches and deleted this on accident

alexfikl:
> Sorry, I was deleting previous merged branches and deleted this on accident

No worries! In it goes :rocket:

tuurep:
Nice, love this improvement btw, exporting makes so much more sense now.

jghauser:
> Nice, love this improvement btw, exporting makes so much more sense now.

Yes, me too! Thanks a lot :heart:

Issue #901, opened 2024-09-01
Tab completions (`click`) performance enhancement


tuurep:
Discussed in https://github.com/papis/papis/discussions/898

We think the tab completions (provided by [click](https://click.palletsprojects.com/en/8.1.x/shell-completion/)) are noticeably slower than in other projects (for example git), and thus feels quite bad to use

Pointers by @alexfikl:

> As far as I can tell, this is slow for two reasons:
> 
> 1. We call `_PAPIS_COMPLETE=zsh_complete papis` on every <TAB> to get the completions in real time from the command itself. This is obviously slower than hardcoding the options in some way.
> 
> 2. Just calling `papis` to e.g. display the help is already quite slow. Last time I checked, this was mostly slow because we dynamically load commands (either from scripts in `~/.config/papis/scripts` or from entrypoints).
> 
> 
> I'm not quite sure how to make this noticeably faster though. It might need some more serious profiling and maybe moving away from the click generated completions.

alexfikl:
For what it's worth, `git` has an 8000 line completion file for zsh with all the commands and flags hardcoded in there. It also doesn't seem to complete third party commands, e.g. I have `git-latexdiff` installed from `texlive` and it's not picked up at all.

A first step for this might be just to get some hard numbers on how slow it is. No idea how to benchmark that though..

tuurep:
> noticeably slower than in other projects (for example git)

Actually when I switched from Bash to Zsh (not too long ago), I found the completions in git also slow in some cases, which made me find this:

https://felipec.wordpress.com/2013/07/31/how-i-fixed-git-zsh-completion/ \
https://github.com/felipec/git-completion

I was referring to the Bash completions in the original discussion, which are apparently much better on performance

Thought that may be of interest if a little offtopic

tuurep:
Probably the most relevant issue from Click repo:

https://github.com/pallets/click/issues/2126

Closed in 2021

PR #1039, opened 2025-07-08
Library tab completion (click)

### LINKED ISSUE #902: Tab completions: allow completing library names
Discussed in https://github.com/papis/papis/discussions/898

I've got some long library names where tab completion could be very useful, for example:

```
$ papis -l sys<Tab>
$ papis -l systematic-literature-review
```

When using git, overwhelmingly the most useful thing to tab complete are long branch names, and similarly in papis it would be long library names.

Pointers by @alexfikl:

> This would be very cool! I'm also not sure how to do that nicely. I assume this has to be added as a `click.Choice` to complete automagically, but that might create some dependency cycle in the code..

> This may also do the trick for completing library names:
https://click.palletsprojects.com/en/latest/shell-completion/#custom-type-completion

> Library completion ... should probably go here:
> https://github.com/papis/papis/blob/95e0b5456afef60ee4f54c4aaec624244093c1e0/papis/commands/default.py#L139-L142
>        
> Using the click custom type completion framework. It could work by just making that a `click.Choice(papis.config.get_libs())`, but should be cleaner to use the recommended way.

Related in case we want to get rid of `click`: #901

tuurep:
Closes #902

Ref: https://github.com/papis/papis/discussions/898#discussioncomment-10513530

Considers also "unnamed libraries", i.e. giving a directory path to a library that's not in Papis config

Tested in `zsh` on my part


papis/commands/default.py
```diff
@@ -121,6 +122,30 @@ def _on_finish() -> None:
...
+        # Named libraries from Papis config
+        for lib in papis.config.get_libs():
+            if lib.startswith(incomplete):
+                completions.append(CompletionItem(lib))
+
+        # Unnamed libraries (paths)
+        # Only shown if prefix exists
+        if incomplete:
+            paths = click.Path(exists=True, file_okay=False)
+            completions += paths.shell_complete(ctx, param, incomplete)
```
tuurep: It looks like the order these go in the final completion list depends on the shell,

For me looks like this when both named and unnamed libs match:

```
 papis -l t
textern/  timer/
thesis
```

Paths first, then named libraries, trailing slashes are probably from a `zsh` setting

tuurep: So, I think it's very well-behaved

If only it wasn't so slow.. :smile:

tuurep:
By the way I always have to press Tab twice on a newly created shell to complete Papis, which is annoying. It's like the first one is ghosted. Any tips on this I'd appreciate, but it's offtopic (nothing to do with PR).

papis/commands/default.py
alexfikl: > If only it wasn't so slow..

How slow are we talking? It's reasonably fast for me, i.e. about as fast as shell completion for other papis flags / commands.

papis/commands/default.py
```diff
@@ -121,6 +122,30 @@ def _on_finish() -> None:
     return _on_finish
 
 
+class LibType(click.ParamType):
```
alexfikl:
```suggestion
class LibraryParamType(click.ParamType):
```
? Also, maybe put this in `papis.cli`?

alexfikl:
> By the way I always have to press Tab twice on a newly created shell to complete Papis, which is annoying. It's like the first one is ghosted. Any tips on this I'd appreciate, but it's offtopic (nothing to do with PR).

I'm afraid I don't really know where that comes from, but I can confirm it happens :( 

However, it only seems to happen on the first one, i.e. going `papis --lib <TAB>` needs two, but when I choose a lib and hit tab again, it only needs one to show the commands. So yeah, no idea :\

papis/commands/default.py
tuurep: Thanks, pushed these changes

papis/commands/default.py
tuurep: It's not any slower than other Papis completions, I was just generally complaining about #901 sorry

tuurep:
> I'm afraid I don't really know where that comes from, but I can confirm it happens :(
>
> However, it only seems to happen on the first one, i.e. going `papis --lib <TAB>` needs two, but when I choose a lib and hit tab again, it only needs one to show the commands. So yeah, no idea :\

Yeah, to be clear does this happen for you like this:
1. open a new terminal
2. `papis <Tab>`
    - nothing happens
3. `papis <Tab>`
    - now completes and completions work on 1 tab for the rest of the session

?

We should probably open an issue on it. First I thought it might have been a `zsh` config issue for me.

jghauser:
I do have exactly the same issue with <tab> needing to be pressed twice (also in zsh) after opening a new shell. I found [this issue](https://github.com/pallets/click/issues/2344) that might be relevant.

alexfikl:
Ignoring the unrelated tab issue, this seems good to go from my end. Anything else you'd want to add?

tuurep:
@jghauser thanks, very helpful. I'll try looking at that more.

@alexfikl yeah, I think we can merge, thanks!

papis/commands/default.py
alexfikl: Ah, yeah, making the completions generally faster seems quite complicated :\ Would need some profiling to see where the time is being spent..

alexfikl:
> @alexfikl yeah, I think we can merge, thanks!

Awesome! In it goes :rocket:

PR #1040, opened 2025-07-08
Update completion scripts generated by click

alexfikl:
Re-ran the click shell completion generator things and noticed that they have a few changes.

cc @jghauser @tuurep Haven't tested if this fixes the two-tab issue, but can't hurt.

### DIFF
--- a/contrib/shell_completion/click/fish/papis.fish
+++ b/contrib/shell_completion/click/fish/papis.fish
@@ -1,9 +1,5 @@
 function _papis_completion;
-    set -l response;
-
-    for value in (env _PAPIS_COMPLETE=fish_complete COMP_WORDS=(commandline -cp) COMP_CWORD=(commandline -t) papis);
-        set response $response $value;
-    end;
+    set -l response (env _PAPIS_COMPLETE=fish_complete COMP_WORDS=(commandline -cp) COMP_CWORD=(commandline -t) papis);
 
     for completion in $response;
         set -l metadata (string split "," $completion);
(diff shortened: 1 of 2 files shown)


jghauser:
It does fix the `<tab>` issue on my computer!

EDIT: actually it's a bit stranger now. Open a new shell, type `papis<tab>` and completions appear. However, if I then `<ctrl-c>`, type `papis<tab>` again, it doesn't work anymore and i need to press `<tab>` again. Still a win in my book :sweat_smile:

alexfikl:
> It does fix the `<tab>` issue on my computer!
> 
> EDIT: actually it's a bit stranger now. Open a new shell, type `papis<tab>` and completions appear. However, if I then `<ctrl-c>`, type `papis<tab>` again, it doesn't work anymore and i need to press `<tab>` again. Still a win in my book 😅

Haha, yeah, I also tried it and it seemed to sometimes not work in a seemingly unpredictable way. Oh well, some improvement :)) Thanks for testing!

tuurep:
Yeah similar findings here, the first tab seems to work now but there was something like this:

```
 papis ex<Tab>
exec     -- Execute a python file in the environment of papis' executable
explore  -- Explore new documents using a variety of resources.
export   -- Export a document from a given library.
 papis expo<Tab>
```

(hand-inserted `po`), nothing happens on first Tab but completes on the second.

tuurep:
As you said, unpredictable. I find it hard to get it to fail in a reproducible manner :grimacing:

Well, the example from above seems to happen 100% of the time (the first time after shell launch).

PR #1045, opened 2025-07-17
Workaround to make the usenix downloader usable in face of cloudlflare

f0rki:
For some reason cloudflare blocks only the access to the bibtex file, but not to the pdf or the html page itself. So while the downloader fails because it doesn't have access to the bibtex file, we can work around this by using the inline bibtex information that usenix provides on the html page they serve. This is not ideal, but seems to work for now.

### DIFF
--- a/papis/downloaders/usenix.py
+++ b/papis/downloaders/usenix.py
@@ -40,19 +40,27 @@ class Downloader(papis.downloaders.Downloader):
         self.logger.debug("Parsed URL: %s.", path_components)
         return path_components[2] + "-" + path_components[4]
 
+    def _ensure_raw_data(self) -> None:
+        if not self._raw_data:
+            resp = self.session.get(self.uri, cookies=self.cookies)
+            self._raw_data = resp.content.decode("utf-8")
+            if not self._raw_data:
+                self.logger.warning("Failed to fetch data from '%s'.", self.uri)
+
     def get_document_url(self) -> Optional[str]:
         import bs4
 
+        # make sure self._raw_data is available
+        self._ensure_raw_data()
         if not self._raw_data:
-            self._raw_data = self.session.get(self.uri).content.decode("utf-8")
-            if not self._raw_data:
-                self.logger.warning("Failed to fetch data from '%s'.", self.uri)
-                return None
+            return None
 
         soup = bs4.BeautifulSoup(self._raw_data, "html.parser")
         extension = (
             self.expected_document_extensions[0]
-            if self.expected_document_extensions else "")
+            if self.expected_document_extensions
+            else ""
+        )
 
         a = list(
             filter(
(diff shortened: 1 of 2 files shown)


f0rki:
This fixes #1044 for now.

papis/downloaders/usenix.py
```diff
@@ -80,26 +89,75 @@ def get_bibtex_url(self) -> Optional[str]:
...
+        """
+        url = self.get_bibtex_url()
+        if not url:
+            return
+        self.logger.info("Downloading BibTeX from '%s'.", url)
+
+        response = self.session.get(url, cookies=self.cookies)
+        self.bibtex_data = response.content.decode().strip()
+        if self.bibtex_data.startswith("<!DOCTYPE html>"):
+            self.logger.debug("downloaded bibtex data: %s", self.bibtex_data)
```
alexfikl:
```suggestion
            self.logger.debug("Downloaded BibTeX data:\n%s", self.bibtex_data)
```

papis/downloaders/usenix.py
```diff
@@ -80,26 +89,75 @@ def get_bibtex_url(self) -> Optional[str]:
...
+
+        # fallback to trying to fetch the bibtex from the _raw_data itself
+        # make sure self._raw_data is available
+        self._ensure_raw_data()
+        if not self._raw_data:
+            return None
+        # setup html parser
+        import bs4
+
+        soup = bs4.BeautifulSoup(self._raw_data, "html.parser")
```
alexfikl, line 146: Does `self._get_soup()` not work here? I think that defaults to an `lxml` parser.

papis/downloaders/usenix.py
```diff
@@ -80,26 +89,75 @@ def get_bibtex_url(self) -> Optional[str]:
...
+        )
+
+        if finds:
+            div = finds[0]
+            text = div.text.replace("<br/>", "\n")
+            # self.logger.debug("got bibtex from html: %s", text)
+
+            self.bibtex_data = text
+        else:
+            self.logger.debug("failed to identify bibtex content in usenix html page!")
```
alexfikl:
```suggestion
            self.logger.debug("Failed to identify BibTeX content in USENIX HTML page!")
```

papis/downloaders/usenix.py
```diff
@@ -80,26 +89,75 @@ def get_bibtex_url(self) -> Optional[str]:
...
+                lambda t: "bibtex-text-entry" in t.get("class", ""),
+                soup.find_all("div"),
+            )
+        )
+
+        if finds:
+            div = finds[0]
+            text = div.text.replace("<br/>", "\n")
+            # self.logger.debug("got bibtex from html: %s", text)
+
```
alexfikl:
```suggestion
```

alexfikl (commented):
Left a couple of nitpicks, but otherwise this looks good to me!

papis/downloaders/usenix.py
f0rki: I probably don't want to fetch the contents again, since `self._raw_data` should already be there, no?

alexfikl: Hm, agreed, but feels like this could be reworked to take advantage of it (`self._get_soup` is also cached), but it's not a pressing issue.

alexfikl:

PR #1048, opened 2025-08-13
fix slugify always replacing hyphens

retroflexivity:
Slugify reserves the hyphen-minus character and always replaces it with the separator, even if the regex pattern excludes it https://github.com/un33k/python-slugify/issues/107. This fixes the problem by doing pre-processing if the hyphen-minus is included in `extra-chars`, as suggested in the issue. It does it for both the hyphen-minus (\u002d) and the hyphen (\u2010).

### DIFF
--- a/papis/paths.py
+++ b/papis/paths.py
@@ -88,14 +88,24 @@ def normalize_path(path: str, *,
     else:
         regex_pattern = fr"[^a-zA-Z0-9.{extra_chars}]+"
 
+    # fix slugify forcefully replacing hyphens
+    if "-" in extra_chars:
+        placeholder = "slugifyhyphenfixer"
+        path = path.replace("\u2010", placeholder).replace("-", placeholder)
+
+        def replace_hyphen(s: str) -> str:
+            return s.replace(placeholder, "-")
+    else:
+        def replace_hyphen(s: str) -> str: return s
+
     import slugify
 
-    return str(slugify.slugify(
+    return replace_hyphen(str(slugify.slugify(
         path,
         word_boundary=True,
         separator=separator,
         regex_pattern=regex_pattern,
-        lowercase=lowercase))
+        lowercase=lowercase)))
 
 
 def is_relative_to(path: PathLike, other: PathLike) -> bool:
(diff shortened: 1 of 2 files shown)


alexfikl:
This looks good, but I have a few questions
1. What's the use case for this? I personally haven't customized the path normalizing at all, so I don't have a good overview of the edge cases.
2. Can you add a little test case with an example? There should be some small tests for `extra-chars` in `tests/test_paths.py`.

papis/paths.py
```diff
@@ -88,14 +88,24 @@ def normalize_path(path: str, *,
     else:
         regex_pattern = fr"[^a-zA-Z0-9.{extra_chars}]+"
 
+    # fix slugify forcefully replacing hyphens
```
alexfikl, line 93: Maybe link to the https://github.com/un33k/python-slugify/issues/107 issue in here as well? So that we don't forget that this is more or less the recommended workaround.

papis/paths.py
```diff
@@ -88,14 +88,24 @@ def normalize_path(path: str, *,
...
+    if "-" in extra_chars:
+        placeholder = "slugifyhyphenfixer"
```
alexfikl: Make this a global variable `_SLUGIFY_HYPHEN_PLACEHOLDER = "..."`?

retroflexivity:
> 1. What's the use case for this? I personally haven't customized the path normalizing at all, so I don't have a good overview of the edge cases.

i personally prefer fancy filenames with spaces and stuff, so i need to modify my separator. when there is a word with a hyphen, i would like it to be preserved.

> 2. Can you add a little test case with an example? There should be some small tests for `extra-chars` in `tests/test_paths.py`.

done

papis/paths.py
retroflexivity: don't think anyone would need to modify this, but sure. though i struggle to understand where vars are stored in the project

alexfikl: > don't think anyone would need to modify this, but sure.

The leading underscore should discourage people from accessing it. My thought was just that it's a global variable, so we might as well make it look like one.

> though i struggle to understand where vars are stored in the project

Just meant to put it on top of the function here (there's no special place).

alexfikl:
> i personally prefer fancy filenames with spaces and stuff, so i need to modify my separator. when there is a word with a hyphen, i would like it to be preserved.

Fair enough. 

Thanks for adding the tests! Once the little linting error is fixed, this looks good to me.

alexfikl:

PR #1049, opened 2025-08-13
Add doctor check to fix issue keys for BibLaTeX

### LINKED ISSUE #1046: Entry renaming while exporting to biblatex
There is an unfortunate usage of the entry name 'issue' in biblatex. Per [4.9.2.10 of its documenation ](https://mirrors.ircam.fr/pub/CTAN/macros/latex/contrib/biblatex/doc/biblatex.pdf), it is the entry 'number' which refers to the issue of the journal. The entry 'issue' gives unexpected output for journal article entries, which looks like the date instead of the issue. Also see https://tex.stackexchange.com/a/424592 and https://tex.stackexchange.com/a/423985. 

This means the exporter needs to change the entry name from 'issue' to 'number' when the output format is biblatex.

alexfikl:
This adds a doctor check for BibTeX keys that look wrong (currently just "issue"). The autofix will rename "issue" to "number" if it looks like that's a good idea.

Fixes #1046.

### DIFF
--- a/tests/test_doctor.py
+++ b/tests/test_doctor.py
@@ -393,3 +393,27 @@ def test_html_tags_check_jats(tmp_config: TemporaryConfiguration,
     error.fix_action()
     assert "\n".join(doc["abstract"].split()) == "\n".join(expected.strip().split())
     assert doc["abstract"] == expected.strip()
+
+
+def test_biblatex_issue_to_number(tmp_config: TemporaryConfiguration) -> None:
+    from papis.commands.doctor import biblatex_key_convert_check
+
+    doc = papis.document.from_data({
+        "author": "Sanger, F. and Nicklen, S. and Coulson, A. R.",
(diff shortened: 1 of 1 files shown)


papis/commands/doctor.py
```diff
@@ -633,12 +635,85 @@ def biblatex_required_keys_check(doc: papis.document.Document) -> List[Error]:
             if not any(key in doc or aliases.get(key) in doc for key in keys)]
 
 
+BIBLATEX_KEY_RENAME_CHECK_NAME = "biblatex-key-rename"
```
alexfikl: I'm really not a fan of this check name. Any suggestions?

I wanted something vaguely general, since this seems like it could be used on other things too, but maybe it's better to make it simple and focused..

alexfikl: Went with `biblatex-key-convert`.. sounds a bit better.

PR #1053, opened 2025-08-14
Add doctor check for basic string cleaning

### LINKED ISSUE #874: `papis doctor` and linebreaks in titles (+ other improvements)

alexfikl:
This tries to implement some basic string cleaning from #874. It seems to work well enough, but there are probably a lot of edge cases to consider..

@jghauser It would be great if you had a chance to test this stuff as well :grin:

Fixes #874.


jghauser:
Will do, but might take me a couple of days!

alexfikl:
> Will do, but might take me a couple of days!

That's completely fine. Thank you!

jghauser:
I have finally gotten around playing with this. Looks very neat and useful and found and fixed a number (96!) of issues in my library! Some remarks:

- Why don't we just check all keys which are defined to be strings?
- 'Key 'author' is incorrectly spaced' seems to show for both spacing issues and missing dots. It also took me a while to figure out what the spacing issue is. Maybe we can say: `Keys 'author' or 'author_list' contains initials that aren't followed by a dot or aren't separated by whitespace`
- I'm not sure the suggestions are helpful as they pretty much just say 'edit the info file' :D. Given that the fixers sort these issues out automatically, why don't we just recommend running the fixer?

alexfikl:
> I have finally gotten around playing with this. Looks very neat and useful and found and fixed a number (96!) of issues in my library!

Awesome! Did you check if it fixed them all "correctly"? The one with adding dots and spaces to authors was quite annoying to get "right" and I don't trust it very much :\ If you have any cases that don't work, it would be great to add them as unit tests!

> * Why don't we just check all keys which are defined to be strings?

Yeah, that sounds like a good idea!
 
> * 'Key 'author' is incorrectly spaced' seems to show for both spacing issues and missing dots. It also took me a while to figure out what the spacing issue is. Maybe we can say: `Keys 'author' or 'author_list' contains initials that aren't followed by a dot or aren't separated by whitespace`

That sounds better! Maybe it's worth splitting the two fixes? If nothing else, it would simplify that horrible regex.

> * I'm not sure the suggestions are helpful as they pretty much just say 'edit the info file' :D. Given that the fixers sort these issues out automatically, why don't we just recommend running the fixer?

Agreed. Those are mostly copy pasted on every error message at this point. I'll go through them and update the wording on those that have fixers.

jghauser:
I have to admit, I only checked a few fixes and then just let it run :sweat_smile:. Went over the changes now again quite carefully, and it seems your dots-and-spaces regex magic works very well. Couldn't find a single case where it went wrong!

jghauser:
> That sounds better! Maybe it's worth splitting the two fixes? If nothing else, it would simplify that horrible regex.

Yeah I think that might be good idea!

alexfikl:
> Yeah I think that might be good idea!

Done! I think I addressed all the comments you had (in particular, [33c4952](https://github.com/papis/papis/pull/1053/commits/33c495253a6a1b6303c9b882a6693d5001d959cd) updates the `suggested_cmd` thing). Let me know if there's anything else!

jghauser:
Looks good to me! Very nice that suggestions are now more useful overall :)

alexfikl:
> Looks good to me! Very nice that suggestions are now more useful overall :)

Great! Thanks for the initial push and help with testing this!

In it goes :rocket:

Issue #1060, opened 2025-08-18
Errors importing formatters


jghauser:
I'm getting errors for most (all?) papis commands with newest main :D.

On my nixos machine:
```
❯ papis config --help                                 
Traceback (most recent call last):
  File "/nix/store/3v7a6jlrvnrscy9fvjqrcbm975w7hgz9-python3.13-papis-8370b6a/bin/.papis-wrapped", line 9, in <module>
    sys.exit(run())
             ~~~^^
  File "/nix/store/zbzpxnsm8j0rqakmdvsrf36zh28xick2-python3.13-click-8.1.8/lib/python3.13/site-packages/click/core.py", line 1161, in __call__
    return self.main(*args, **kwargs)
           ~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/nix/store/zbzpxnsm8j0rqakmdvsrf36zh28xick2-python3.13-click-8.1.8/lib/python3.13/site-packages/click/core.py", line 1082, in main
    rv = self.invoke(ctx)
  File "/nix/store/zbzpxnsm8j0rqakmdvsrf36zh28xick2-python3.13-click-8.1.8/lib/python3.13/site-packages/click/core.py", line 1694, in invoke
    super().invoke(ctx)
    ~~~~~~~~~~~~~~^^^^^
  File "/nix/store/zbzpxnsm8j0rqakmdvsrf36zh28xick2-python3.13-click-8.1.8/lib/python3.13/site-packages/click/core.py", line 1443, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/zbzpxnsm8j0rqakmdvsrf36zh28xick2-python3.13-click-8.1.8/lib/python3.13/site-packages/click/core.py", line 788, in invoke
    return __callback(*args, **kwargs)
  File "/nix/store/zbzpxnsm8j0rqakmdvsrf36zh28xick2-python3.13-click-8.1.8/lib/python3.13/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/nix/store/3v7a6jlrvnrscy9fvjqrcbm975w7hgz9-python3.13-papis-8370b6a/lib/python3.13/site-packages/papis/commands/default.py", line 220, in run
    for libname in papis.config.get_libs()
                   ~~~~~~~~~~~~~~~~~~~~~^^
[...]

PR #1061, opened 2025-08-18
Export formatters from papis.format

alexfikl:
For backwards compatibility mostly.

cc @jghauser This should fix at least some of the problems. Let me know if there's anything else!
xref: #1060


jghauser:
Everything works again on my end after changing the imports. So I think we're good here!

PR #1064, opened 2025-08-20
Refactor pickers to separate folder

alexfikl:
Similar to #1062. Main changes are
* Moved the `fzf` and `papis` pickers to a separate folder. Also added the alias `prompt_toolkit` for the `papis` picker, in case another one comes along (hello `textual`!) that would be better as a default.
* Moved most of the things in `papis.tui` into `papis.tui.picker` to emphasize that they're just there for the default picker TUI.


alexfikl:
cc @supersambo You're the only other person I know who has a custom picker :grin: I tested it and all this refactoring doesn't seem to break anything, but another pair of eyes are very welcome. Feel free to yell if this breaks some compatibility!

supersambo:
yes seems to be working fine! Thanks for testing it!

Issue #1063, opened 2025-08-20
update with --append: support other fields


avonmoll:
This is somewhat of a feature request and a continuation of #692:

I tried to run

```
papis update --append project 'my new paper'
```

and got the error:

```
[ERROR] commands.update: We cannot append to key 'project', because we do not know the intended type. Please use `papis update --set` instead.
[ERROR] commands.update: Aborting operation. No documents have been changed.
```

I assume the reason is that `project` is not a builtin field (like `tags`) and thus papis does not know if `project` may be a string or list. Is it possible to get around this somehow? 

Some ideas:

1. if field exists, try to determine whether it is a string or list, append appropriately
2. if field doesn't exist, default to one of the types (e.g. start the field as a list)
3. add a flag to the `update` CLI to specify the type or default type

Alternatively, if I really wanted to use `update --append` I could exclusively use `tags` field. However, my current workflow is based on using `tags` for research topic areas and `projects` for filtering references for a particular manuscript I'm writing.

alexfikl:
> I assume the reason is that `project` is not a builtin field (like `tags`) and thus papis does not know if `project` may be a string or list. Is it possible to get around this somehow?

Hm, apparently this could use some better documentation! The `--append` flag depends on the [`doctor-key-type-keys`](https://papis.readthedocs.io/en/latest/configuration.html#confval-doctor-key-type-keys) setting to have a guess at how to append things to a given key. You can check the default ones using
```
papis config doctor-key-type-keys
```
You can add more of them using the [`doctor-key-type-keys-extend`](https://papis.readthedocs.io/en/latest/configuration.html#confval-doctor-key-type-keys-extend) setting like (only builtin types are supported)
```
doctor-key-type-keys-extend = ["project:list"]
```

Haven't tried it, but hopefully that works. Let us know! :grin:

avonmoll:
That worked! Thank you so much! I must admit that I did not spend much time reading docs pertaining to `doctor`.

PR #1065, opened 2025-08-20
Improve documentation for papis update --append

alexfikl:
Adds a few sentences that `papis update --append` needs the `doctor-key-type-keys` configuration option to work for new keys.

cc @avonmoll Hope this helps!
cc @jghauser I couldn't find any other places where this could be mentioned, but let me know if anything else comes to mind :grin:

xref: #1063

### DIFF
--- a/papis/commands/update.py
+++ b/papis/commands/update.py
@@ -80,6 +80,12 @@ Examples
     doesn't yet exist, it will be created. All duplicate items will be removed
     from the list.
 
+    As you might have guessed, the ``--append`` flag needs to know the type of
+    the key it is appending to. It does this by looking at the
+    :confval:`doctor-key-type-keys` (and :confval:`doctor-key-type-keys-extend`)
+    configuration options. If the key you are appending to is not in that list,
+    the command will fail.
+
 - To remove an item from a list, use ``--remove``:
 
     .. code:: sh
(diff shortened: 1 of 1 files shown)


papis/commands/update.py
```diff
@@ -80,6 +80,12 @@
     doesn't yet exist, it will be created. All duplicate items will be removed
     from the list.
 
+    As you might have guessed, the ``--append`` flag needs to know the type of
+    the key it is appending to. It does this by looking at the
+    :confval:`doctor-key-type-keys` (and :confval:`doctor-key-type-keys-extend`)
+    configuration options. If the key you are appending to is not in that list,
+    it will not be appended to.
```
jghauser: I would maybe replace this line with `the command will fail.` I think it makes it a bit clearer what exactly will happen.

jghauser:
Looks good to me! In a way, it would be nice if `--help` showed this information, but I think it would make the help text too long. And with the error message, the relevant info should be conveyed only a moment later :)

papis/commands/update.py
alexfikl: Fixed. Thanks!

alexfikl:
> Looks good to me! In a way, it would be nice if `--help` showed this information, but I think it would make the help text too long. And with the error message, the relevant info should be conveyed only a moment later :)

Thanks for taking a look!

Hm, we could put some more information into the main help text of the command? e.g. `papis update --help` just shows "Update document metadata." at the moment (below "Usage" and over "Options"). I'm not sure what's a good limit there too though.. especially for more complex commands like this.

jghauser:
That would be very nice, although this is also the text that shows when doing `papis` and hitting <tab>, and that text should be short. No idea if the two could be disentangled somehow.

Alternatively, having man pages autogenerated would make the docs more accessible.

alexfikl:
> Alternatively, having man pages autogenerated would make the docs more accessible.

Sphinx should be able to generate those, but I'm not sure how to install them. I thought the code in `tools/make-wheels.sh` added them to the wheels, but it doesn't seem like they're there. Even if they were though, I'm not sure that would get picked up when doing `man papis`..

PR #1068, opened 2025-09-03
Always follow ref-format. Fixes #1067

### LINKED ISSUE #1067: ref-format should take precedence over `--from bibtex`
I have in my `.config/papis/config`

```
[settings]
ref-format = '{doc[author_list][0][family]!l}{doc[year]}'
```

When I run `papis add --from bibtex test.bib`, the `ref` field is set to the one in the test.bib file, and not re-generated according to ref-format. `ref-format` works correctly when adding from doi, isbn, etc. 

So for example if `test.bib` is

```
@article{lewis1993many,
  title={Many, but almost one},
  author={Lewis, David},
  journal={Ontology, causality and mind},
  pages={23--42},
  year={1993},
  publisher={Cambridge University Press Cambridge}
}
```
then the generated document has `lewis1993many` as the `ref`, instead of `lewis1993`, which is what I would expect given my `ref-format`.

There should at the very least be a setting (`ref-format-rewrite` or something) to toggle this behavior.

I would argue that the correct default behavior is to follow ref-format no matter what, and the flag should switch back to the current behavior. In my opinion, if I have gone to the length of setting `ref-format`, it's because I want _all_ my references to follow the same format, and just because I imported from bibtex should not change that. 

I cannot really imagine why someone would _want_ the current behavior (maybe to have the bibtex export correspond to the import, given no other changes?), but that should be a flag.

pinusc:
Ignore ref provided by import data, and always follow ref-format when adding a new reference.

I did some limited testing and this seems to work. Following feedback in #1067, I did not bother adding a config flag to restore the previous behavior for now, but I could work on that if necessary.

### DIFF
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -264,13 +264,12 @@ def run(paths: list[str],
     # reference building
     # NOTE: this needs to go before any papis.format calls, so that those can
     # potentially use the 'ref' key in the format patterns.
-    if "ref" not in data:
-        from papis.bibtex import create_reference
+    from papis.bibtex import create_reference
 
-        new_ref = create_reference(data)
-        if new_ref:
-            logger.info("Created reference '%s'.", new_ref)
-            tmp_document["ref"] = new_ref
+    new_ref = create_reference(data, True)
+    if new_ref:
+        logger.info("Created reference '%s'.", new_ref)
+        tmp_document["ref"] = new_ref
 
     if auto_doctor:
         logger.info("Running doctor auto-fixers on document: '%s'.",
--- a/tests/commands/test_add.py
+++ b/tests/commands/test_add.py
@@ -97,6 +97,7 @@ def test_add_auto_doctor_run(tmp_library: TemporaryLibrary) -> None:
 
     # add document with auto-doctor on
     papis.config.set("doctor-default-checks", ["keys-missing", "key-type", "refs"])
+
     run(paths, data=data, auto_doctor=True)
 
     # check that all the broken fields are fixed
@@ -105,7 +106,7 @@ def test_add_auto_doctor_run(tmp_library: TemporaryLibrary) -> None:
 
     assert doc["author_list"] == [{"given": "Werner", "family": "Kutzelnigg"}]
     assert isinstance(doc["year"], int) and doc["year"] == 2009
-    assert doc["ref"] == "2FJT2E3A"
+    assert doc["ref"] == "How_many_body_p_Kutzel_2009"
 
 
 def test_add_set_cli(tmp_library: TemporaryLibrary) -> None:


jghauser:
Thanks for the PR, looks good to me! What do you think @alexfikl, should we have an option to preserve the previous behaviour?

EDIT: Maybe when collaborating, people might get bibtex entries from others and want to keep refs consistent? Just trying to come up with use cases...

pinusc:
So, even without a flag, there is a workaround to allow for this behavior:

```
formatter=jinja2
ref-format = '{{ doc["ref"] if doc["ref"] else doc["author_list"][0]["family"].lower() ~ doc["year"]}}'
```
would be equivalent to "keep `ref` if provided, else use"
```
ref-format = '{doc[author_list][0][family]!l}{doc[year]}'
```

Jinja is necessary because AFAIK the PythonFormatter method does not allow conditionals, and if we have `ref-format = '{doc[ref]}'` but with an empty `ref` it just does not create one (which is expected). 

This is somewhat tricky, so maybe we could add an example to the documentation? Even just mentioning in `ref-format` documentation that the previously-given value for `ref` is available for use.

alexfikl:
Thanks for working on this! I agree with the general idea, but not the implementation here. I would recommend:
1. Actually fixing the BibTeX importer to not save the reference key (i.e. just deleting it in the `fetch_data` method, if it's there). Most of the importers don't do this by default, so it's good to be consistent.
2. Add few lines like in `papis/commands/update.py`:

https://github.com/papis/papis/blob/6ac9b4383241b07fbb12ee04ed497bc13f5d2ff6/papis/commands/update.py#L557-L564

somewhere around here (after the importer data is merged):

https://github.com/papis/papis/blob/6ac9b4383241b07fbb12ee04ed497bc13f5d2ff6/papis/commands/add.py#L540-L542

Just straight up ignoring the `ref` key like in this PR doesn't seem like a good idea, since it would also overwrite whatever the user set manually using `--set ref <MY_REF>`.

I think this should also fix your issue. What do you think?

alexfikl:
> EDIT: Maybe when collaborating, people might get bibtex entries from others and want to keep refs consistent? Just trying to come up with use cases...

Hm, I'm not sure we have a good workflow around something like that.. although it would be lovely, since I also hit that annoying issue quite a few times :(. 

My thought was that that would be a better fit for the `papis bibtex` command? That could be a bit more careful with reading/writing existing `refs`, while `papis add` is more of a "chug this into my library with my preferred formats" sort of command.

jghauser:
> My thought was that that would be a better fit for the `papis bibtex` command? That could be a bit more careful with reading/writing existing `refs`, while `papis add` is more of a "chug this into my library with my preferred formats" sort of command.

Yes, that makes sense to me!

alexfikl:
@pinusc Still interested in getting this fix in? :grin:

PR #1073, opened 2025-09-13
Refactor importers to separate folder

alexfikl:
This mostly moves all the importers to `papis/importer` and adds a bunch of new helper functions (moved from `papis.utils`).

xref: #1064


alexfikl:
Hm, those CodeQL errors are there in main too. I guess they're reappearing here because the code in question was moved around a bit.

Hopefully nothing was broken from moving all these imports around, so in it goes :worried:

PR #1077, opened 2025-09-21
Fix issues with click 8.3.0

alexfikl:
It seems like click 8.3.0 started having an issue with the duplicate `--prefix` flag in `papis git`. The `papis git` command is a bit weird because it's constructed as a deep copy of the `papis run` command, so maybe we were doing something that `click` wasn't expecting there..

This seems to fix it for me on Python 3.13.


alexfikl:
Hm, still not working on Python 3.10..

EDIT: Went ahead and just made it a standalone command. There's a bit of copy pasting, but many commands are a bit copy-pastey, so worth avoiding the trouble with `click`.

PR #1079, opened 2025-10-08
Reduce global imports

alexfikl:
This aggressively reduces global imports:
* In most files only `papis.config` and `papis.logging` are imported.
* Commands also use `papis.cli`
* Occassionally, we also need to import like `get_available_importers` to show them on the command line. It would be nice to make that lazy as well, since loading importers from plugins is a big chunk of startup time.

The main reason for this was that things like
```python
import papis.config

print(papis.database.get())
```
don't actually show up as any sort of error (with `ruff` or `mypy`). This tries to make most of the imports explicit `from ... import ...` and local because they're (a) generally faster and (b) all tools complain if they're unused / wrong.


alexfikl:
I think `pyright` / `basedpyright` catches the incorrect import error, so we might want to look into switching to that in the future. Although with some promising Rust type checkers on the horizon, this might not be the best time..

alexfikl:
I'll go ahead and merge this, since it just moves imports around. We can follow up with some fixes if this messed up some implicit import order :\

PR #1082, opened 2025-10-12
Start testing on py3.14

alexfikl:
Don't really expect this to work, but just to see where we are.

### DIFF
--- a/doc/source/testing.rst
+++ b/doc/source/testing.rst
@@ -4,8 +4,8 @@ Testing
 Papis uses :mod:`pytest` for its testing infrastructure and makes use of some of
 its more advanced features (such as
 `fixtures <https://docs.pytest.org/en/latest/explanation/fixtures.html>`__) to
-set everything up. The command-line interface is based on :mod:`click` and is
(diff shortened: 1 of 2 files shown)


alexfikl:
Hm, amusingly only `ubuntu-latest` fails on Python 3.14. Not quite sure what that is about right now.. the only thing that comes to mind is some changes related to `multiprocessing` that could be messing things up. Seems easy enough to fix though :grin:

alexfikl:
Set `PAPIS_NP=0` to disable multiprocessing and that seems to pass all the tests, so the assumption seems correct!

[`7703f63` (#1082)](https://github.com/papis/papis/pull/1082/commits/7703f63df835d43dbad740befcf7d9dce7a0044c)

alexfikl:
Figured out what the underlying issues was! The new default forkserver backend for `multiprocessing` on Python 3.14 uses more pickling and did not appreciate how `DocMatcher` works.

This PR adds a little `DocumentMatcher` class that has the query arguments as instance attributes instead of class attributes. That seems to work nicely!

PR #1091, opened 2025-10-27
updated nix flake, s.t., it works with current nixpkgs

### LINKED ISSUE #1088: Update flake.lock
@jghauser I tried doing a `nix flake update` in #1082 and it failed pretty much instantly with some weird errors: https://github.com/papis/papis/actions/runs/18801035098/job/53648586170?pr=1082

Can you take a look when you get a chance? Also, the `DeterminateSystems` actions are showing some deprecation warnings and talking about not using "upstream nix" anymore from January? Not quite sure what that is about..

f0rki:
fixes #1088

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -20,11 +20,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1742069588,
-        "narHash": "sha256-C7jVfohcGzdZRF6DO+ybyG/sqpo1h6bZi9T56sxLy+k=",
+        "lastModified": 1761373498,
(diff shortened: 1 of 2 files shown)


alexfikl:
Looks good to me! :rocket:

Anything to worry about with these messages from DeterminateSystems?
```
Timeline
* November 10: we are changing the installer to default to Determinate Nix.
  You can add the `--prefer-upstream-nix` flag now to keep upstream Nix as the default.
* January 1: we are removing support for installing upstream Nix.
  The `--prefer-upstream-nix` flag will not have an effect any longer.
```

f0rki:
@alexfikl I am not entirely sure about the implications of this. I haven't looked into determinate nix in detail, but I think it is probably best to use upstream nix instead. I will open a separate PR that uses a different nix install action so that we get upstream nix.

PR #1092, opened 2025-10-27
ci: clean up nix-related ci workflow

f0rki:
* remove determinate nix installer and switch to upstream nix using the nixbuild/nix-quick-install action
* introduce nix store caching using nix-community/cache-nix-action
* fail workflow if nixpkgs is outdated using flake-checker-action

PR #1093, opened 2025-10-28
updated and reworked nix CI actions 

f0rki:
This replaces the PR #1092

* remove determinate nix installer and switch to upstream nix using the nixbuild/nix-quick-install action
* introduce nix store caching using nix-community/cache-nix-action
* removed the flake-checker-action
* new: add a separate workflow to open a PR once a month with an updated nix flake.

Note: this also reverted to a previous version of nixpkgs in the flake.lock, s.t., the workflow will open a PR for testing.
**After the merge**, trigger the new workflow manually, which opens a PR to move flake.lock back to latest nixpkgs.

### DIFF
--- a/.github/actions/nix-setup/action.yml
+++ b/.github/actions/nix-setup/action.yml
@@ -1,0 +1,25 @@
+name: setup nix and nix store cache
+runs:
+  using: "composite"
+  steps:
+    - name: Install nix
+      uses: nixbuild/nix-quick-install-action@v34
+    - name: Restore and save nix store from cache
+      uses: nix-community/cache-nix-action@v6.1.3
+      with:
+        # save a new cache every time the ci file or the flake changes
+        primary-key: nix-cache-${{ hashFiles('.github/workflows/nix-flake.yaml', 'flake.nix', 'flake.lock') }}
+        # if no hit, restore current versions of individual caches
+        restore-prefixes-first-match: nix-cache-
+        # purge all versions of the cache
+        purge: true
+        purge-prefixes: nix-cache-
+        # created more than 0 seconds ago relative to the start of the `Post Restore` phase
+        purge-created: 0
+        # except the version with the `primary-key`, if it exists
+        purge-primary-key: never
+        # and collect garbage in the Nix store until it reaches this size in bytes
+        gc-max-store-size: 0M
+    - name: check installed nix version
+      run: which nix; nix build --version
+      shell: bash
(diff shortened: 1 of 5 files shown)


f0rki:
@alexfikl what do you think about a CI action that periodically updates the nix flake.lock?

alexfikl:
> @alexfikl what do you think about a CI action that periodically updates the nix flake.lock?

Yeah, sure! I think we already have the necessary permissions enabled because of the PyPI publishing action, so that should just work.

.github/actions/nix-setup/action.yml
```diff
@@ -0,0 +1,25 @@
+name: setup nix and nix store cache
+runs:
+  using: "composite"
+  steps:
+    - name: Install nix
+      uses: nixbuild/nix-quick-install-action@v34
```
alexfikl, line 6: Any idea if these will still get picked up by dependabot? Our config says `directory: "/"`, so I'm hoping it's not an issue 🤷

f0rki: Not sure, but these are standard paths, so I would hope so?

f0rki:
FYI:
issue regarding depandabot and nix flake support: https://github.com/dependabot/dependabot-core/issues/7340

alexfikl:
> FYI: issue regarding depandabot and nix flake support: [dependabot/dependabot-core#7340](https://github.com/dependabot/dependabot-core/issues/7340)

That seems to be about updating the flake itself? We're not even using dependabot to update our Python dependencies, so that seems like a way away, haha.

alexfikl:
@f0rki Is this good to go? :grin:

f0rki:
@alexfikl yes it is!

alexfikl:
> @alexfikl yes it is!

Awesome! Thanks for working on it. I may steal the "automatic update" parts of it for other things as well :grin:

In it goes :rocket:

f0rki:
awesome :) @alexfikl reminder to manually trigger the workflow here to check whether everything works and see the PR opening in action: https://github.com/papis/papis/actions/workflows/nix-flake-update-pr.yml

alexfikl:
> awesome :) @alexfikl reminder to manually trigger the workflow here to check whether everything works and see the PR opening in action: [papis/papis/actions/workflows/nix-flake-update-pr.yml](https://github.com/papis/papis/actions/workflows/nix-flake-update-pr.yml)

Right, right! Running it now :D

PR #1099, opened 2025-11-14
Add an openreview importer

manuelhaussmann:
An importer for [openreview.net](https://openreview.net/). 

Usage as with arXiv etc: `papis add --from openreview "https://openreview.net/forum?id=KurYdcCbjv"`

How to parse the notes could probably be more elegant than what I'm currently doing. 

Currently, it only imports the main pdf of the article. This could be extended to include supplementary material, but I'm not sure whether potentially huge zip files should be added without further confirmation. (Maybe via an extra flag, such as `--with-supp`?)


alexfikl:
Awesome! Very welcome :grin: I'll take a closer look later, but some quick notes:
* Add the new file to `papis/importer` instead, so that it's there with all the other ones.
* This seems to add a new dependency on `openreview-py`. Could that be made optional? I'm not sure we can just have the plugin not register, but at least it could always return `False` in `match`.
* Some tests would be very welcome! :grin:

> Currently, it only imports the main pdf of the article.

I think that's fine. None of the other importers / downloaders look at supplementary material either, so it'll be a bigger effort to try and get something like that in uniformly.

manuelhaussmann:
> Add the new file to papis/importer instead, so that it's there with all the other ones.  

Fixed.

> This seems to add a new dependency on openreview-py

Yes, it currently relies on openreview's api to get the metadata and files, but it probably makes sense to make this an optional dependency for those who want to use this importer.
An alternative would be to get everything from the website itself, I'll have a look at what the easiest way could be to get everything from the raw source later today.

If I understand the distinction correctly, this second approach would belong to the `Downloader` family instead of the `Importer` one, right?

In either case, should the tests be in `papis/tests` or in `papis/tests/downloaders`? (In the second case probably yes, but in the first?)


Not really related to this pull request, but would a second one on pmlr/jmlr (https://proceedings.mlr.press/) be of interest for papis? With the arxiv/openreview/pmlr combo I have covered almost everything I daily do manually with papis, but I'm not sure it makes sense to have importers/downloaders for too many sources in the main repo, especially given that the papis user base for pmlr/jmlr is probably just a single person (me ;) ).

alexfikl:
> Yes, it currently relies on openreview's api to get the metadata and files, but it probably makes sense to make this an optional dependency for those who want to use this importer.

It's fine to make it a proper optional dependency. There's no need to scrape the website if there is a well maintained library for it :grin: I was mostly worried because `openreview-py` doesn't seem to be packaged in Arch Linux at all, so it would be a bit of a pain to get, haha.

> In either case, should the tests be in papis/tests or in papis/tests/downloaders? (In the second case probably yes, but in the first?)

Maybe just in `papis/tests`? The tests aren't particularly well structured at the moment, so no point in overthinking it.

> Not really related to this pull request, but would a second one on pmlr/jmlr ([proceedings.mlr.press](https://proceedings.mlr.press/)) be of interest for papis?

Yeah, definitely! The downloaders don't work very well these days because cloudflare blocks more and more scraping that doesn't go through a full browser, so adding support for more things through some API is very welcome.

If it's also just an optional dependency (or just some raw API calls), I wouldn't worry about it too much. I'm sure there are other people who'll find it useful :grin:

manuelhaussmann:
Yes, setting up and maintaining an aur package for  `openreview-py` might be a bit much. I can make it optional so that it's up to the user to figure out a way to add it if they want to use the importer. 

For people that are fine with not managing papis via pacman but use a separate package manager for python like uv, sth simple like `uv tool install papis --with openreview-py` should already be sufficient to get a working version. 

>  An alternative would be to get everything from the website itself, I'll have a look at what the easiest way could be to get everything from the raw source later today.

Turns out that with later today I meant sometime towards the end of December... I probably won't be able to get to it earlier after all.

alexfikl:
> Turns out that with later today I meant sometime towards the end of December... I probably won't be able to get to it earlier after all.

No worries!

alexfikl:
@manuelhaussmann Do you have any plans to get back to this and finish it up? It seemed to be pretty close.

manuelhaussmann:
Yes the plan is still there... the last months have been a sprint from one deadline to the next. But every time I manually add a paper I remember again that I should finalize the implementation.  

If things go according to plan, I'll finally have some time to finish things in June. (And then think about the next one, as a pmlr one (https://proceedings.mlr.press/) would also be very nice.)

alexfikl:
> Yes the plan is still there... the last months have been a sprint from one deadline to the next.

No worries at all! If you don't find the time, let us know. You already did most of the hard work, so I don't mind finishing it up so it doesn't get lost.

manuelhaussmann:
I still have it on my list, but keep being delayed by the real-world. If you want to take it over feel free to do so, otherwise I hopefully find some time soonish (in approx ~1month).

PR #1101, opened 2025-11-16
Fix: `update --rename` and `tag --rename` and add new value only if old one exists

jghauser:
This fixes a bunch of things with `papis update` and `papis tag`. Mostly, there's currently a bug with `--rename`. This command first removes an old value from a list and then appends the new value. Before, it appended the value whenever the removing succeeded. Unfortunately, success doesn't actually indicate that something was removed, leading to the new value being added when the old value didn't exist.

While I was fixing that I also:
- made messages to users consistent and less confusing. Before, we said we're 'Updating' documents even when no changes to them are made. Now it just says 'Processing'. Messages are also now similar for both the `update` and `tag` command.
- there was a type error that I fixed

### DIFF
--- a/papis/commands/tag.py
+++ b/papis/commands/tag.py
@@ -175,6 +175,9 @@ def cli(
             success = run_rename(ctx.data, to_rename_tuples, key_types, False)
 
         if success:
+            from papis.document import describe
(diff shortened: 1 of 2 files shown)


jghauser:
Thanks for taking a look!

Issue #948, opened 2024-11-17
Replace unmaintained `isbnlib` dependency


mycorrhizan:
No commits have been made since 2024-07-28 ([ddc62f2](https://github.com/xlcnd/isbnlib/commit/ddc62f2b77214205a7413b5d049c2a008c5271d4)) and CI/CD tests have been consistently failing since 2023-10-19 ([#2007](https://github.com/xlcnd/isbnlib/actions/runs/6570266426)) starting with the Windows tests, with the latest CI/CD test as of this issue failing for all tests ([#2404](https://github.com/xlcnd/isbnlib/actions/runs/11876293124)).

alexfikl:
Do you know of any possible replacement?

From a quick skim through the failures, they seem to be some HTTP failures for too many requests, so not quite that dire, but they might hide other things :( The contributions graph looks scary https://github.com/xlcnd/isbnlib/graphs/contributors

alexfikl:
> The most suitable replacement I could find is [idutils](https://pypi.org/project/idutils/), though I'm unsure if it covers all requirements. At the very least it does seem to be consistently maintained.

The feature we actually use from isbnlib is downloading metadata from various sources based on the ISBN. I think they support Google Books and OpenLibrary.org (plus a bunch of plugins), so that would be hard to replace :\

alexfikl:
> Would this be a suitable replacement for those sources?

To start with, we can definitely add some alternative to `isbnlib` for working with ISBNs using those libraries.

alexfikl:
> Sounds like a plan. Is there a list of available sources in the documentation or source code I can refer to for finding replacements?

I meant that we can add some importer plugins as alternatives to the `isbnlib` ones. There's some documentation on how to add those here
https://papis.readthedocs.io/en/latest/plugins.html
and a lot of examples through the code (all the classes that inherit from `papis.importer.Importer`).

fdf-uni:
Hey, I recently stumbled upon this issue as I also got the warning from papis regarding `setuptools`.
I've since implemented a small (~200 lines) python script which I think would cover the needs of papis (i.e. checking whether a string is a valid ISBN and fetching metadata from Google Books, Open Library and Wikipedia), at least from what I've gathered after grepping the code for `isbnlib`.

I'd happily contribute it in some way if you'd wish so and could perhaps give some instructions in what way to do this. :)
[...]

PR #1102, opened 2025-11-17
add: replacement for unmaintained isbnlib

fdf-uni:
As discussed in #948, this commit implements a replacement for the unmaintained `isbnlib` dependency (sorry for taking a bit long, I sadly didn't find the time to do this sooner).
As a side note, thanks a lot for the helpful instructions given there. :)

I did a bit more than just implementing a new `Importer` plugin since I realized when taking a closer look at the code and starting to add my implementation that `isbnlib` is also used in `papis/isbn.py` and `papis/explorers/isbn.py`. Moreover, there was a lot of overlap, so I thought that it would perhaps be better to put my code into `papis/isbn_standalone.py`[^1] and then implement a corresponding `Importer` plugin, as well as an `Explorer` (the latter two are almost completely copy-pasted from the previous versions). Perhaps one drawback is that in this way, there are currently two ISBN importers available, `isbn` (the previous version) and `isbn_standalone` (the new one).

If this goes against what you had originally intended, please just let me know, and I'll refactor the code correspondingly, for example by putting everything only into an `Importer` plugin.

[^1]: I also put a short attribution blurb there, but am not sure whether they way I did it was sufficient.


alexfikl:
> I did a bit more than just implementing a new `Importer` plugin since I realized when taking a closer look at the code and starting to add my implementation that `isbnlib` is also used in `papis/isbn.py` and `papis/explorers/isbn.py`.

It would have been fine if it was just the importer, but since you did all the extra work, it's very welcome :grin:

> Moreover, there was a lot of overlap, so I thought that it would perhaps be better to put my code into papis/isbn_standalone.py

This actually also has a lot of overlap with the `isbnlib` based stuff (which is probably expected). Maybe we can just merge the two? Something like switching this line
https://github.com/papis/papis/blob/9fbdfc955fb30da36430469e298a31557e7b7da3/papis/isbn.py#L28
to
```python
try:
    from isbnlib import isbn_from_words, meta
except ImportError: 
    # use current implementation from `isbn_standalone`
    isbn_from_words = isbn_from_words_fallback
    meta = meta_fallback
```
with your implementations from `isbn_standalone`? It should really be just a drop in, but without the plugin support that `isbnlib` has, which is ok. It would also avoid the very similar importer and explorer. What do you think?

This way isbnlib becomes an optional dependency and if it breaks, that's fine :grin:

fdf-uni:
Thanks a lot for the feedback! :)
I think your suggestion on merging the implementations makes a lot of sense, so I've refactored the code a bit.

The isbn explorer is now in fact completely unchanged and for the importer I only changed an import statement and moved one check into `papis/isbn.py`.
This way, all logic on whether `isbnlib` is used or not takes place in that file.

Perhaps one last question I'd have is regarding the `isbn_from_words` method.
In the original implementation of `isbnlib`, it essentially does a Google search for "ISBN" together with the given text and scrapes the result for ISBN-like strings.
In my implementation I've replaced it by an API call to Google Books.
I'm mostly mentioning this since no matter which `isbn-service` one has configured, both implementations will currently send a request to Google which is perhaps slightly undesirable.  
I think that it should be possible to simply replace the first line in

https://github.com/fdf-uni/papis/blob/c20dd0abc3c5a5e53debbe9194aceae5cf1c2b48/papis/isbn.py#L223C1-L224C52

by
```python
isbn = "".join(c for c in query.upper() if c in "0123456789X")
```
as long as we assume the provided queries to be ISBNs.
If you think that this would be a better idea, I'd go ahead and test this a bit, before creating an updated commit. :)

alexfikl:
> If you think that this would be a better idea, I'd go ahead and test this a bit, before creating an updated commit. :)

Hm, personally, I do just copy paste (sometimes badly formatted) ISBNs into the importer, so this would be completely fine. However, the `explorer` seems to just pass in random queries to `get_data`, which would break in a fun way :D 

How about modifying `get_data` to something like
```python
def get_data(query: str = "",
			 isbn_like: str = "",
             service: str | None = None) -> list[dict[str, Any]]:
```
Then the importer can do `get_data(isbn_like=uri)` and the explorer can continue doing `get_data(query=query)`. We could change the logic to
```python
if isbn_like:
	isbn = "".join(c for c in query.upper() if c in "0123456789X")
elif query:
	isbn = isbn_from_words(query)
else:
	raise ValueError("must provide either 'isbn_like' or 'query'")
```
? What do you think? It does seem like that `isbn_from_words` functionality is needed if we don't want to break things..

fdf-uni:
Hey, I really liked your proposal, so I updated the code accordingly. :)

As it turned out, the previous version would have broken pretty much in the exact way you had described, since I had misunderstood the purpose of the explorer and thus wrongly queried the Google Books API for a ISBN... :sweat_smile: 
Therefore, I adjusted the `isbn_from_words_fallback` method again, so that it is now rather similar to the original implementation in `isbnlib`.
I also tested whether one could change the search engine later on, and replacing the line
```python
    search_provider_url = "http://www.google.com/search?q="
```
by
```python
    search_provider_url = "https://duckduckgo.com/html/?q="
```
seemed to work rather well, so there's at least the possibility to perhaps make this customizable in the future, although one would then probably also want to either omit calling `googlebooks_isbn_search` at the end of `isbn_from_words_fallback` or adjust this line in another way (perhaps naively checking whether a string is a ISBN using only `notisbn_fallback` would suffice; currently any found ISBN-like strings are searched for using the Google Books API in order to verify them as proper ISBNs).

Please let me know if you have any other suggestions! :)

papis/isbn.py
```diff
@@ -11,24 +17,256 @@
...
+        return s % 11 != 0
+    # Check currently available GS1 prefixes for ISBN-13
+    if isbn[0:3] not in {"978", "979"}:
+        return True
+    # Validate checksum digit for ISBN-13
+    return sum((1 + 2 * (i % 2)) * int(isbn[i]) for i in range(13)) % 10 != 0
+
+
+def json_request(url: str) -> Any | None:
+    r = requests.get(url, headers={"user-agent": USER_AGENT})
```
alexfikl: We have a session helper in `papis.utils.get_session` that should be used here. It uses the User Agent set in the configuration file, etc.

It's probably a good idea to pass that around to all the fallback functions.

papis/isbn.py
```diff
@@ -11,24 +17,256 @@
...
+    except requests.exceptions.JSONDecodeError:
+        return None
+    return data
+
+
+def googlebooks_isbn_search(query: str) -> str:
+    # This assumes the query to be an ISBN
+    query = strip_isbnlike(query)
+    base_api_link = "https://www.googleapis.com/books/v1/volumes?q=isbn:"
+    data = json_request(requests.utils.requote_uri(base_api_link + query))
```
alexfikl: This would look nicer if passed a `params={"isbn": isbn}` sort of thing that just gets forwarded to `requests.get`? So we don't have to do the awkward string concatenation / quoting in here.

papis/isbn.py
```diff
@@ -11,24 +17,256 @@
...
+    if not data:
+        return ""
+    try:
+        items = data["items"]
+    except KeyError:
+        return ""
```
alexfikl: Maybe just a `if "items" not in data: return`?

papis/isbn.py
```diff
@@ -11,24 +17,256 @@
...
+    for i in items:
+        for d in i.get("volumeInfo", {}).get("industryIdentifiers", {}):
+            # isbnlib also doesn't consider ISBN-10 results
+            if d.get("type") == "ISBN_13":
+                return str(d["identifier"])
+    return ""
+
+
+def isbn_from_words_fallback(query: str) -> str:
+    search_provider_url = "http://www.google.com/search?q="
```
alexfikl: Could this also use the Google Books API? With something like `q={query}`, so that it searches through everything.

Is isbnlib also just scrapping Google results? Also, we have some config options that could be useful here: [`search-engine`](https://papis.readthedocs.io/en/latest/configuration.html#confval-search-engine) and [`browse-query-format`](https://papis.readthedocs.io/en/latest/configuration.html#confval-browse-query-format).

papis/isbn.py
```diff
@@ -11,24 +17,256 @@
...
+    if r.status_code != 200:
+        return None
+    try:
+        data = r.json()
+    except requests.exceptions.JSONDecodeError:
+        return None
+    return data
+
+
+def googlebooks_isbn_search(query: str) -> str:
```
alexfikl: This should probably just return `None` if it fails. (or raise some sort of exception?)

papis/isbn.py
```diff
@@ -11,24 +17,256 @@
...
+    r = requests.get(
+        search_url,
+        headers={
+            "user-agent": USER_AGENT,
+            "Content-Type": 'text/plain; charset="UTF-8"',
+            "Content-Transfer-Encoding": "Quoted-Printable",
+        },
+    )
+    regex = re.compile(
+        r"97[89]-?[0-9]{10}|" r"97[89]-[-0-9]{13}|" r"[-0-9]{9,15}[0-9xX]",
```
alexfikl:
```suggestion
        r"97[89]-?[0-9]{10}|97[89]-[-0-9]{13}|[-0-9]{9,15}[0-9xX]",
```
? Not quite sure that that space is doing there?

papis/isbn.py
alexfikl: Would it be cleaner to pass in `params={"q": f"isbn:{query}"}` here and just let requests quote it whenever it wants?

papis/isbn.py
```diff
@@ -11,24 +17,256 @@
...
+                break
+    return isbn
+
+
+def meta_goob(isbn: str) -> dict[str, Any] | None:
+    url = (
+        "https://www.googleapis.com/books/v1/volumes?q={isbn_query}&fields="
+        "items/volumeInfo(title,subtitle,authors,publisher,publishedDate,"
+        "language,industryIdentifiers,description,imageLinks)&maxResults=1"
+    )
```
alexfikl: Would it be possible to add most of these things in a `params` dictionary to pass to `requests`? It looks very nasty like this, haha.

papis/isbn.py
```diff
@@ -11,24 +17,256 @@
...
+    meta["ISBN-13"] = isbn
+    title = volumeinfo.get("title", "").replace(" :", ":")
+    subtitle = volumeinfo.get("subtitle", "")
+    meta["Title"] = format_title(title, subtitle)
+    meta["Authors"] = volumeinfo.get("authors", [""])
+    meta["Publisher"] = volumeinfo.get("publisher", "").strip('"')
+    if "publishedDate" in volumeinfo and len(volumeinfo["publishedDate"]) >= 4:
+        meta["Year"] = volumeinfo["publishedDate"][0:4]
+    else:
+        meta["Year"] = ""
```
alexfikl, line 151: Would it be better to just not add these fields if they don't exist in `volumeinfo`? What does `isbnlib` do?

papis/isbn.py
```diff
@@ -11,24 +17,256 @@
...
+    try:
+        volumeinfo = data["items"][0]["volumeInfo"]
+    except KeyError:
+        return None
+    # Check that ISBNs match
+    ids = repr(volumeinfo.get("industryIdentifiers", ""))
+    if "ISBN_13" in ids and isbn not in ids:
+        raise ValueError("ISBN mismatch")
+    meta: dict[str, Any] = {}
+    meta["ISBN-13"] = isbn
```
alexfikl, line 142: I might have missed something, but are we guaranteed this is an ISBN-13?

papis/isbn.py
```diff
@@ -11,24 +17,256 @@
...
+    elif service == "wiki":
+        return meta_wiki(isbn)
+    else:
+        return None
+
+
+try:
+    from isbnlib import notisbn
+
+    __all__ = ["notisbn"]
```
alexfikl: Other functions are also imported from this file. Is this necessary for something?

papis/isbn.py
```diff
@@ -11,24 +17,256 @@
...
         service = papis.config.getstring("isbn-service")
 
     if service not in ISBN_SERVICE_NAMES:
-        logger.error("ISBN service '%s' is not known. Available services: '%s'.",
-                     service, "', '".join(ISBN_SERVICE_NAMES))
+        logger.error(
+            "ISBN service '%s' is not known. Available services: '%s'.",
+            service,
+            "', '".join(ISBN_SERVICE_NAMES),
+        )
```
alexfikl: Too much `ruff format`-ing? :grin: Could you revert these changes (and some below too)?

papis/isbn.py
```diff
@@ -11,24 +17,256 @@
...
+        try:
+            if not isbn:
+                isbn = isbn_from_words(query)
+            data = meta(isbn, service=service)
+        except ISBNLibException:
+            return []
+    except ImportError:
+        if not isbn:
+            isbn = isbn_from_words_fallback(query)
+        data = meta_fallback(isbn, service=service)
```
alexfikl: From my comment above, this should probably be wrapped in a session, i.e.
```
with get_session() as session:
	if not isbn:
		isbn = isbn_from_words_fallback(session, query)
	data = meta_fallback(session, isbn, service=service)
```

papis/isbn.py
```diff
@@ -11,24 +17,256 @@
 #: A list of services supported by :mod:`isbnlib`. Note that not all versions
 #: have support for all of these versions.
 ISBN_SERVICE_NAMES = ("goob", "openl", "wiki")
+USER_AGENT = "Mozilla/5.0"
+
+
+def format_title(title: str | None, subtitle: str | None = None) -> str:
+    return (f"{title or ''} - {subtitle}" if subtitle else title or "").strip(
+        ",.:;-_ "
+    )
```
alexfikl: Is this how isbnlib normalizes titles too?

papis/isbn.py
```diff
@@ -11,24 +17,256 @@
...
+def strip_isbnlike(isbn_like: str) -> str:
+    return "".join(c for c in isbn_like.upper() if c in "0123456789X")
+
+
+def notisbn_fallback(isbn: str | None) -> bool:
+    if isbn is None:
+        return True
+    # Remove non-required symbols such as hyphens or spaces
+    isbn = strip_isbnlike(isbn)
+    # Check length and position of X (only possible for ISBN-10 at position 10)
```
alexfikl, line 43: Add a link to some place that describes the ISBN format a bit? Wikipedia page is fine if it has most of these requirements. It just needs to be little start for those of us who are not ISBN experts :grin:

alexfikl (commented):
This is looking good functionality wise! :grin: Left a few comments and nitpicks here and there.

It would be nice if we could also modify the tests to check the fallback versions too. Not quite sure how to do that in a nice way though..

papis/isbn.py
alexfikl: Hm, I see that our conversion code also just looks at ISBN-13, so this is probably needed :\

papis/isbn.py
fdf-uni: I don't see anything that would speak against just using the Google Books API, however I followed the implementation of `isbnlib` here which also just scraps Google results.
Feel free to let me know which approach you'd prefer and I'll adjust the code accordingly! :)

papis/isbn.py
fdf-uni: Oh, sorry, this was originally a longer regex which had to be split over multiple lines.

papis/isbn.py
fdf-uni: `isbnlib` also sets them to the empty string, see for example https://github.com/xlcnd/isbnlib/blob/ddc62f2b77214205a7413b5d049c2a008c5271d4/isbnlib/_goob.py#L35.
Since I wasn't sure what `papis` would expect in these cases, and whether it perhaps requires these fields to be existent, I followed the original implementation, but I could have a look into whether we can also simply leave the fields non-existent, if this would be preferred.

papis/isbn.py
fdf-uni: I added this since I wanted to abstract the import of `notisbn` in the ISBN importer, where I replaced
```python
from isbnlib import notisbn
```
by
```python
from papis.isbn import notisbn
```
so that whether `isbnlib` or the new fallback is used is entirely contained within `papis/isbn.py`.
Without `__all__ = ["notisnb"]` the `notisbn` method from `isbnlib` could not be passed on to the importer code.

I could of course also add another try except statement in the importer if you'd prefer that, but I'm also open for any other suggestions. :)

papis/isbn.py
fdf-uni: Sorry about that! :sweat_smile:

papis/isbn.py
fdf-uni: Yes, see e.g. here:
https://github.com/xlcnd/isbnlib/blob/ddc62f2b77214205a7413b5d049c2a008c5271d4/isbnlib/_goob.py#L27
https://github.com/xlcnd/isbnlib/blob/ddc62f2b77214205a7413b5d049c2a008c5271d4/isbnlib/dev/_data.py#L46

I added the `or ''` in case, the title might be `None`.

fdf-uni:
Wow, thank you so much for the detailed feedback, I really appreciate it! :)
I've answered some of your comments and would start implementing your corrections once we've decided on some of the details (e.g. whether to scrap Google results or use the Google Books API). :+1:

papis/isbn.py
fdf-uni: Just to make sure, are you referring to `googlebooks_isbn_search`?

papis/isbn.py
alexfikl: I see, then let's leave it as is to match `isbnlib`.

papis/isbn.py
alexfikl: Yep. It was returning some empty strings in some failure cases and `None` seems more appropriate?

papis/isbn.py
alexfikl: I think our `data_to_papis` handles missing values just fine, so it shouldn't be a problem on that end. It's fine to leave it as is for now if this is how `isbnlib` did it.

papis/isbn.py
alexfikl: Hm, is this because of `mypy` or something complaining about an implicit export? It should be fine to just add a `type: ignore` or `noqa` in that case.

papis/isbn.py
alexfikl: Makes sense, thanks!

fdf-uni:
Hey, I'm very sorry for the long delay, but I finally got around to finishing this PR. :)
I didn't have a look at updating the tests so far, since I saw that you recently deprecated `isbnlib` in #1147, so I wasn't entirely sure how you'd like to handle them.

alexfikl:
> I didn't have a look at updating the tests so far, since I saw that you recently deprecated `isbnlib` in #1147, so I wasn't entirely sure how you'd like to handle them.

That just made `isbnlib` an optional dependency so that it doesn't break everything for users on newer versions. Since this copies over the stuff we actually use, I think it's better to just remove the `isbnlib` dependency completely.

For what it's worth, there seems to be a maintained fork of `isbnlib` at https://github.com/hans-fritz-pommes/isbnlib. I'm wondering if it's worth adding a fallback for that? Not quite sure how long it's going to survive though..

fdf-uni:
Ah, I see. Should I go ahead and remove `isbnlib` from `pyproject.toml`?

I think the other question would be whether we also remove these two `try` blocks which are the only places where `isbnlib` comes into play:
https://github.com/papis/papis/blob/2eddc7e689bf7536f67247f6cfbbe24399f83033/papis/isbn.py#L225-L228
https://github.com/papis/papis/blob/2eddc7e689bf7536f67247f6cfbbe24399f83033/papis/isbn.py#L250-L262

Right now, we could simply add the new `isbnlib` fork as a dependency and it would be used, so in some sense there already is a fallback for it. On the other hand, in case we remove the try blocks, I'd rename the earlier functions and remove the `fallback` suffix.

alexfikl:
> Ah, I see. Should I go ahead and remove `isbnlib` from `pyproject.toml`?

Yeah, just nuke it :grin:

> Right now, we could simply add the new isbnlib fork as a dependency and it would be used

I think that can wait a bit. It's not particularly clear how long it will survive..

I actually had a question that I keep forgetting. A lot of the commits to both `isbnlib` and `isbnlib2` seem to be about updating the mapping from ISBN prefixes to countries / publishers (in `_data/data4info.py`). We don't use/need that, do we?

fdf-uni:
> I actually had a question that I keep forgetting. A lot of the commits to both isbnlib and isbnlib2 seem to be about updating the mapping from ISBN prefixes to countries / publishers (in _data/data4info.py). We don't use/need that, do we?

Exactly, at the moment we don't make use of ISBN prefixes.👍

alexfikl:
@fdf-uni Is this ready to go? I just took a quick look and it looks ready to go for me :grin: (once the little CI failure is fixed)

fdf-uni:
Yes, I think this implementation is ready now! :)
And sorry for missing that formatting error.

alexfikl:
Awesome! Thanks again for working on this and sorry for all the delays :facepalm:

In it goes as soon as the CI passes :rocket:

fdf-uni:
Absolutely no worries, I admittedly feel like the delays were actually mostly on my end. ;)

And once again thanks for all your feedback, I really enjoyed the contribution process and learned some new things throughout it!😁

Issue #894, opened 2024-08-02
papis rename does not rename the PDF files when add-file-name is set


stylerw:
* papis version (`$ papis --version` or commit number): 129773d

**Expected Behavior**:

Running `papis rename -a` will change the folder names, and filenames of associated PDFs, when run, according to the rules in .config for add-folder-name and add-file-name as below

```
add-file-name = {doc[author]}_{doc[year]}_{doc[title]}
add-folder-name = {doc[author]}_{doc[year]}
```

**Observed Behavior**:

Running `papis rename -ab` will change the folder names, but filenames will not be updated to match the set standard

**Why this is a problem**: Many of us use tools like fzf to find files rapidly, and semantic filenames are useful. It is also silly to have to manually change the filename when a bibtex entry exists which has the information I'd like.

Thanks!

pmiam:
I have also been having problems with `papis rename` in structured library I recently set up for a long-term project. I'll try to take a look at this soon.

klb2:
Hi,
Are there any updates on this issue or a temporary solution?

alexfikl:
> Hi, Are there any updates on this issue or a temporary solution?

There's some talk in #1104 to merge the move and rename commands that should include this bug. It's still early stages (as you can probably see :grin:)

klb2:
Great thank you. Looking forward to this

jghauser:
Just a quick comment on this. While I've made a fair bit of progress on #1104, this will not actually fix the issue here. The proposal at the moment is to improve `papis update` to rename files instead. @pmiam [once](https://github.com/papis/papis/pull/826#issuecomment-2467159515) expressed interest in working towards that, but I'm not sure whether they're still interested in it. I might get to it once I finish the work on `papis mv`. Eventually, we might also want to have a flag in `papis mv` to trigger a filename update with `papis update`.

alexfikl:
> Just a quick comment on this. While I've made a fair bit of progress on [#1104](https://github.com/papis/papis/pull/1104), this will not actually fix the issue here.

Oops, sorry for the confusion!

> I might get to it once I finish the work on papis mv. Eventually, we might also want to have a flag in papis mv to trigger a filename update with papis update.

This fell off my radar as well, but it sounds rather simple to modify `papis update` to rename files as mentioned, so I might give it a go over the weekend. :grin:

jghauser:
[...]

PR #1104, opened 2025-11-17
Improve `papis mv`

jghauser:
---
~~For now this PR is just a placeholder to discuss things~~
---

I want to take another stab at improving `papis mv` based on the work done by @kiike (thanks! :heart:) in #826. I think the work in that PR is moving things in the right direction and I'd like to push it over the finish line. There was a lengthy discussion at the time on how to best do this, and I've come up with the below proposal based on what was said then. I would love getting feedback on this from anyone but especially those involved in the discussion at the time (@pmiam, @alexfikl, @kiike (though I think you don't use Papis anymore)).

In short, I want to achieve all of kiike's goals, but I think something along the lines of @pmiam's syntax proposal is most minimal/understandable.

## `papis mv -t/--to`
1. `papis mv -t somefolder QUERY`
  - moves folder to somefolder (a rename)
  - if you do this with multiple documents, you will get a warning about overwriting after the first mv operation (at the latest)
2. `papis mv -t somefolder/ QUERY`
  - moves folder *into* somefolder (a move)

Both versions should:
- should support format patterns
- should allow `somefolder/subfolder(/)`
- should split at `/` and sanitize the rest

Note that @pmiam proposed globs '*' instead of trailing `/`. I think that could lead to issues with regular shell globbing. Trailing `/` don't present that problem and are how e.g. `rsync` does it.

## `papis mv QUERY`

This command uses the `add-folder-name` option to determine the target folder name.

## General requirements

In general, `papis mv` should:
- handle multiple files (from picker or with `--all`) (with confirmation or without when using `--batch`)
- handle git
   - automatic commit after mv
- warn about overwrites (but not when using `--batch`?)
- use stuff from `papis.paths` wherever possible

## Regarding `papis rename`

At the time I thought we should preserve `papis rename`, but now I think we should remove it. The command is confusing (since it's unclear whether it acts on folders or files, see #894), and its current functionality should be included it `papis mv`. Renaming of files (and notes) should be done by `papis update`.

## Questions and remarks

- The `--batch` flag in general removes all need to interactively approve things. However, we might want to be able to move a ton of docs at once but still be warned about overwrites. What should we do? Having `--batch` for this but with overwrite warnings is inconsistent with other uses of `--batch`. Maybe we just live with the fact that the user will have to press 'ok' a bunch of times?
- The proposed implementation has no way of moving folder into another folder and renaming it to the `add-folder-name` option
  - You would need to specify the format pattern explicitly, e.g. `papis mv -t somefolder/{doc[author]}-{doc[title]}`
  - I guess that's fine?
- As mentioned by @alexfikl in the original discussion, what should we do on failure: abort or continue?
- The proposed syntax is inconsistent with `papis add`, which has `--subfolder` and `--folder-name` arguments
  - I think the old syntax is rather confusing, and I would prefer adding `-t/--to` to `papis add` and maybe eventually deprecate the old arguments

### DIFF
--- a/papis/commands/mv.py
+++ b/papis/commands/mv.py
@@ -1,111 +1,489 @@
 """
-This command can be used to move a document to a new folder.
+The ``mv`` command is used to move document folders around your library (or to
+another library).
 
-Command-line interface
+It will (except when run with ``--all``) bring up the picker with a list of
+documents that match the query. You can then select the document or documents
+you'd like to move.
+
+The command works similarly to the Unix
+`mv <https://en.wikipedia.org/wiki/Mv_(Unix)>`__ command. Hence, when the
+specified target folder exists, the document's current main folder is moved into
+the target folder. If the specified target folder doesn't exist, then the
+current folder is renamed to the target folder name. Unlike Unix ``mv``, if a
+target folder isn't given, it defaults to the :confval:`add-folder-name` pattern,
+so ``papis mv QUERY`` renames matching documents to their default folder names.
+
+Folder names are cleaned, so that characters that can lead to invalid file paths
+are avoided (see :confval:`doc-paths-extra-chars` for a list of allowed characters
+and how to change it). This cleaning applies to all path components,
+including any literal subfolders passed to ``--to``. This means and you cannot use
+the ``mv`` command to move docs into a folder containing excluded characters.
+
+The ``mv`` command will first check whether moving the folders as specified will
+fail. If any errors are encountered, they are displayed and you will
+be prompted whether you want to abort or move the unproblematic documents (if any
+exist).
+
+Examples
+^^^^^^^^
+
+- Query documents by author "Rick Astley" and move some of them. After picking
+  the relevant documents, they will be renamed to their default names
+  specified by the :confval:`add-folder-name` option.
+
+   .. code:: sh
+
+        papis mv author:"Rick Astley"
+
+- You can use ``--to`` to pass the path of the target folder, relative to the
+  library root. Here we'll move a document by the author "Rick Astley" to a
+  folder named ``rick-astley``:
+
+   .. code:: sh
+
+        papis mv --to rick-astley author:"Rick Astley"
+
+  If the folder "rick-astley" doesn't exist, then the document's folder will be
+  renamed to that so that the new folder becomes "rick-astley". If the folder
+  "rick-astley" exists, then the document's folder will be moved into it. If, say,
+  the document's folder is currently "rick-s-the-best", then the new folder will
+  be "rick-astley/rick-s-the-best".
+
+  If multiple selected documents resolve to the exact same folder name, Papis will
+  warn you and skip those documents. Consider the above case when the target folder
+  doesn't exist. Here, each document would be renamed to ``rick-astley``, causing
+  Papis to skip the documents to avoid collisions. In contrast, if the target folder
+  already exists, each document will be placed into its own subfolder (e.g.
+  "rick-astley/rick-s-the-best" and "rick-astley/rick-s-very-cool"), so there is
+  no risk of collision.
+
+- You can use formatting patterns to generate the folder names. For instance, to
+  organise all documents by author "Rick Astley" by year, you can use this command:
+
+   .. code:: sh
+
+        papis mv --to "{doc[year]}" --all author:"Rick Astley"
+
+  Note that the ``--all`` flag means that the picker isn't shown and all
+  documents matching the query will be moved.
+
+- You can use the ``--batch`` flag to skip all prompts. You will not be
+  prompted if there are problematic moves and all unproblematic moves will
+  happen automatically. If an unexpected error occurs while moving a
+  document (e.g. a permission error), the command logs a warning and
+  continues with the next document instead of aborting.
+
+   .. code:: sh
+
+        papis mv --batch --all author:"Rick Astley"
+
+Command-line Interface
 ^^^^^^^^^^^^^^^^^^^^^^
 
 .. click:: papis.commands.mv:cli
     :prog: papis mv
 """
+
 from __future__ import annotations
 
-import os
-from typing import TYPE_CHECKING
+import collections
+import shutil
+from pathlib import Path
+from typing import TYPE_CHECKING, NamedTuple
 
 import click
 
 import papis.cli
 import papis.config
+import papis.database
+import papis.document
+import papis.git
 import papis.logging
+import papis.paths
+import papis.tui.utils
+from papis.exceptions import DocumentFolderNotFound
+from papis.strings import AnyString, no_documents_retrieved_message
 
 if TYPE_CHECKING:
     from papis.document import Document
 
+
+class PlannedMove(NamedTuple):
+    document: Document
+    source: Path
+    target: Path
+
+
 logger = papis.logging.get_logger(__name__)
 
 
-def run(document: Document,
-        new_folder_path: str,
-        git: bool = False) -> None:
-    from papis.document import describe
+def _check_not_nested(target: Path, lib_dir: Path) -> str | None:
+    """Check that *target* would not be placed inside another document folder.
 
-    folder = document.get_main_folder()
-    if not folder:
-        from papis.exceptions import DocumentFolderNotFound
-        raise DocumentFolderNotFound(describe(document))
+    Walks up from *target.parent* to *lib_dir*. If any ancestor is a
+    document folder, this is considered an error.
 
-    if git:
-        from papis.git import mv as git_mv
-        git_mv(folder, new_folder_path)
+    :param target: the target path for the document
+    :param lib_dir: the root directory of the target library.
+    :returns: ``None`` if not nested in a document folder, an error string otherwise
+    """
+    current = target.parent
+    while current not in {lib_dir, current.parent}:
+        if papis.document.is_document_folder(current):
+            return f"would nest inside another document's folder: {current}"
+        current = current.parent
+    return None
+
+
+def _check_target_already_exists(path: Path) -> str | None:
+    """Check that *path* doesn't already exist.
+
+    :returns: ``None`` if the path doesn't already exist, an error string otherwise
+    """
+    if path.exists():
+        return f"target already exists: {path}"
+    return None
+
+
+def _rename_target_interactively(
+    document: Document,
+    default_target: Path,
+    lib_dir: Path,
+) -> Path:
+    """Prompt the user to rename until a valid target is found.
+
+    :param document: the document being moved.
+    :param default_target: the default/computed target path (absolute).
+    :param lib_dir: the root directory of the target library.
+    :returns: a valid absolute target path.
+    """
+    while True:
+        new_name = papis.tui.utils.prompt(
+            f"Folder name for '{papis.document.describe(document)}'",
+            default=str(default_target.relative_to(lib_dir)),
+        )
+        new_target = Path(
+            papis.paths.get_document_folder(
+                document, str(lib_dir), folder_name_format=new_name
+            )
+        )
+
+        error = _check_not_nested(new_target, lib_dir)
+        if error is None:
+            error = _check_target_already_exists(new_target)
+        if error is None:
+            return new_target
+
+        logger.warning(
+            "'%s': %s.",
+            papis.document.describe(document),
+            error,
+        )
+
+
+def _register_move(
+    move: PlannedMove,
+    moves: dict[Path, PlannedMove],
+    conflicts: dict[Path, list[PlannedMove]],
+) -> None:
+    """Insert *move* into *moves*, or bump it to *conflicts* on collision.
+
+    Resolves the move's target path and checks whether it is already claimed.
+    If the slot is free the move goes into *moves*; if occupied the existing
+    occupant is moved to *conflicts* together with the new move.
+    """
+    target_path = move.target.resolve()
+    if target_path in moves:
+        existing = moves.pop(target_path)
+        conflicts[target_path] = [existing, move]
+    elif target_path in conflicts:
+        conflicts[target_path].append(move)
     else:
-        import shutil
-        shutil.move(folder, new_folder_path)
+        moves[target_path] = move
 
-    from papis.database import get_database
-    db = get_database()
-    db.delete(document)
 
-    new_document_folder = os.path.join(
-        new_folder_path,
-        os.path.basename(folder))
-    logger.debug("New document folder: '%s'.", new_document_folder)
+def run(
+    document: Document,
+    target_path_abs: str,
+    target_library: str | None = None,
+    git: bool = False,
+) -> None:
+    """Move a document's folder to *target_path_abs*.
 
-    document.set_folder(new_document_folder)
-    db.add(document)
+    :param document: the document to move.
+    :param target_path_abs: the target folder path (absolute)
+    :param target_library: name of the target library. If ``None``, uses the current
+        library.
+    :param git: if ``True``, record the move in git history.
+    :raises DocumentFolderNotFound: if the document has no main folder on disk.
+    :raises FileNotFoundError: if the source folder does not exist.
+    :raises FileExistsError: if the target folder already exists. This will also be
+        raised if the target folder is the same as the source folder.
+    """
+    if target_library is None:
+        target_library = papis.config.get_lib_name()
+
+    source_lib_dir = Path(papis.config.get_lib().path)
+
+    target_lib_dir = Path(papis.config.get_lib_from_name(target_library).path)
+
+    doc_folder_old = document.get_main_folder()
+    if not doc_folder_old:
+        raise DocumentFolderNotFound(papis.document.describe(document))
+
+    main_folder_old = Path(doc_folder_old)
+    main_folder_new = Path(target_path_abs)
+
+    if not main_folder_old.exists():
+        raise FileNotFoundError(f"Source folder does not exist: '{main_folder_old}'")
+
+    if main_folder_new.exists():
+        raise FileExistsError(f"Target folder already exists: '{main_folder_new}'")
+
+    source_library = papis.config.get_lib_name()
+    papis.config.set_lib_from_name(target_library)
+
+    main_folder_new.parent.mkdir(
+        parents=True,
+        exist_ok=True,
+        # NOTE: 0o755 is default, but explicitness is needed to avoid type error
+        mode=papis.config.getint("dir-umask") or 0o755,
+    )
+
+    papis.config.set_lib_from_name(source_library)
+
+    if target_lib_dir == source_lib_dir:
+        if git:
+            papis.git.mv_and_commit_resource(
+                str(main_folder_old),
+                str(main_folder_new),
+                f"Move '{papis.document.describe(document)}'",
+            )
+        else:
+            shutil.move(main_folder_old, main_folder_new)
+    else:
+        shutil.move(doc_folder_old, main_folder_new)
+
+        if git:
+            papis.git.remove(str(source_lib_dir), str(main_folder_old))
+            papis.git.commit(
+                str(source_lib_dir),
+                f"Remove '{papis.document.describe(document)}' "
+                f"(moved to '{target_library}' library)",
+            )
+
+            papis.git.add_and_commit_resource(
+                str(target_lib_dir),
+                str(main_folder_new),
+                f"Add '{papis.document.describe(document)}'",
+            )
+
+    source_db = papis.database.get()
+    source_db.delete(document)
+
+    target_db = papis.database.get(library_name=target_library)
+
+    document.set_folder(str(main_folder_new))
+
+    target_db.add(document)
+
+    logger.info(
+        "Moved document '%s' to '%s' (library '%s').",
+        papis.document.describe(document),
+        main_folder_new,
+        target_library,
+    )
 
 
 @click.command("mv")
 @click.help_option("--help", "-h")
+@click.option(
+    "-t",
+    "--to",
+    help="Path relative to the library root",
+    default=None,
+    type=papis.cli.FormatPatternParamType(),
+)
+@click.option(
+    "--target-library",
+    help="Target library",
+    default=lambda: papis.config.get_lib_name(),  # noqa: PLW0108
+    type=str,
+)
+@papis.cli.bool_flag(
+    "-b",
+    "--batch",
+    help="Do not prompt. Warnings are logged but moves proceed automatically.",
+)
 @papis.cli.query_argument()
 @papis.cli.git_option()
 @papis.cli.sort_option()
 @papis.cli.doc_folder_option()
-def cli(query: str,
-        git: bool,
-        sort_field: str | None,
-        doc_folder: tuple[str, ...],
-        sort_reverse: bool) -> None:
-    """Move a document into some other path."""
-    # Leave this imports here for performance
-    import prompt_toolkit
-    import prompt_toolkit.completion
+@papis.cli.all_option()
+def cli(
+    query: str,
+    git: bool,
+    _all: bool,
+    to: AnyString | None,
+    target_library: str,
+    batch: bool,
+    sort_field: str | None,
+    doc_folder: tuple[str, ...],
+    sort_reverse: bool,
+) -> None:
+    """Move a document into another folder."""
 
-    documents = papis.cli.handle_doc_folder_query_sort(query,
-                                                       doc_folder,
-                                                       sort_field,
-                                                       sort_reverse)
+    documents = papis.cli.handle_doc_folder_query_all_sort(
+        query, doc_folder, sort_field, sort_reverse, _all
+    )
     if not documents:
-        from papis.strings import no_documents_retrieved_message
         logger.warning(no_documents_retrieved_message)
         return
 
-    document = documents[0]
+    source_library = papis.config.get_lib_name()
+    papis.config.set_lib_from_name(target_library)
 
-    lib_dir = os.path.expanduser(papis.config.get_lib().path)
+    if to is None:
+        to = papis.config.getformatpattern("add-folder-name")
 
-    completer = prompt_toolkit.completion.PathCompleter(
-        only_directories=True,
-        get_paths=lambda: [lib_dir]
-    )
+    target_lib_dir = Path(papis.config.get_lib_from_name(target_library).path)
 
-    try:
-        new_folder = os.path.join(
-            lib_dir,
-            prompt_toolkit.prompt(
-                message=(
-                    "Enter directory  : (Tab completion enabled)\n"
-                    f"Current directory: ({document.get_main_folder_name()})\n >"),
-                completer=completer,
-                complete_while_typing=True
-            ))
-    except Exception as exc:
-        logger.error("Failed to choose directory.", exc_info=exc)
+    moves: dict[Path, PlannedMove] = {}
+    conflicts: dict[Path, list[PlannedMove]] = collections.defaultdict(list)
+
+    # Gather moves, detect disk conflicts + selection collisions
+    for document in documents:
+        main_folder_old_str = document.get_main_folder()
+        if not main_folder_old_str:
+            logger.warning(
+                "'%s': source folder not defined. Cannot move.",
+                papis.document.describe(document),
+            )
+            continue
+
+        main_folder_old = Path(main_folder_old_str)
+        if not main_folder_old.exists():
+            logger.warning(
+                "'%s': source folder does not exist on disk. Cannot move: %s.",
+                papis.document.describe(document),
+                main_folder_old,
+            )
+            continue
+
+        target_abs_path = Path(
+            papis.paths.get_document_folder(
+                document, target_lib_dir, folder_name_format=to
+            )
+        )
+
+        if main_folder_old == target_abs_path:
+            logger.info(
+                "Skipping '%s': source and target are the same.",
+                papis.document.describe(document),
+            )
+            continue
+
+        # Resolve rename vs move-into semantics
+        if target_abs_path.is_dir():
+            main_folder_new = target_abs_path / main_folder_old.name
+        else:
+            main_folder_new = target_abs_path
+
+        # Check for nesting or disk conflict
+        error = _check_not_nested(main_folder_new, target_lib_dir)
+        if error is None:
+            error = _check_target_already_exists(main_folder_new)
+        if error:
+            logger.warning(
+                "'%s': %s.",
+                papis.document.describe(document),
+                error,
+            )
+            if batch:
+                continue
+            logger.warning("You can skip this document or rename it manually.")
+            if papis.tui.utils.confirm(
+                "Skip this document? Answer 'no' to rename it instead.",
+                yes=True,
+            ):
+                continue
+
+            main_folder_new = _rename_target_interactively(
+                document, main_folder_new, target_lib_dir
+            )
+
+        move = PlannedMove(document, main_folder_old, main_folder_new)
+        _register_move(move, moves, conflicts)
+
+    # Resolve selection collisions.
+    #
+    # At this point every target path is in one of two lists
+    #   moves[path]     - a single PlannedMove, no collision
+    #   conflicts[path] - list of PlannedMove that want the same folder
+    # The loop below lets the user rename colliding moves. Paths that collide again
+    # stay in the loop for another round.
+    while conflicts:
+        if batch:
+            for conflict_group in conflicts.values():
+                for c in conflict_group:
+                    logger.warning(
+                        "'%s': duplicate target folder. Skipping: %s → %s.",
+                        papis.document.describe(c.document),
+                        c.source,
+                        c.target,
+                    )
+            break
+
+        resolved: list[PlannedMove] = []
+
+        for conflict_path, conflict_group in list(conflicts.items()):
+            msg_lines = [
+                f"{len(conflict_group)} documents want the same"
+                f" target folder '{conflict_path}':"
+            ]
+            for c in conflict_group:
+                msg_lines.append(f"    - '{papis.document.describe(c.document)}'")
+
+            logger.warning("\n".join(msg_lines))
+
+            logger.warning("You can skip these documents or rename them manually.")
+            if papis.tui.utils.confirm("Skip these documents?", yes=True):
+                del conflicts[conflict_path]
+                continue
+
+            for c in conflict_group:
+                new_target = _rename_target_interactively(
+                    c.document, c.target, target_lib_dir
+                )
+                resolved.append(PlannedMove(c.document, c.source, new_target))
+
+            del conflicts[conflict_path]
+
+        new_conflicts: dict[Path, list[PlannedMove]] = collections.defaultdict(list)
+        for move in resolved:
+            _register_move(move, moves, new_conflicts)
+
+        if not new_conflicts:
+            break
+
+        conflicts = new_conflicts
+
+    if not moves:
+        logger.info("No valid moves to perform.")
         return
 
-    logger.info("New document folder: '%s'.", new_folder)
-
-    if not os.path.exists(new_folder):
-        logger.info("Creating path '%s'.", new_folder)
-        os.makedirs(new_folder, mode=papis.config.getint("dir-umask") or 0o666)
-
-    run(document, new_folder, git=git)
+    # Move
+    papis.config.set_lib_from_name(source_library)
+    for move in moves.values():
+        try:
+            run(move.document, str(move.target), target_library=target_library, git=git)
+        except Exception as exc:
+            if batch:
+                logger.warning(
+                    "Failed to move '%s': %s",
+                    papis.document.describe(move.document),
+                    exc,
+                )
+                continue
+            raise
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,27 @@ papis open 'title:^Quantum'
 Note that this new query syntax still uses the `match-format` configuration setting
 for free-form terms in the query (i.e. ones not using a `key:value` format).
 
+### Major: Improve `mv` command ([#2822](https://github.com/papis/papis/pull/2822))
+
+The `mv` command has been completely rewritten with many new features. It now
+works similarly to the Unix `mv` command for moving document folders: if the
+target folder already exists, the document's folder is moved *into* it (e.g.
+`papis mv --to existing-folder query` places the document at
+`existing-folder/<doc-folder>`); if the target does not exist, the document is
+renamed to that name. When no target is given, it defaults to the
+`add-folder-name` pattern, which means `papis mv query` now does
+everything that `papis rename` used to do.
+
+The new `--to` flag accepts formatting patterns (e.g.
+`papis mv --to "{doc[year]}" query`). The command now also allows moving documents
+between libraries by specifying a target library.
+
+The new command performs pre-flight checks on all planned moves before touching
+any files. Issues such as duplicate target paths, nesting a document inside
+another document's folder, or missing source folders are warned about upfront. The
+user can abort, skip problematic moves, or fix collisions interactively. A new
+`--batch` flag allows skipping all interactive prompts.
+
 ## Other noteworthy features
 
 - Remove undocumented `dirs` option (only allow one `dir` per library).
--- a/papis/document.py
+++ b/papis/document.py
@@ -11,6 +11,7 @@ import papis.logging
 
 if TYPE_CHECKING:
     from collections.abc import Callable, Sequence
+    from pathlib import Path
     from typing import Literal
 
     from papis.strings import AnyString
@@ -576,6 +577,19 @@ def from_folder(folder_path: str) -> Document:
     :param folder_path: absolute path to a valid ``papis`` folder.
     """
     return Document(folder=folder_path)
+
+
+def is_document_folder(path: Path) -> bool:
+    """Check whether *path* is an existing document folder.
+
+    A folder is considered a document folder if it contains an info file
+    (see :confval:`info-name`).
+
+    :param path: absolute path to a folder on the filesystem.
+    :returns: *True* if the folder contains a document info file.
+    """
+    info_name = papis.config.getstring("info-name")
+    return path.joinpath(info_name).exists()
 
 
 def to_json(document: Document) -> str:
--- a/tests/commands/test_mv.py
+++ b/tests/commands/test_mv.py
@@ -1,30 +1,690 @@
+"""Tests for papis.commands.mv."""
+
 from __future__ import annotations
 
+import logging
 import os
-import tempfile
+import shutil
 from typing import TYPE_CHECKING
 
 import papis.database
+import papis.document
+import papis.tui.utils
+from papis.commands.mv import cli
+from papis.testing import PapisRunner, TemporaryLibrary
 
 if TYPE_CHECKING:
-    from papis.testing import TemporaryLibrary
+    from pathlib import Path
+
+    from _pytest.monkeypatch import MonkeyPatch
+
+    from papis.database.base import Database
 
 
-def test_mv_run(tmp_library: TemporaryLibrary) -> None:
+import pytest
+
+
+def _doc_by_author(db: Database, author: str) -> papis.document.Document:
+    """Find a document by author. Assumes unique match for simplicity."""
+    return next(d for d in db.get_all_documents() if d["author"] == author)
+
+
+def _setup_runner_db_doc() -> tuple[PapisRunner, Database, papis.document.Document]:
+    """Return a runner, the current DB, and Turing's document."""
+    db = papis.database.get()
+    return PapisRunner(), db, _doc_by_author(db, "Turing, A. M.")
+
+
+def _assert_moved(old_folder: str, expected_folder: str) -> None:
+    """Assert the move completed correctly on disk and in the DB."""
+    assert not os.path.exists(old_folder)
+    assert os.path.exists(expected_folder)
+
+    matches = [
+        doc
+        for doc in papis.database.get().get_all_documents()
+        if doc.get_main_folder() == expected_folder
+    ]
+    assert len(matches) == 1
+    assert matches[0].get_main_folder() == expected_folder
+
+
+# ---------------------------------------------------------------------------
+# Tests for cli()
+# ---------------------------------------------------------------------------
+
+
+def test_to_renames_folder(tmp_library: TemporaryLibrary) -> None:
+    """--to renames the document folder."""
+    cli_runner, _, doc = _setup_runner_db_doc()
+    old_folder = doc.get_main_folder()
+    assert old_folder
+
+    result = cli_runner.invoke(
+        cli, ["--to", "renamed-turing", "Turing"], catch_exceptions=False
+    )
+    assert result.exit_code == 0, result.output
+
+    expected = os.path.join(tmp_library.libdir, "renamed-turing")
+    _assert_moved(old_folder, expected)
+
+
+def test_to_with_formatpattern(tmp_library: TemporaryLibrary) -> None:
+    """--to with a format pattern produces the expected folder."""
+    cli_runner, _, doc = _setup_runner_db_doc()
+    old_folder = doc.get_main_folder()
+    assert old_folder
+
+    result = cli_runner.invoke(
+        cli, ["--to", "{doc[year]}", "Turing"], catch_exceptions=False
+    )
+    assert result.exit_code == 0, result.output
+
+    expected = os.path.join(tmp_library.libdir, "1937")
+    _assert_moved(old_folder, expected)
+
+
+def test_without_to_renames_to_default(tmp_library: TemporaryLibrary) -> None:
+    """Without --to, the default folder name is used."""
+    cli_runner, _, doc = _setup_runner_db_doc()
+    old_folder = doc.get_main_folder()
+    assert old_folder
+    doc_id = doc["papis_id"]
+
+    result = cli_runner.invoke(cli, ["Turing"], catch_exceptions=False)
+    assert result.exit_code == 0, result.output
+
+    assert not os.path.exists(old_folder)
+    expected = os.path.join(tmp_library.libdir, doc_id)
+    assert os.path.exists(expected)
+
+
+def test_to_slugification_of_special_chars(tmp_library: TemporaryLibrary) -> None:
+    """--to slugifies the target folder name."""
+    cli_runner, _, doc = _setup_runner_db_doc()
+
+    old_folder = doc.get_main_folder()
+    assert old_folder
+
+    result = cli_runner.invoke(
+        cli, ["--to", "Rick Astley", "Turing"], catch_exceptions=False
+    )
+    assert result.exit_code == 0, result.output
+
+    expected = os.path.join(tmp_library.libdir, "rick-astley")
+    _assert_moved(old_folder, expected)
+
+
+def test_to_existing_dir_moves_into(tmp_library: TemporaryLibrary) -> None:
+    """--to pointing to an existing directory moves into it."""
+    cli_runner, _, doc = _setup_runner_db_doc()
+    old_folder = doc.get_main_folder()
+    assert old_folder
+    old_basename = os.path.basename(old_folder)
+
+    target_dir = os.path.join(tmp_library.libdir, "papers-archive")
+    os.makedirs(target_dir)
+
+    result = cli_runner.invoke(
+        cli, ["--to", "papers-archive", "Turing"], catch_exceptions=False
+    )
+    assert result.exit_code == 0, result.output
+
+    expected = os.path.join(target_dir, old_basename)
+    _assert_moved(old_folder, expected)
+
+
+def test_to_same_folder_is_skipped(tmp_library: TemporaryLibrary) -> None:
+    """Moving to the same folder is a no-op."""
+    cli_runner, _, _ = _setup_runner_db_doc()
+
+    # moving to this first, since the default folder name is something
+    # that would be slugified, making renaming to the same name impossible
+    result = cli_runner.invoke(
+        cli, ["--to", "turing-folder", "Turing"], catch_exceptions=False
+    )
+    assert result.exit_code == 0, result.output
+
+    folder_path = os.path.join(tmp_library.libdir, "turing-folder")
+    result = cli_runner.invoke(
+        cli, ["--to", "turing-folder", "Turing"], catch_exceptions=False
+    )
+    assert result.exit_code == 0, result.output
+    assert os.path.exists(folder_path)
+
+
+def test_to_existing_other_doc_folder_warns(
+    tmp_library: TemporaryLibrary,
+    caplog: pytest.LogCaptureFixture,
+    monkeypatch: MonkeyPatch,
+) -> None:
+    """Moving to another doc folder warns about nesting."""
+    caplog.set_level(logging.WARNING)
+    cli_runner, db, doc = _setup_runner_db_doc()
+    old_folder = doc.get_main_folder()
+    assert old_folder
+
+    # Find another doc (Krishnamurti) and pre-move it to a clean slugified name
+    krish = next(d for d in db.get_all_documents() if d["author"] == "J. Krishnamurti")
+    assert krish.get_main_folder()
+    result = cli_runner.invoke(
+        cli, ["--to", "krish-doc", "Krishnamurti"], catch_exceptions=False
+    )
+    assert result.exit_code == 0, result.output
+
+    # Now try to move Turing into Krish's folder
+    def mock_confirm(msg: str, **kwargs: object) -> bool:
+        return True  # skip
+
+    with monkeypatch.context() as m:
+        m.setattr(papis.tui.utils, "confirm", mock_confirm)
+        result = cli_runner.invoke(
+            cli, ["--to", "krish-doc", "Turing"], catch_exceptions=False
+        )
+        assert result.exit_code == 0, result.output
+
+    warning_messages = [r.message for r in caplog.records]
+    assert any("would nest inside another document" in msg for msg in warning_messages)
+    assert os.path.exists(old_folder)
+
+
+def test_to_target_already_exists_warns(
+    tmp_library: TemporaryLibrary,
+    caplog: pytest.LogCaptureFixture,
+    monkeypatch: MonkeyPatch,
+) -> None:
+    """Disk conflict warns."""
+    caplog.set_level(logging.WARNING)
+    cli_runner, _, doc = _setup_runner_db_doc()
+    old_folder = doc.get_main_folder()
+    assert old_folder
+    old_basename = os.path.basename(old_folder)
+
+    # Pre-create target dir with a subfolder matching the doc's name
+    target_dir = os.path.join(tmp_library.libdir, "target-dir")
+    os.makedirs(os.path.join(target_dir, old_basename))
+
+    def mock_confirm(msg: str, **kwargs: object) -> bool:
+        return True  # skip
+
+    with monkeypatch.context() as m:
+        m.setattr(papis.tui.utils, "confirm", mock_confirm)
+        result = cli_runner.invoke(
+            cli, ["--to", "target-dir", "Turing"], catch_exceptions=False
+        )
+        assert result.exit_code == 0, result.output
+
+    warning_messages = [r.message for r in caplog.records]
+    assert any("target already exists" in msg for msg in warning_messages)
+
+
+@pytest.mark.skipif(
+    os.environ.get("PAPIS_DATABASE_BACKEND") == "whoosh",
+    reason="Whoosh cannot query documents whose folders have been deleted",
+)
+def test_source_folder_missing_warns(
+    tmp_library: TemporaryLibrary, caplog: pytest.LogCaptureFixture
+) -> None:
+    """Missing source folder warns."""
+    caplog.set_level(logging.WARNING)
+    cli_runner, _, doc = _setup_runner_db_doc()
+    old_folder = doc.get_main_folder()
+    assert old_folder
+    shutil.rmtree(old_folder)
+
+    result = cli_runner.invoke(
+        cli, ["--to", "target", "Turing"], catch_exceptions=False
+    )
+    assert result.exit_code == 0, result.output
+
+    warning_messages = [r.message for r in caplog.records]
+    assert any("does not exist on disk" in msg for msg in warning_messages)
+
+
+def test_batch_skips_confirm_with_warnings(
+    tmp_library: TemporaryLibrary,
+    caplog: pytest.LogCaptureFixture,
+    monkeypatch: MonkeyPatch,
+) -> None:
+    """Batch mode skips confirm prompts, warns, and moves valid docs."""
+    caplog.set_level(logging.WARNING)
+    cli_runner, db, _ = _setup_runner_db_doc()
+
+    # there are two test_author docs, we delete one to trigger a warning
+    docs = [d for d in db.get_all_documents() if d["author"] == "test_author"]
+    assert len(docs) >= 2
+    warned_doc = docs[0]
+    valid_doc = docs[1]
+    warned_folder = warned_doc.get_main_folder()
+    assert warned_folder
+    valid_old_folder = valid_doc.get_main_folder()
+    assert valid_old_folder
+    valid_id = valid_doc["papis_id"]
+    shutil.rmtree(warned_folder)
+
+    confirm_called = False
+
+    def track_confirm(*args: object, **kwargs: object) -> bool:
+        nonlocal confirm_called
+        confirm_called = True
+        return True
+
+    with monkeypatch.context() as m:
+        m.setattr(papis.tui.utils, "confirm", track_confirm)
+        result = cli_runner.invoke(
+            cli, ["--batch", "--all", "test_author"], catch_exceptions=False
+        )
+        assert result.exit_code == 0, result.output
+
+    assert not confirm_called
+    warning_messages = [r.message for r in caplog.records]
+    assert len(warning_messages) == 1
+    assert "source folder does not exist on disk" in warning_messages[0]
+
+    # this works since that's the papis_id is the default folder name
+    expected = os.path.join(tmp_library.libdir, valid_id)
+    _assert_moved(valid_old_folder, expected)
+
+
+def test_multiple_docs_moved(tmp_library: TemporaryLibrary) -> None:
+    """Multiple docs moved into a directory each get their own subfolder."""
+    cli_runner, db, _ = _setup_runner_db_doc()
+
+    docs = [d for d in db.get_all_documents() if d["author"] == "test_author"]
+    assert len(docs) >= 2
+
+    old_folders: list[str] = []
+    for doc in docs:
+        folder = doc.get_main_folder()
+        assert folder is not None
+        old_folders.append(folder)
+
+    target_dir = os.path.join(tmp_library.libdir, "batch-target")
+    os.makedirs(target_dir)
+
+    result = cli_runner.invoke(
+        cli,
+        ["--to", "batch-target", "--all", "test_author"],
+        catch_exceptions=False,
+    )
+    assert result.exit_code == 0, result.output
+
+    # Both docs moved into target dir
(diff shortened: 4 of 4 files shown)


tuurep:
I'm very into this, thank you for the initiative! I should catch up with the history of it and then can offer to help :)

Primary motivation: I want moving docs between libraries to be as painless as possible.

jghauser:
Thanks @tuurep! That's an interesting use case especially because most commands work on an individual library. By default I think it would make sense to have `papis mv` always treat paths as relative to the library root folder. But then, we need some special functionality to handle your use case. Curious to know what others think!

alexfikl:
This would be great! I don't often need to move things around these days, but when I do it's always unwieldy :\ Some random thoughts below after reading the introduction.. with hopefully minimal repeating or contradicting myself from #826, but oh well :grin:.

> papis mv -t somefolder/ QUERY

I'm personally not a big fan of this.. I somehow never remember if I need to add a trailing forward slash in `rsync` :\ How about just adding a `--to` and an `--into`? Or add a `--mode=contents/directory`? Something explicit either way, haha.

It is nice and concise though, so if more frequent users of `papis mv` would prefer it, I'm fine with that.

> The --batch flag in general removes all need to interactively approve things.

I would recommend doing the lazy thing for now? i.e. skipping the document if it overwrites. The user can do a separate `--to some-unique-folder` after the batch operation completes to fix it. This is more or less how `papis add` works on duplicates too at least?

> Maybe we just live with the fact that the user will have to press 'ok' a bunch of times?

I don't agree. The `--batch` flag should really mean "no user interaction needed" here too. 

> As mentioned by @alexfikl in the original discussion, what should we do on failure: abort or continue?

Going forward, we may want to add some `--continue-on-error` type flag to these things so the user can choose if they want to just warn but run through the documents (good for e.g. `papis doctor --fix`) or error and stop (better for e.g. `papis update`). It does seem like both are valid use cases..

For now.. skip/continue on `--batch` and ask/abort otherwise? :grin: I don't know.. the important thing is not to lose any data here (i.e. don't move two documents into the same folder, basically losing the first one).

> The proposed syntax is inconsistent with `papis add`

Yeah, it would be nice to be consistent between the two, since they do seem to sort of have the same requirements regarding "document in (sub)folder placement". I mostly agree that the `--subfolder` / `--folder-name` flags look a bit quirky/non-standard, but they're very explicit!

> At the time I thought we should preserve `papis rename`, but now I think we should remove it. 

Agreed!

alexfikl:
> Something explicit either way, haha.

Maybe have the default be `--to somefolder/` (i.e. always write into the given directory) and add a `--rename` flag to switch to the `--to somefolder` mode? Although it does seem like the `--to somefolder` mode is the common case..

> I would recommend doing the lazy thing for now? i.e. skipping the document if it overwrites.

Could also just add a suffix `-a` like `papis add` does sometimes too? I guess we need to decide if on error/overwrite it's easier to: (a) go back and move all the folders that were skipped or (b) rename the moved folders if we don't want the suffix.

jghauser:
> > papis mv -t somefolder/ QUERY
> 
> I'm personally not a big fan of this.. I somehow never remember if I need to add a trailing forward slash in `rsync` :\ How about just adding a `--to` and an `--into`? Or add a `--mode=contents/directory`? Something explicit either way, haha.

I kinda like the `--to` and `--into` arguments. I mean we need an argument in any case, adding one more isn't so bad. We could even have `--to` with trailing `/` be treated like `--into` in a sort of poweruse mode. Though maybe that's just adding confusion for not much benefit. In any case, I think `--to` and `--into` are more descriptive than `--folder-name` and `--subfolder`, so I'd prefer moving to that.

I guess what remains unclear then is what to do when a user just uses `--into` -- should we rename the folder to `add-folder-name` or not? Cause my `papis mv QUERY` example implies that we should (no `--to` -> rename to `add-folder-name`), but that might be a bit unexpected...

> > Maybe we just live with the fact that the user will have to press 'ok' a bunch of times?
> 
> I don't agree. The `--batch` flag should really mean "no user interaction needed" here too.

Sorry, my bad. I meant that a careful user who wants to move a bunch of docs but doesn't want to use `--batch` will have to press 'ok' a bunch of times. So I agree with you!

alexfikl:
> Sorry, my bad. I meant that a careful user who wants to move a bunch of docs but doesn't want to use `--batch` will have to press 'ok' a bunch of times. So I agree with you!

Ah, yeah, that's fine! I even have an `alias rm="rm -i"` to always ask me, which I have yet to regret, haha.

jghauser:
I just realised that kiike's original plan of having tab completion as part of the mv procedure (you run `papis mv something something` and then get prompted for the destination folder in a way that allows tab completion) isn't really possible with the new plan. I think it makes more sense to use `add-folder-name` by default and provide shell completion when completing the relevant arguments (though Papis' shell completion is a bit iffy, so I'm not sure how well it'd work).

pmiam:
I support slowly moving towards UI overhaul. Some comments:

Julian Hauser ***@***.***> writes:

> Note that @pmiam proposed globs &#39;*&#39; instead of trailing `/`. I think that could lead to issues with regular shell globbing. Trailing `/` don&#39;t present that problem and are how e.g. `rsync` does it.

Good point about the globs. I think a trailing slash is not so clear,
but there is definitely precedent for it especially in file transfer.

> I kinda like the `--to` and `--into` arguments. I mean we need an argument in any case, adding one more isn't so bad.

I don't like this. I really disagree with leaning on option flags as a
design crutch. It is difficult to describe, but I have a strong
intuition for certain flags, and unfortunately papis does not conform
with my own "sense" in most places. I use papis every day and still
struggle to speak its language. I need to work on articulating this.

Alex Fikl ***@***.***> writes:

> I would recommend doing the lazy thing for now? i.e. skipping the document if it overwrites.

This may or may not be lazy, but I think it's the right thing to do.
With a loud and clear warning, of course.

Julian Hauser ***@***.***> writes:

> - The `--batch` flag in general removes all need to interactively approve things. However, we might want to be able to move a ton of docs at once but still be warned about overwrites. What should we do? Having `--batch` for this but with overwrite warnings is inconsistent with other uses of `--batch`. Maybe we just live with the fact that the user will have to press &#39;ok&#39; a bunch of times?

The --batch should not do the job of suppressing safety features.
I think adding --force to perform overwrites is the next step.

> At the time I thought we should preserve `papis rename`, but now I think we should remove it. It&#39;s confusing (since it&#39;s unclear whether it acts on folders or files, see #894), and its current functionality should be included it `papis mv`. Renaming of files (and notes) should be done by `papis update`.

Strongly Agree. I agree with the comments in #894 to use papis update as
the enforcer of info.yaml.

> - The proposed implementation has no way of moving folder into another folder and renaming it to the `add-folder-name` option
>   - You would need to specify the format pattern explicitly, e.g. `papis mv -t somefolder/{doc[author]}-{doc[title]}`
>   - I guess that&#39;s fine?

I think especially with jinja templates this is very cumbersome. I don't
have a better suggestion, but maybe we can find something coherent by
trying to address inter-library transfers together with this scenario.

> I would prefer adding `-t/--to` to `papis add` and maybe eventually deprecate the old arguments

Yes please.

jghauser:
Another idea: why don't we copy `mv` behaviour? If a document is `papis mv`ed to a directory that doesn't exist, it's a rename (to). If a document is `papis mv`ed to a directory that does exist, it's a move (into). We could complement this with a warning/prompt when a user tries to `papis mv` a document into a directory that is another document's main dir (e.g. a dir containing an `info.yaml`).

Limitations:
- Can't so easily handle the case of moving current directory *into* new directory. Could solve this by adding trailing slash like `mv` does (this is quite rare and maybe Alex would be fine with it) and/or the user could also just specify the whole path incl the main folder name.
- Can't do force overwrite of existing folder (but that's not desirable anyway).

Advantage: Our command is called `papis mv` so it would actually be very neat if it just copied the well-known `mv` functionality.

jghauser:
I just committed a first version, implementing most of the suggestions discussed (using the POSIX `mv` semantics). I definitely haven't tested it thoroughly yet, so don't test it on your main libraries. I would be very interested in any feedback/suggestions you may have!

jghauser:
If anyone has any thoughts about the approach this is following, do let me know. In particular I'd like to know whether 
- following unix `mv` semantics makes sense
- whether you have any thoughts on the various NOTEs in the source code
- and if there's anything else you would like to see :)

alexfikl:
> If anyone has any thoughts about the approach this is following, do let me know. In particular I'd like to know whether

This totally dropped off my radar, sorry. I'll take a look tomorrow :grin:

jghauser:
> This totally dropped off my radar, sorry. I'll take a look tomorrow 😁

No worries at all and no rush! It's been a busy time here anyway.

papis/commands/mv.py
```diff
@@ -1,12 +1,50 @@
 """
-This command can be used to move a document to a new folder.
+The ``mv`` command is used to move document folders around your library (or to
+another library).
 
-Command-line interface
+It will (except when run with ``--all``) bring up the picker with a list of
+documents that match the query. You can then select the document or documents
+you'd like to move. The command works similar to the POSIX mv command. Hence,
```
alexfikl:
```suggestion
you'd like to move. The command works similar to the POSIX `mv <https://en.wikipedia.org/wiki/Mv_(Unix)>`__ command. Hence,
```
?

papis/commands/mv.py
```diff
@@ -1,12 +1,50 @@
...
+when the specified target folder exists (or ends with a `/`), the document's
```
alexfikl: I just tried
```sh
> mkdir folder1
> touch folder1/info.yml
> mv folder1 folder2/
```
and it doesn't move the folder to `folder2/folder1` (as expected). If we stick with this, we should say we use `rsync` semantics or something instead of `mv`?

papis/commands/mv.py
```diff
@@ -1,12 +1,50 @@
...
+current main folder is moved into the target folder. If the specified target
+folder doesn't exist, then the current folder is renamed to the target folder
+name. Folder names are cleaned, so that various problematic characters (white
+spaces, punctuation, capital letters, and some others) are avoided.
```
alexfikl: Just link to `doc-paths-extra-chars` and friends?

papis/commands/mv.py
```diff
@@ -1,12 +1,50 @@
...
+
+Examples
+^^^^^^^^
+
+- Query documents by author "Rick Astley" and move some of them. After picking
+  the relevant documents, they will be renamed to their default names
+  specified by the `add-folder-name` option.
```
alexfikl:
```suggestion
  specified by the :confval:`add-folder-name` option.
```

papis/commands/mv.py
```diff
@@ -1,12 +1,50 @@
...
+
+   .. code:: sh
+
+        papis mv author:"Rick Astley"
+
+- You can use ``--to`` to pass the name of the target folder. Here we'll move a
+  document by the author "Rick Astley" to a folder named ``rick_astley``:
```
alexfikl: Maybe say what happens if the user selects more documents and they all get moved to `rick_astley`?

papis/commands/mv.py
```diff
@@ -16,96 +54,224 @@
...
 
 
-def run(document: Document,
-        new_folder_path: str,
-        git: bool = False) -> None:
-    from papis.document import describe
+def run(
+    document: Document,
+    doc_folder_new: str,
+    doc_folder_old: str,
```
alexfikl: Isn't this available in `document.get_main_folder()`?

papis/commands/mv.py
```diff
@@ -16,96 +54,224 @@
...
+    batch: bool,
+    target_lib_dir: str,
+    target_library: str,
```
alexfikl: These should have some defaults? If not, this function takes a whole lot of mandatory arguments :grin:

papis/commands/mv.py
```diff
@@ -16,96 +54,224 @@
...
+    "-t",
+    "--to",
+    help="Target path. Uses POSIX mv semantics.",
+    default=lambda: papis.config.getformatpattern("add-folder-name"),
+    type=papis.cli.FormatPatternParamType(),
+)
+@click.option(
+    "--target-library",
+    help="Target library",
+    default=papis.config.get_lib_name,
```
alexfikl:
```suggestion
    default=lambda: papis.config.get_lib_name(),
```

papis/commands/mv.py
```diff
@@ -16,96 +54,224 @@
...
         return
 
-    document = documents[0]
+    target_lib_dir = os.path.expanduser(
+        papis.config.getstring(key="dir", section=target_library)
+    ).rstrip(os.path.sep)
 
-    lib_dir = os.path.expanduser(papis.config.get_lib_dirs()[0])
+    if not git and os.path.exists(os.path.join(target_lib_dir, ".git")):
+        git = papis.tui.utils.confirm("Add the changes to git?")
```
alexfikl: Not sure I like this? First because it's the only command that would check this, but also:
* Personally, I have all libraries in a git repo, but `git` is turned off because I prefer making the commits myself (usually in bulk).
* I would expect this to work the other way mainly, i.e. if `git = True` and there is no git repo, it should just set it to False for me.
* The `.git` directory could be in a parent.

papis/commands/mv.py
```diff
@@ -16,96 +54,224 @@
...
 
-    completer = prompt_toolkit.completion.PathCompleter(
-        only_directories=True,
-        get_paths=lambda: [lib_dir]
-    )
+    moves = []
+    for document in documents:
+        main_folder_old = document.get_main_folder()
+        if not main_folder_old:
+            raise DocumentFolderNotFound(papis.document.describe(document))
```
alexfikl: This should be a `os.path.exists`? I think all the documents in the database will have a folder, but someone might have deleted it in the meantime.

papis/commands/mv.py
```diff
@@ -16,96 +54,224 @@
...
-                    f"Current directory: ({document.get_main_folder_name()})\n >"),
-                completer=completer,
-                complete_while_typing=True
-            ))
-    except Exception as exc:
-        logger.error("Failed to choose directory.", exc_info=exc)
-        return
+        # Determine operation mode
+        if isinstance(to, FormatPattern):
+            folder_name_format = to
```
alexfikl: Have you checked this? I think the `to` is always a `FormatPattern` because of `FormatPatternParamType`.

papis/commands/mv.py
```diff
@@ -16,96 +54,224 @@
...
+            if is_move_into:
+                main_folder_new = os.path.join(to, os.path.split(main_folder_old)[1])
+
+                folder_name_format = FormatPattern(None, main_folder_new)
+            else:
+                folder_name_format = FormatPattern(None, to)
+
+        main_folder_new = papis.paths.get_document_folder(
+            document, target_lib_dir, folder_name_format=folder_name_format
+        )
```
alexfikl: I'm not sure I follow the logic here. Maybe a little comment?

papis/commands/mv.py
```diff
@@ -16,96 +54,224 @@
...
+
+        # Skip if source and destination are the same
+        if main_folder_new.rstrip(os.path.sep) == main_folder_old.rstrip(os.path.sep):
```
alexfikl: [`os.path.samefile`](https://docs.python.org/3/library/os.path.html#os.path.samefile)?

papis/commands/mv.py
```diff
@@ -16,96 +54,224 @@
...
+                        "Skipping document '%s'.",
+                        papis.document.describe(document),
+                    )
+                    continue
+            else:
+                logger.info(
+                    "Skipping document '%s' in batch mode.",
+                    papis.document.describe(document),
+                )
+                continue
```
alexfikl: Does this need to be so complicated? I would expect just a "this directory is not empty, are you sure you want to write into it?".

papis/commands/mv.py
```diff
@@ -16,96 +54,224 @@
...
-    run(document, new_folder, git=git)
+    for document, main_folder_old, main_folder_new in moves:
+        run(
+            document,
+            main_folder_new,
+            main_folder_old,
+            batch=batch,
+            git=git,
+            target_lib_dir=target_lib_dir,
+            target_library=target_library,
```
alexfikl: Just pass `target_library` here? It seems like the rest can be obtained from that and it just makes the interface clunky.

papis/commands/mv.py
```diff
@@ -16,96 +54,224 @@
...
+        if os.path.exists(target_info_path):
+            logger.warning(
+                "Target directory '%s' contains an '%s' and appears to be "
+                "another document's folder.",
+                main_folder_new_parent,
+                info_name,
+            )
+            # TODO: not sure what to do here re handling of batch.
+            # Maybe we do want 'batch' and 'force'?
+            # - batch: run without prompts but safely
```
alexfikl: What does "safely" mean?

In general, I think it makes sense to have a `--batch` and a `--force` flag if they work on different issues (i.e. not just have `--force` mean "REALLY --batch"), e.g.
* `--batch` does not ask for user confirmation, i.e. if I tell it to overwrite/remove it better overwrite/remove.
* `--force` should ignore errors and continue as best it can (this can mean unrolling changes and going to the next document).

Seeing as we keep hitting this, we probably should pick a command and update it to have both (I'd say in a different PR?). That would allow us to document what they mean: e.g., what do we ask confirmation for, which errors are force-able, what still causes a hard exit, etc.

papis/commands/mv.py
```diff
@@ -16,96 +54,224 @@
...
+                return
+        else:
+            return
 
-    if git:
-        from papis.git import mv as git_mv
-        git_mv(folder, new_folder_path)
+    if target_lib_dir == source_lib_dir:
+        # TODO: there was some mention of committing if tracked? can i just do
+        #       the below or do i need to check if it's tracked?
```
alexfikl: Pretty much everyone just commits with no checks if `git` is true, so it should be ok for now.

There's a bit of extra complication here, because the source library could be tracked but the target could not be, but oh well :\

papis/commands/mv.py
```diff
@@ -16,96 +54,224 @@
...
+                doc_folder_new,
+                f"Move '{papis.document.describe(document)}'",
+            )
+        else:
+            shutil.move(doc_folder_old, doc_folder_new)
     else:
-        import shutil
-        shutil.move(folder, new_folder_path)
+        if git:
+            # TODO: should I add a remove+commit helper function to papis.git?
```
alexfikl: I think this is fine.

papis/commands/mv.py
```diff
@@ -16,96 +54,224 @@
...
+            papis.git.commit(
+                source_lib_dir, f"Remove '{papis.document.describe(document)}'"
+            )
+
+        shutil.move(doc_folder_old, doc_folder_new)
 
-    from papis.database import get_database
-    db = get_database()
-    db.delete(document)
+        if git and os.path.exists(os.path.join(target_lib_dir, ".git")):
```
alexfikl: I'm not sure these checks for `.git` make much sense: the `.git` folder could be in some parent folder, more libraries could be in the same git repository, etc.

I would leave this for a separate PR, since it would be better to update all the uses. Ideally we can just do this check in the `papis.git` functions and raise an error if we're not in a git repository.

alexfikl (commented):
This looks a lot more usable now! :grin: 

I left a bunch of nitpicks (that can mostly wait) and tried to at least start a discussion on the various TODOs. I think some of them are a bit large in scope, so maybe best left for a future PR?

papis/commands/mv.py
jghauser: Done thanks!

papis/commands/mv.py
jghauser: You are totally right, this is my bad. I think we should stick to `mv` functionality. It's well-known and easy to understand (quite unlike the confusing `rsync` final `/` thing that trips me up too).

Now, that entails it's not possible to `mv` a folder into a folder that doesn't yet exist. Right now my thinking is that that's fine, and Papis doesn't need to add some special functionality for it. Either you're moving a folder into some custom location (in which case you're probably aware of whether the folder exists, and you can just create it with `mkdir` if needed). Or you're using format patterns, in which case you should never encounter the problem. If the format pattern leads to moving 'folder1' to 'folder2/folder3' this will work even if 'folder2' doesn't exist. I don't think it's worth adding an extra 'move into missing folder' flag. What do you think?

alexfikl: Agreed! We can always add something like that later if someone needs it and has a good idea for the UX.

papis/commands/mv.py
jghauser: Thanks!

papis/commands/mv.py
jghauser: Good idea, thanks!

papis/commands/mv.py
jghauser: What exactly do you mean here? Above I get the (old) main folder with `document.get_main_folder()` and here I pass this on to `run()`. Are you saying I should just get the main folder again in `run()`? I guess I was trying to avoid running that twice, but happy to change it if you think that's better :).

papis/commands/mv.py
jghauser: Gotcha! Didn't know about the subtle difference here.

papis/commands/mv.py
jghauser: I guess that would make it easier to call from elsewhere?

papis/commands/mv.py
jghauser: The reason here was the same as above with `main_folder_old`, that is not to have to get things twice. But I can see it makes the interface clunkier without much real benefit.

papis/commands/mv.py
jghauser: Yup, you're right!

papis/commands/mv.py
jghauser: That makes sense. I checked other commands (e.g. `papis add`) and there the explanation of the git flag is:
```
    :param git: Whether or not to ask user for committing before adding,
        in the case of course that the library is a git repository.
```
And the code seems to just attempt to do a git operation without checking if there's a `.git` directory. This fails with a message from git and the document is added without doing any git stuff. I think I'll just do the same here.

papis/commands/mv.py
jghauser: We get some issues with types if we don't check if `main_folder_old` exists (because `get_main_folder()` returns `str | None`), but your consideration is important too. Should I just do:

```python
        if not main_folder_old or not os.path.exists(str(main_folder_old)):
```

papis/commands/mv.py
jghauser: It doesn't appear to be. It's only in the default case (`default=lambda: papis.config.getformatpattern("add-folder-name"`) that we're in this branch. Otherwise `to` is a string. I'm not quite sure I understand how exactly all of this is working or if there's a better way of setting things up. Do let me know :smile:

papis/commands/mv.py
jghauser: Yeah that might be useful :sweat_smile:! I think I also found some issues with this as well...

papis/commands/mv.py
jghauser: I don't think that would be specific enough. We quite commonly want to move folders into non-empty folders. What we don't want is move a folder into some other document's folder. And that's the case this is trying to catch. I think it's a reasonable safety feature. What do you think? Should we ditch all such checks before moving (just checking if the folder is non-empty would create too much noise I think)?

papis/commands/mv.py
jghauser: Ok, it makes sense to move this to a different PR and find some good consistent way of doing this across the board. For now, I'm leaving `batch` and making it so that it doesn't move folders into another document's folders. This is the 'safe' that I had in mind (similar to how `papis add --batch` will not add a document if another matching one was found).

papis/commands/mv.py
jghauser: Ok, gotcha. And yeah, there isn't really an elegant way of handling target libraries, but I think as long as it works decently, it'll be useful :)

papis/commands/mv.py
jghauser: I removed all the checks for a `.git` folder!

papis/commands/mv.py
alexfikl: I meant that passing both document and `doc_folder_old` would allow things to get out of sync. What happens if `document.get_main_folder() != doc_folder_old`? Will it move some other document instead?

papis/commands/mv.py
alexfikl: Actually your version may work too? This just makes it obvious that we want to make the call lazy and it matches the rest of the commands.

papis/commands/mv.py
alexfikl: That looks good, yeah!

papis/commands/mv.py
alexfikl: Hm.. It should be converted to a format pattern because of that `type=FormatPatternType` thing in the option definition. I'll check it out!

papis/commands/mv.py
alexfikl: Hm, fair enough. Maybe refactor these to some helper functions? Like `if is_document_folder(folder): ...`.

papis/commands/mv.py
alexfikl: Those are pretty cheap calls in general, so I wouldn't worry too much about it (at least not without actual benchmarking).

papis/commands/mv.py
alexfikl: Sounds good!

papis/commands/mv.py
jghauser: Ok, I'll do that!

jghauser:
i'm working on `papis mv` again, which is really twisting my mind lol. i think the current version has some weirdnesses because we detect whether target folders have `info.yaml`s in a first loop, but then in a second loop we might still move a folder into a folder with an `info.yaml` (because they've now been created). we ask whether to overwrite such folders, but it's not at all obvious that saying 'no' here will move the folder inside the target folder.

i see two ways of fixing this. the first is to move all the checks to the second loop and the second to check before moving anything. i know that only `papis update` so far checks upfront but i'm kinda tempted to do it here too (because mv'ing is such an important thing to handle in a way that isn't confusing). it would also avoid the problem of what to do when `shutil.move` fails (should we show an error message that might get lost in stdout and carry on or should we abort with the error after having moved just a few files). what do you think?

alexfikl:
> i see two ways of fixing this. the first is to move all the checks to the second loop and the second to check before moving anything. i know that only `papis update` so far checks upfront but i'm kinda tempted to do it here too (because mv'ing is such an important thing to handle in a way that isn't confusing). it would also avoid the problem of what to do when `shutil.move` fails (should we show an error message that might get lost in stdout and carry on or should we abort with the error after having moved just a few files). what do you think?

Yeah, I agree we should check upfront like in `papis update`. I would expect this to (sort of repeating what you said to make sure I understand :D)
1. Construct a list of full paths to move to. For each document
	1.1. Check if that path is valid, i.e. it does not exist + it's not overwriting some existing document + it's still inside the library + whatever else comes to mind. This should also check against the document paths before it, not just ones on the filesystem.
	1.2. If it's ok, add it to the list of movable documents. 
	1.3. This could abort the whole thing if `--batch` like `papis update`? Or just add a `-suffix` thing like `papis add` does? Guess it depends on the kind of error the check finds?
2. Go through the list of all the ones that are ok and move them. 
3. Missed anything? :D

jghauser:
Third time's the charm! I've ended up changing things quite a lot to catch all sorts of warnings in the the first loop so that we can be confident in correctly and unsurprisingly moving documents once we enter `run()`. I'm quite happy with how things turned out. I think I've addressed all sorts of edge cases and the things you (@alexfikl) pointed out in previous reviews. I also think the flow should be much easier to follow now. Let me know what you think :)

EDIT: Regarding `--batch` functionality. We collect warnings in `cli()`. With `--batch`, we do not prompt whether we should continue and simply skip moving any problematic docs. There is no way of executing problematic moves. I feel that given the delicate nature of moves, this is a sensible thing to do here. Hmm... Though now I'm thinking that maybe I do need to be a bit smarter. Currently the code checks whether a target folder with the same name exists, but it simply assumes that this is the same doc. But that's not a given... I'll think about it :sweat_smile: 

EDIT2: We actually *are* a bit smarter since we compare the doc's old folder name with the new folder name to check if it's same doc. I guess that still leaves open the question you raised regarding what to do when we find a folder with the same name (which isn't the doc's old folder). At the moment, we warn (`target already exists`) and don't move the doc. But your idea of adding a suffix is also sensible. Given how rarely this is going to happen, I think maybe erring on the side of caution (and not moving) might be the better call? But I don't have a strong opinion here...

EDIT3: One more difficulty I came across. Currently `papis mv` assumes that all paths in the library are good paths (in the sense that they have been created with Papis and are slugified etc). This shows when we immediately format the user's `--to` argument and means that it's impossible for a user to move a doc into a folder they created that isn't slugified (e.g. `some_subfolder` with underscore, which we'll slugify to `some-subfolder`). Is this ok? I suppose `papis mv` is papis specific moving, so the assumption is ok. Users should use plain `mv` if they want something else?

papis/commands/mv.py
```diff
@@ -1,111 +1,350 @@
...
+Examples
+^^^^^^^^
+
+- Query documents by author "Rick Astley" and move some of them. After picking
+  the relevant documents, they will be renamed to their default names
+  specified by the :confval:`add-folder-name` option.
+
+   .. code:: sh
+
+        papis mv author:"Rick Astley"
```
alexfikl, line 37: So move acts like rename by default? Since we mention "Unix mv" semantics, I would assume that this this does nothing and I need to do a `[papis] mv source destination` sort of thing to move.

papis/commands/mv.py
```diff
@@ -1,111 +1,350 @@
...
+        papis mv --to rick-astley author:"Rick Astley"
+
+  If the folder "rick-astley" exists, then the document's folder will be moved into it.
+  If, say, the document's folder is currently "rick-s-the-best", then the new folder
+  will be "rick-astley/rick-s-the-best". If the folder "rick-astley" doesn't exist,
+  then the document's folder will be renamed to that so that the new folder becomes
+  "rick-astley".
+
+  If you select multiple documents to be moved to the same target folder, Papis will
+  warn you and refuse to move the documents.
```
alexfikl: So I can't move multiple documents to a given subfolder? That sounds like it's in a bit of a conflict with "moving into it if it exists".

papis/commands/mv.py
```diff
@@ -1,111 +1,350 @@
...
+   .. code:: sh
+
+        papis mv --to "{doc[year]}" --all author:"Rick Astley"
+
+  Note that the ``--all`` flag means that the picker isn't shown and all
+  documents matching the query will be moved.
+
+- You can use the ``--batch`` flag to skip all warnings. You will not be prompted
+  if there are problematic moves and all unproblematic moves will happen
+  automatically. The command will only abort if an error occurs while moving a document.
```
alexfikl: This is also a bit confusing.. seems to be saying that "unproblematic moves" will still happen, but then it aborts? What if there are "unproblematic moves" after the one that caused an error?

papis/commands/mv.py
```diff
@@ -1,111 +1,350 @@
...
 
 
-def run(document: Document,
-        new_folder_path: str,
-        git: bool = False) -> None:
-    from papis.document import describe
+def _is_document_folder(path: Path) -> bool:
+    """Check whether `path` is an existing document folder."""
+    info_name = papis.config.getstring("info-name")
+    return path.joinpath(info_name).exists()
```
alexfikl: This looks like it should go into `papis.document`?

papis/commands/mv.py
```diff
@@ -1,111 +1,350 @@
...
+
+
+def _get_target_library_dir(target_library: str) -> Path:
+    """Get the directory path for the given library."""
+    return Path(papis.config.getstring(key="dir", section=target_library)).expanduser()
```
alexfikl: This looks like it should be `papis.config.get_lib_from_name().path`?

papis/commands/mv.py
```diff
@@ -1,111 +1,350 @@
...
+    doc_folder_new: str,
+    target_library: str | None = None,
+    git: bool = False,
+) -> None:
+    """Move a document's folder to *doc_folder_new*.
+
+    :param document: The document to move.
+    :param doc_folder_new: The target path for the document's folder.
+    :param target_library: Name of the target library. If None, uses the current
+        library.
```
alexfikl: Is the `doc_folder_new` supposed to be a relative path? the docs should mention that a bit. It should also mention if it's relative to the `target_library`.

papis/commands/mv.py
```diff
@@ -1,111 +1,350 @@
...
+    :param git: If True, record the move in git history.
+    :returns: None
```
alexfikl:
```suggestion
```
The type annotation is clear enough.

papis/commands/mv.py
```diff
@@ -1,111 +1,350 @@
...
+    :raises DocumentFolderNotFound: If the document has no main folder on disk.
+    :raises FileNotFoundError: If the source folder does not exist.
+    :raises FileExistsError: If the target folder already exists. This will also be
+        raised if the target folder is the same as the source folder.
+    """
+    if target_library is None:
+        target_library = papis.config.get_lib_name()
+
+    source_lib_dir = Path(papis.config.getstring(key="dir")).expanduser()
```
alexfikl: This should also go through `papis.config.get_lib().paths`.

papis/commands/mv.py
```diff
@@ -1,111 +1,350 @@
...
-        from papis.git import mv as git_mv
-        git_mv(folder, new_folder_path)
+    main_folder_old = Path(doc_folder_old)
+    main_folder_new = Path(doc_folder_new)
+
+    # sanity check even if covered in `cli()`
+    if not main_folder_old.exists():
+        raise FileNotFoundError(f"Source folder does not exist: '{main_folder_old}'")
+
+    # sanity check even if covered in `cli()`
```
alexfikl:
```suggestion
```
This function can also be called outside of the `cli`, so this is actually very good, not just a sanity check :grin:

papis/commands/mv.py
```diff
@@ -1,111 +1,350 @@
...
+    main_folder_new.parent.mkdir(
+        parents=True,
+        exist_ok=True,
+        # NOTE: 0o755 is default, but explicitness is needed to avoid type error
+        mode=papis.config.getint("dir-umask") or 0o755,
+    )
+    if target_lib_dir == source_lib_dir:
+        if git:
+            papis.git.mv_and_commit_resource(
+                str(main_folder_old),
```
alexfikl, line 259: All this `str -> Path -> str` is not great. I agree that we want to use `pathlib` going forward, but that's probably a separate cleanup PR?

papis/commands/mv.py
```diff
@@ -1,111 +1,350 @@
...
-        shutil.move(folder, new_folder_path)
+        shutil.move(doc_folder_old, main_folder_new)
+
+        if git:
+            papis.git.remove(str(source_lib_dir), str(main_folder_old))
+            papis.git.commit(
+                str(source_lib_dir), f"Remove '{papis.document.describe(document)}'"
+            )
+
+            # NOTE: this will fail when the target library isn't git tracked
```
alexfikl: All of these will fail if the libraries aren't git tracked :grin: It seems like they just use `subprocess.run(...)`, so they also fail silently (right?).

papis/commands/mv.py
```diff
@@ -1,111 +1,350 @@
...
+    target_db = papis.database.get(library_name=target_library)
 
-    document.set_folder(new_document_folder)
-    db.add(document)
+    document.set_folder(str(main_folder_new))
+
+    target_db.add(document)
+
+    logger.info(
+        "Moved document '%s' to '%s'.",
```
alexfikl:
```suggestion
        "Moved document '%s' to '%s' (library '%s').",
```
?

papis/commands/mv.py
```diff
@@ -1,111 +1,350 @@
...
         logger.warning(no_documents_retrieved_message)
         return
 
-    document = documents[0]
+    target_lib_dir = _get_target_library_dir(target_library)
 
-    lib_dir = os.path.expanduser(papis.config.get_lib_dirs()[0])
+    moves: list[
+        tuple[Document, Path, Path]
+    ] = []  # (document, main_folder_old, main_folder_new)
```
alexfikl: Make a little `NamedTuple` for this?

papis/commands/mv.py
```diff
@@ -1,111 +1,350 @@
...
-    except Exception as exc:
-        logger.error("Failed to choose directory.", exc_info=exc)
-        return
+            continue
+
+        target_abs_path = Path(
+            papis.paths.get_document_folder(
+                document, target_lib_dir, folder_name_format=to
+            )
+        )
```
alexfikl, line 378: From what I can tell `to` is the `add-folder-name` from the current library? If I'm trying to move this to a different library, I'd expect it to use the format specified there.

papis/commands/mv.py
```diff
@@ -1,111 +1,350 @@
...
+    for group in conflict_groups.values():
+        if len(group) == 1:
+            deduplicated_moves.append(group[0])
+        else:
+            for doc_i, main_folder_old_i, main_folder_new_i in group:
+                warnings.append((
+                    doc_i,
+                    f"duplicate target folder with {len(group) - 1} other document(s). "
+                    f"Skipping move: {main_folder_old_i} → {main_folder_new_i}.",
+                ))
```
alexfikl: This should all be folded into the for loop above by making `moves` be a `dict[main_folder_new, (document, main_folder_old)]` to begin with?

papis/commands/mv.py
```diff
@@ -1,111 +1,350 @@
...
+
+    if warnings:
```
alexfikl: Not quite sure I understand why these warnings are printed after we go through all the documents?

papis/commands/mv.py
```diff
@@ -1,111 +1,350 @@
...
-    if not os.path.exists(new_folder):
-        logger.info("Creating path '%s'.", new_folder)
-        os.makedirs(new_folder, mode=papis.config.getint("dir-umask") or 0o666)
+        if not batch:
+            if not papis.tui.utils.confirm(
+                f"{len(warnings)} warning(s) listed above. "
+                f"Continue with the {len(deduplicated_moves)} valid move(s)?",
+                yes=False,
+            ):
+                return
```
alexfikl: I'm not sure I like this. Again, to the familiar `mv` :grin:, doing 
```
mv --interactive sources destinationdir
```
will ask one by one if it's ok to move (`mv: overwrite '...'?`). I would expect a similar behavior here too in non-batch mode.

I feel like we're going back and forth a bit here.. but it's been a long time 🤦

alexfikl (commented):
Took another look at this and left some comments. Most of them are nitpicks + doc clarifications, but there are some things that I still don't like (that I've maybe mentioned before, sorry about that 🤦)

>  Regarding --batch functionality. We collect warnings in cli(). With --batch, we do not prompt whether we should continue and simply skip moving any problematic docs. 

I don't like the collecting warnings, but the rest sounds good, i.e. it should just skip.

> Currently the code checks whether a target folder with the same name exists, but it simply assumes that this is the same doc.

I'm not sure I understand why that would matter? If the target exists and it's a doc folder, we should skip it regardless. No?

> But your idea of adding a suffix is also sensible.

I'm not 100% into that, it has to be said :grin: For `papis add` it makes sense because you're just adding one document. Here, we could potentially move a lot of them and just make a typo in the format string.. and BOOM, we get a load of `2020-a, 2020-b, 2020-c, ...`. It's not the end of the world, because the same query can match them again and move them to the correct location, but still..

> One more difficulty I came across. Currently papis mv assumes that all paths in the library are good paths (in the sense that they have been created with Papis and are slugified etc).

Hm, that's a good point. Personally, I see two scenarios:
1. We move documents in a library using something like
```
papis mv --to "..." QUERY
```
so it seems natural to use the `add-folder-name` and slugify things. We're very upfront about renaming things in all the docs, so it shouldn't come as a surprise.

This could fail if the format is `sub_folder/{doc[...]}-...`, but `papis add` also normalizes all those paths, so that document shouldn't be there to begin with if it was added with Papis?

2. We move to another library. This should work pretty much the same, i.e. the library path can be whatever it is and we don't slugify it. You can also do something like
```
papis --lib /my/special/directory-for-books mv --to ".." QUERY
```
to move things to some non-standard path? Did I miss anything?

jghauser:
Thanks a lot for the comments! For my answers, see below. The only really open question concerns the first point.

> > Regarding --batch functionality. We collect warnings in cli(). With --batch, we do not prompt whether we should continue and simply skip moving any problematic docs.
>
> I don't like the collecting warnings, but the rest sounds good, i.e. it should just skip.

How do you mean this?
- First, do you not like collecting warnings at all or not in `--batch` mode? My reasoning for collecting warnings is that (a) it's hard to know what exactly the problematic moves are if they're mixed in with all the other messages in stdout and (b) we skip moves that would lead to 2 docs having the same folder, but we cannot know which docs are like that before collecting all moves. if we prompted first, it's possible that the user then says 'yes' to a move that later is skipped for sharing the same target folder as another one, which is confusing.
- Second, is your concern that we should be able to say yes/no to each problematic move individually or that warnings are presented all at the same time? I do think collecting warnings is better UX (when you get prompted you know that once you've handled all prompts you're done), but I can see it diverges somewhat from other commands.

> > Currently the code checks whether a target folder with the same name exists, but it simply assumes that this is the same doc.
>
> I'm not sure I understand why that would matter? If the target exists and it's a doc folder, we should skip it regardless. No?

we should skip in either case, but if it's the same doc, then it's a no-op while if it's a different doc, then it's a problematic move that should give rise to a warning. but as my EDIT2 mentions, we do check quite specifically if it's the same doc (by comparing the doc's old folder name to the new one).

> > But your idea of adding a suffix is also sensible.
>
> I'm not 100% into that, it has to be said 😁 For `papis add` it makes sense because you're just adding one document. Here, we could potentially move a lot of them and just make a typo in the format string.. and BOOM, we get a load of `2020-a, 2020-b, 2020-c, ...`. It's not the end of the world, because the same query can match them again and move them to the correct location, but still..

that makes sense. so, let's just warn as the code does at the moment.

> > One more difficulty I came across. Currently papis mv assumes that all paths in the library are good paths (in the sense that they have been created with Papis and are slugified etc).
>
> Hm, that's a good point. Personally, I see two scenarios:
>
>     1. We move documents in a library using something like
>
>
> ```
> papis mv --to "..." QUERY
> ```
>
> so it seems natural to use the `add-folder-name` and slugify things. We're very upfront about renaming things in all the docs, so it shouldn't come as a surprise.
>
> This could fail if the format is `sub_folder/{doc[...]}-...`, but `papis add` also normalizes all those paths, so that document shouldn't be there to begin with if it was added with Papis?
>
>     2. We move to another library. This should work pretty much the same, i.e. the library path can be whatever it is and we don't slugify it. You can also do something like
>
>
> ```
> papis --lib /my/special/directory-for-books mv --to ".." QUERY
> ```
>
> to move things to some non-standard path? Did I miss anything?

Yeah I think your thinking here is right. I added another sentence to the docs to drive the implication home.

Gonna work through the rest of your comments now :smile:

papis/commands/mv.py
jghauser: Good catch. Added the following to explain default target folders:

```
Unlike Unix ``mv``, if a
target folder isn't given, it defaults to the :confval:`add-folder-name` pattern,
so ``papis mv QUERY`` renames matching documents to their default folder names.
```

alexfikl:
> How do you mean this?
> 
> * First, do you not like collecting warnings at all or not in `--batch` mode? My reasoning for collecting warnings is that (a) it's hard to know what exactly the problematic moves are if they're mixed in with all the other messages in stdout and (b) we skip moves that would lead to 2 docs having the same folder, but we cannot know which docs are like that before collecting all moves. if we prompted first, it's possible that the user then says 'yes' to a move that later is skipped for sharing the same target folder as another one, which is confusing.

(a) That's fair, but then we should maybe silence the other messages? What else is shown during the moves here? Since we're just going through the documents, I would expect we have full control of what gets displayed. If it feels noisy, just put some of them in `logger.debug`?

(b) I'm not sure I follow. If document doc[0] wants to move to folder0 and doc[1] also wants to move to folder0, we'll know because folder0 is already in the list of new folders (i.e. you can keep a moving version of that `conflict_groups` dict and just not add a new folder to `moves` if a key already exists). The way it currently works with 3 loops to check gather+deduplicate can just be one loop, I think.

> * Second, is your concern that we should be able to say yes/no to each problematic move individually or that warnings are presented all at the same time? I do think collecting warnings is better UX (when you get prompted you know that once you've handled all prompts you're done), but I can see it diverges somewhat from other commands.

Yeah, I would expect that the command asks me what to do for each problematic move. That's how `mv --interactive this that` seems to work too, i.e it goes like "wanna overwrite? (y/n)" for each file. Same for `rm`, etc.

I'm not sure what to do if it just shows me 32 warning messages and goes "do you want to continue moving the remaining 5 documents?" (what documents even? my screen is now full of warnings! :shrug:)

papis/commands/mv.py
jghauser: I changed the docs to clarify that collisions concern only the final resolved paths:

```
  If multiple selected documents resolve to the exact same folder name, Papis will
  warn you and skip those documents. Consider the above case when the target folder
  doesn't exist. Here, each document would be renamed to ``rick-astley``, causing
  Papis to skip the documents to avoid collisions. In contrast, if the target folder
  already exists, each document will be placed into its own subfolder (e.g.
  "rick-astley/rick-s-the-best" and "rick-astley/rick-s-very-cool"), so there is
  no risk of collision.
```

papis/commands/mv.py
jghauser: currently `mv` aborts when we encounter some unexpected error in the second loop (e.g. permissions error). should i update the docs to make this clearer with e.g.

```
- You can use the ``--batch`` flag to skip all warnings. You will not be prompted
  if there are problematic moves and all unproblematic moves will happen
  automatically. If an unexpected error occurs while moving a document (e.g. a
  permission error), the command aborts immediately and does not attempt any further
  moves.
```

or should I change how the command works so we continue trying to move all the other docs?

jghauser:
> (a) That's fair, but then we should maybe silence the other messages? What else is shown during the moves here? Since we're just going through the documents, I would expect we have full control of what gets displayed. If it feels noisy, just put some of them in `logger.debug`?

Fair enough! And also moot given you want to handle each problem as it arises (which I'm fine with).

> (b) I'm not sure I follow. If document doc[0] wants to move to folder0 and doc[1] also wants to move to folder0, we'll know because folder0 is already in the list of new folders (i.e. you can keep a moving version of that `conflict_groups` dict and just not add a new folder to `moves` if a key already exists). The way it currently works with 3 loops to check gather+deduplicate can just be one loop, I think.

Are you saying the first doc should work just fine but then the second (and third etc) should be skipped? Or are you saying that once the second doc comes around, we should emit warnings about both being skipped (and then further warnings if/when additional collisions arise)? I guess the second of these is fine with me too (the first feels a bit arbitrary about which move happens to be considered problematic).

papis/commands/mv.py
jghauser: Just to make sure, you're saying this is fine for now but we should clean it up in a future PR to use `pathlib` across the codebase?

papis/commands/mv.py
jghauser: Sounds good!

alexfikl:
> Are you saying the first doc should work just fine but then the second (and third etc) should be skipped? Or are you saying that once the second doc comes around, we should emit warnings about both being skipped (and then further warnings if/when additional collisions arise)? I guess the second of these is fine with me too (the first feels a bit arbitrary about which move happens to be considered problematic).

Hm, that's fair.. I didn't really think about the second case. My main preference for the first case is that it's very simple, haha, but yeah, without any sorting it's pretty arbitrary.

For the second case, we need multiple loops like what you implemented :\ Ok, so this is indeed the least surprising / most informative like you said. The algorithm would be
1. Gather all the moves with a `moves[newfolder] = (doc, oldfolder)` and a `conflicts[newfolder] = [doc, doc, doc]`. Popping existing things from `moves` when we find a conflict. 
    - These are just conflicts between the selected documents, if the folders exist on disk already we should ask/skip based on `--batch` directly here.
2. Go through the conflicts 
	- on `--batch` just skip them,
	- otherwise ask the user if they want to keep any of them? Move the ones they want to keep to `moves`.
3. Move them.

Did I miss anything? :facepalm: Is that what you had in mind? More or less..

papis/commands/mv.py
alexfikl: I guess it's fine. It just really stuck out because we don't use pathlib almost at all and there were a lot of forced conversions.

jghauser:
> For the second case, we need multiple loops like what you implemented :\ Ok, so this is indeed the least surprising / most informative like you said. The algorithm would be
> 
>     1. Gather all the moves with a `moves[newfolder] = (doc, oldfolder)` and a `conflicts[newfolder] = [doc, doc, doc]`. Popping existing things from `moves` when we find a conflict.
>        
>        * These are just conflicts between the selected documents, if the folders exist on disk already we should ask/skip based on `--batch` directly here.
> 
>     2. Go through the conflicts
>        
>        * on `--batch` just skip them,
>        * otherwise ask the user if they want to keep any of them? Move the ones they want to keep to `moves`.
> 
>     3. Move them.
> 
> 
> Did I miss anything? 🤦 Is that what you had in mind? More or less..

Yeah, that's what I had in mind. I have one last question: Do you want to ask the user whether they want to move one of the docs that collide? So out of each set of docs with the same target path, they may, if they want, select one (but no more than one) doc to add to `moves`?

alexfikl:
> Yeah, that's what I had in mind. I have one last question: Do you want to ask the user whether they want to move one of the docs that collide? So out of each set of docs with the same target path, they may, if they want, select one (but no more than one) doc to add to `moves`?

Hm.. I don't know.. yes? My personal use case for this will probably be to just rename folders to their `add-folder-name`, in which case what I really want is a `--rename-on-conflict` sort of functionality to match what `papis add` is doing.

What do you think? I can't really see a case where I'd move more documents to something like `{doc[year]}-{doc[author]}`, have collisions and just pick one of them while leaving the others unmoved. Pff.. Maybe just ask "rename or skip?"?

papis/commands/mv.py
jghauser: yeah paths are relative to the target library's root. I've amended the docstring and also added some text to the examples section that clarifies this.

papis/commands/mv.py
jghauser: haha good point :)

papis/commands/mv.py
jghauser: Fair, I'll make a note to more generally move the codebase twoards `pathlib` in a future PR

papis/commands/mv.py
jghauser: not so much silently as without effect. when `git` can't do what we ask it to do it will complain and that will show up in user facing messages, but we don't do anything with that information. but i don't really see what else we can do there... good enough?

alexfikl: Yeah, right! I think that's fine to do in a separate PR, not super urgent. `subprocess.run(...)` can eat stdout / stderr if you tell it, so we can get rid of those messages (or display them in a nicer way).

papis/commands/mv.py
alexfikl: Yeah, that sounds clearer. :grin:

papis/commands/mv.py
jghauser: Gotcha!

papis/commands/mv.py
jghauser: I'll make an issue for it. I don't think we should swallow it but rather display them nicely as warnings the "Papis" way.

papis/commands/mv.py
jghauser: Just to make sure: do you think we should abort on first error of this kind (and improve the docs) or should we continue with the remaining moves in batch mode? I'm now actually tending towards just carrying on and trying to move the other documents. That seems to be the expectation in batch mode elsewhere...

alexfikl: Personally, on `--batch` I would also like to just skip and go to the next document.

papis/commands/mv.py
jghauser: Good point. Unfortunately, this applies to various other config values too (for sure the ones relevant for slugification, `doc-paths-lowercase`, `doc-paths-extra-chars`, `doc-paths-word-separator`). Should we just handle `add-folder-name` for now and make a note of the rest for future work with proper context switching between different library's configs?

papis/commands/mv.py
jghauser: Ok, I agree. I'll change it

papis/commands/mv.py
alexfikl: I think you can just get all the values you need from the current library and then do a `set_lib_from_name(target_library)` to switch for the rest of the code? I'm not sure if that would break too much :-s

papis/commands/mv.py
jghauser: Discussed elsewhere

papis/commands/mv.py
jghauser: Now handled differently

papis/commands/mv.py
jghauser: New approach handles problems individually

papis/commands/mv.py
jghauser: oops my bad, this one's absolute actually. changed! (I thought we were in the `cli()` function :facepalm:)

papis/commands/mv.py
jghauser: Oh, nice idea!

jghauser:
Now with individual warning handling and offer for manual renaming whenever relevant :)

jghauser:
Though clearly I did something wrong when rebasing... ugh. will fix later

EDIT: fixed

alexfikl:
> Now with individual warning handling and offer for manual renaming whenever relevant :)

:tada: Will try to take another look in the next few days. I think this is pretty good now, so feel free to add tests and whatever else is needed to finish it up.

jghauser:
Great, thanks a lot! I'll add some more tests, and an entry to the changelogs. One more question: `papis mv` now does everything that one might have wanted to use `papis rename` for. I think there was a consensus around removing that command (if I remember correctly I was the holdout, but I now think we really don't need it anymore). should I add a commit to this PR removing it? Or follow-up PR?

jghauser:
I'll investigate why some tests fail in CI later :)

jghauser:
Our library switching shenanigans still don't work quite always. There's stuff in `run()` that also queries config values (e.g. `dir-umask`). Additionally, there's an issue with the fastAPI server, which will run into trouble once there's concurrent requests. I'm thinking of fixing this properly by replacing the module-level `CURRENT_LIBRARY` variable with a `contextvars.ContextVar` and exposing a `papis.config.library_context(name)` context manager. We can then scope target-library operations inside a `with` block and in the server I can set a context per request. Probably makes sense to open another PR for this and then rebase `mv` on top of the changes. Does this sound sensible?

alexfikl:
> There's stuff in run() that also queries config values (e.g. dir-umask). 

I'm not sure I understand why this is a problem? If the library sets a different `dir-umask` than the global one, we should also use that.

> Does this sound sensible?

I guess? I haven't actually used `contextvars`, so I'm not very familiar with all the problems they solve. They work with `asyncio` at least, so it's probably enough for the server part..

The main issue seems to be that our configuration is a mutable global, so this just solves one part of that? And you need to remember when to use the context manager because any function could theoretically call either `papis.config.get` or `papis.config.set`..

jghauser:
> > There's stuff in run() that also queries config values (e.g. dir-umask).
> 
> I'm not sure I understand why this is a problem? If the library sets a different `dir-umask` than the global one, we should also use that.

The problem is that it doesn't currently do that :sweat_smile:. I need to wrap parts of the `run()` function into `    papis.config.set_lib_from_name()` as well. To be fair, that's a different problem, which can be solved without `contextvars`.

> > Does this sound sensible?
> 
> I guess? I haven't actually used `contextvars`, so I'm not very familiar with all the problems they solve. They work with `asyncio` at least, so it's probably enough for the server part..
> 
> The main issue seems to be that our configuration is a mutable global, so this just solves one part of that? And you need to remember when to use the context manager because any function could theoretically call either `papis.config.get` or `papis.config.set`..

How do you mean it only solves one part? You do still need to wrap anything into `contextvars` blocks if needed, so it's not much different from what I'm currently doing in `mv`, except it solves the server's concurrency problem. The 'proper' way I see to solve this is to have everything that does `config.set` or `config.get` require a `library` argument. But that would be a rather intimidating undertaking...

alexfikl:
> How do you mean it only solves one part? 

I meant that we have other mutable global variables, not just `CURRENT_LIBRARY`, so there are probably many more concurrency issues. I wouldn't be surprised if the databases also don't quite work well..

> The 'proper' way I see to solve this is to have everything that does `config.set` or `config.get` require a `library` argument. But that would be a rather intimidating undertaking...

Hm.. not sure I like that anyway. As it stands, for me, a "library" isn't particularly different than any other section in the config, so it would feel weird to special case it like that. We probably need a restructure the config a bit to make this nice..

jghauser:
> > How do you mean it only solves one part?
> 
> I meant that we have other mutable global variables, not just `CURRENT_LIBRARY`, so there are probably many more concurrency issues. I wouldn't be surprised if the databases also don't quite work well..
> 
> > The 'proper' way I see to solve this is to have everything that does `config.set` or `config.get` require a `library` argument. But that would be a rather intimidating undertaking...
> 
> Hm.. not sure I like that anyway. As it stands, for me, a "library" isn't particularly different than any other section in the config, so it would feel weird to special case it like that. We probably need a restructure the config a bit to make this nice..

Both good points. I'll make a note to look into the concurrency issues for the API. Here, I'll just fix the `run()` function to properly switch configs.

papis/commands/mv.py
```diff
@@ -1,111 +1,473 @@
...
+
+The command works similarly to the Unix
+`mv <https://en.wikipedia.org/wiki/Mv_(Unix)>`__ command. Hence, when the
+specified target folder exists, the document's current main folder is moved into
+the target folder. If the specified target folder doesn't exist, then the
+current folder is renamed to the target folder name. Unlike Unix ``mv``, if a
+target folder isn't given, it defaults to the :confval:`add-folder-name` pattern,
+so ``papis mv QUERY`` renames matching documents to their default folder names.
+
+Folder names are cleaned, so that various problematic characters are avoided (see
```
alexfikl:
```suggestion
Folder names are cleaned to contain only the character set included with :confval:`doc-paths-extra-chars` (see :func:`papis.paths.normalize_path_part`)
```
or something like that? "problematic" sounds weird here (when should I expect problems? what kind of problems?).

papis/commands/mv.py
```diff
@@ -1,111 +1,473 @@
...
+:confval:`doc-paths-extra-chars`). This cleaning applies to all path components,
+including any literal subfolders passed to ``--to``, and you cannot use this
+command to move docs into a folder with problematic characters.
+
+The ``mv`` command will first check whether moving the folders as specified will
+result in problems. If any problems are found, they are displayed and you will
```
alexfikl:
```suggestion
The ``mv`` command will first check whether moving the folders as specified will
fail. If any errors are encountered, they are displayed and you will
```
? it feels weird to "display problems" :grin:

papis/commands/mv.py
```diff
@@ -1,111 +1,473 @@
...
+
+    :param target: the target path for the document
+    :param lib_dir: the root directory of the target library.
+    :returns: ``None`` if not nested in a document folder, an error string otherwise
+    """
+    current = target.parent
+    while current not in {lib_dir, current.parent}:
+        if papis.document.is_document_folder(current):
+            return f"would nest inside another document's folder: {current}"
+        current = current.parent
```
alexfikl, line 135: When would this sort of nesting happen? I would expect one level at most.

papis/commands/mv.py
```diff
@@ -1,111 +1,473 @@
...
+    """Check that *path* doesn't already exist.
+
+    :returns: ``None`` if the path doesn't already exist, an error string otherwise
+    """
+    if path.exists():
+        return f"target already exists: {path}"
+    return None
+
+
+def _resolve_target_interactively(
```
alexfikl: `resolve` tends to bring other things to mind: https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve

`rename` maybe?

papis/commands/mv.py
```diff
@@ -1,111 +1,473 @@
...
+    git: bool = False,
+) -> None:
+    """Move a document's folder to *doc_folder_new*.
 
-    folder = document.get_main_folder()
-    if not folder:
-        from papis.exceptions import DocumentFolderNotFound
-        raise DocumentFolderNotFound(describe(document))
+    :param document: The document to move.
+    :param doc_folder_new: The target folder path (absolute)
```
alexfikl:
```suggestion
    :param doc_folder_new: the target folder path (absolute)
```
? We don't really uppercase the first word in these things, but you can leave it. 

Also, if the doc string is just 3 words, maybe you want to rename `doc_folder_new` to `target_path_abs` and make it redundant :grin:

papis/commands/mv.py
```diff
@@ -1,111 +1,473 @@
...
-        shutil.move(folder, new_folder_path)
+        shutil.move(doc_folder_old, main_folder_new)
+
+        if git:
+            papis.git.remove(str(source_lib_dir), str(main_folder_old))
+            papis.git.commit(
+                str(source_lib_dir), f"Remove '{papis.document.describe(document)}'"
+            )
+
+            # NOTE: this will fail when the target library isn't git tracked
```
alexfikl:
```suggestion
```

papis/commands/mv.py
```diff
@@ -1,111 +1,473 @@
...
+            shutil.move(main_folder_old, main_folder_new)
     else:
-        import shutil
-        shutil.move(folder, new_folder_path)
+        shutil.move(doc_folder_old, main_folder_new)
+
+        if git:
+            papis.git.remove(str(source_lib_dir), str(main_folder_old))
+            papis.git.commit(
+                str(source_lib_dir), f"Remove '{papis.document.describe(document)}'"
```
alexfikl:
```suggestion
                str(source_lib_dir), f"Remove '{papis.document.describe(document)}' (moved to '{target_library}' library)"
```
? :grin:

papis/commands/mv.py
```diff
@@ -1,111 +1,473 @@
...
-    db = get_database()
-    db.delete(document)
+    source_db = papis.database.get()
+    source_db.delete(document)
 
-    new_document_folder = os.path.join(
-        new_folder_path,
-        os.path.basename(folder))
-    logger.debug("New document folder: '%s'.", new_document_folder)
+    logger.debug("New document folder: '%s'.", main_folder_new)
```
alexfikl:
```suggestion
```
There's an info just below that says where it moved?

papis/commands/mv.py
```diff
@@ -1,111 +1,473 @@
...
+                document, target_lib_dir, folder_name_format=to
+            )
+        )
+
+        if main_folder_old == target_abs_path:
+            logger.info(
+                "Skipping '%s': source and target are the same.",
+                papis.document.describe(document),
+            )
+            continue
```
alexfikl, line 385: Shouldn't this also ask for a rename?

papis/commands/mv.py
```diff
@@ -1,111 +1,473 @@
...
+                papis.document.describe(document),
+                error,
+            )
+            if batch:
+                continue
+            logger.warning("You can skip this document or rename it manually.")
+            if papis.tui.utils.confirm(
+                "Skip this document?",
+                yes=True,
+            ):
```
alexfikl, line 409: Haha, `ruff format` sure is eager to break things into many lines.

papis/commands/mv.py
```diff
@@ -1,111 +1,473 @@
...
+            logger.warning(
+                "'%s': %s.",
+                papis.document.describe(document),
+                error,
+            )
+            if batch:
+                continue
+            logger.warning("You can skip this document or rename it manually.")
+            if papis.tui.utils.confirm(
+                "Skip this document?",
```
alexfikl:
```suggestion
            if papis.tui.utils.confirm(
                "Skip this document? (continue to rename it instead)",
```
or something like that? I can call `papis --log ERROR` and never see the actual message, so anything really important should be in the prompt.

alexfikl (approved):
Left a few nitpicks in there, but nothing interesting this time around. The logic seems ok to me now, so feel free to merge it in when you think it's ready!

The "conflict resolution" part looks horribly convoluted and will probably be a pain to maintain, but we'll see 🤷

papis/commands/mv.py
jghauser: Not sure I like pointing users in the direction of source code, but I'll rephrase it otherwise.

jghauser: Maybe this?

```
Folder names are cleaned, so that characters that can lead to invalid file paths
are avoided (see :confval:`doc-paths-extra-chars` for a list of allowed characters
and how to change it). This cleaning applies to all path components,
including any literal subfolders passed to ``--to``. This means and you cannot use
the ``mv`` command to move docs into a folder containing any such excluded
characters.
```

papis/commands/mv.py
jghauser: Exactly the kind of pedantry I can get behind! :laughing:

papis/commands/mv.py
alexfikl: Yeah, that sounds less confusing!

papis/commands/mv.py
jghauser: Usually you're right in that for most users a doc folder is just one level deep relative to the library root. but that's not a given, users can put their folders in arbitrarily nested subfolders, which this addresses.

papis/commands/mv.py
jghauser: lowercased and renamed :)

papis/commands/mv.py
jghauser: Good idea!

papis/commands/mv.py
jghauser: I'm not sure I like this. I was thinking of the common case of renaming things to their defaults. If we prompt here, then all documents that happen to generate no changes in their doc folders will trigger a prompt. I think it might be better for the common case to assume that 'same doc folder name' is intentional.

papis/commands/mv.py
jghauser: it's called foresight: now that we made the string longer to tell the user that saying 'no' will allow them to rename the folder, it really does need to be on multiple lines :wink:

papis/commands/mv.py
```diff
@@ -1,111 +1,473 @@
...
+            continue
 
-    if not os.path.exists(new_folder):
-        logger.info("Creating path '%s'.", new_folder)
-        os.makedirs(new_folder, mode=papis.config.getint("dir-umask") or 0o666)
+        target_abs_path = Path(
+            papis.paths.get_document_folder(
+                document, target_lib_dir, folder_name_format=to
+            )
+        )
```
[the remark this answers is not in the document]
jghauser: leaving as is since the error indicates a wrongly set `folder_name_format` and the error is left to propagate in `add` as well.

papis/commands/mv.py
```diff
@@ -1,111 +1,473 @@
...
+                papis.document.describe(document),
+            )
+            continue
+
+        # Resolve rename vs move-into semantics
+        if target_abs_path.is_dir():
+            main_folder_new = target_abs_path / main_folder_old.name
+        else:
+            main_folder_new = target_abs_path
+
```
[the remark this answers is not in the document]
jghauser: this implies the user's library doesn't follow good practices. leaving as is

papis/commands/mv.py
```diff
@@ -1,111 +1,473 @@
...
+                moves[target_path] = move
+
+        if not new_conflicts:
+            break
+
+        conflicts = new_conflicts
+
+    if not moves:
+        logger.info("No valid moves to perform.")
+        return
```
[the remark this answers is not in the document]
jghauser: leaving as is as the return closes the cli command and the next invocation will start afresh

jghauser:
In it goes :tada:. Thanks everyone! :heart:

PR #1108, opened 2025-11-30
Allow papis update to rename files

### LINKED ISSUE #366: [Question/Suggestion] Is there any quick way to update the PDF file itself via `papis update` or a similar command?
As titled. Currently, I have a few documents (books and lecture notes - so most `.yaml` info is manually written) in my library that receive updates or are in the process of being written (i.e. new draft every few months).

Currently, my workflow has been:
1. looking up the document path via `papis list` 
2. and directly overwriting the PDF itself. 

However, I feel like there should be a better way to do this, perhaps by feeding a file to `papis update` or some similar command, allowing me to select which document's file to replace with the new one? (Or something similar, as long as it's a `papis` command)

I have not been able to find an answer to this in the docs (`papis update` seems only capable of updating info in the `.yaml` file), so this is both a question and a suggestion

____________________________________

_I'm aware that I also have the alternative of making folder names more human-readable (by changing settings) or manually deleting and adding the documents again. Obviously, both too much manual effort that should probably be automated - especially as these are mostly books and lecture notes that I'm referring to (with no ISBN, arXiv ID, etc.)_

alexfikl:
This reworks `papis update` quite extensively to better handle error cases and rename files in different scenarios.

That being said, the way this works now is that
```sh
# sets the notes file to the given value/format pattern
papis update --set notes notes.md <QUERY>
# sets the notes file to the default pattern from `notes-name`
papis update --reset notes <QUERY>
```
If a "notes" file already exists, it's also renamed in both cases. If it doesn't exist, this just sets the value in the `info.yaml` file, so that it can be created on the next `papis edit` or whatever.

Then, for files, this updates it to work something like
```sh
# sets the n-th file in the files list to the given value/format pattern
papis update --set files n:my-file-name.pdf <QUERY>
# sets the whole files field to something
papis update --set files '["my-file-name.pdf"]' QUERY
# sets all the files in the files list to their default pattern from `add-files-name`
papis update --reset files <QUERY>
```
All the files are also normalized (exactly the same as in `papis add`), so the name on the command line may not always match what gets put in the document.

TODO:
- [x] Need a nice way to set all the files to the default.
- [x] Need some tests, since this is all a bit finicky.
- [x] Probably need to think of a more natural interface..

~@jghauser Like you said, none of the flags in `papis update` allow easily setting a thing to some existing default, so this is not particularly nice. Do you have any better ideas?~

cc @pmiam for some constructive criticism :grin:

Fixes #366

### DIFF
--- a/papis/commands/update.py
+++ b/papis/commands/update.py
@@ -1,19 +1,21 @@
 """
 This command allows you to update the document metadata stored in the
-``info.yaml`` file. With it, you can either change individual values manually
+``info.yaml`` file. With it, you can either set individual fields' values manually
 or update a document with information automatically retrieved from a variety
-of sources.
+of sources. The command also renames files and notes on disk when needed.
 
-When using it to add information, Papis formatting patterns and Python
-expressions can be used. See below examples for more information. The command
-also tries to sanitise filenames so that they don't contain any problematic
-characters.
+Papis formatting patterns and Python expressions can be used. See below examples
+for more information. The command also sanitises filenames so that they don't
+contain any problematic characters (see :confval:`doc-paths-extra-chars`).
 
-Normally, ``papis update`` will prompt you when it encounters an error. If you
-want to continue errors without input and apply as many changes as possible,
-use the ``--batch`` flag. With this in mind, the command is mostly meant to be
-used to apply sweeping changes to a larger number of documents. If you just
-want to update a single document, ``papis edit`` might be more appropriate.
+By default, ``papis update`` will prompt you when it encounters an error. If you
+want to skip errors automatically and apply as many changes as possible,
+use the ``--batch`` flag.
+
+This command is mostly meant to be used to apply sweeping changes across a large
+number of documents. If you just want to update a single document, ``papis edit``
+might be more appropriate.
+
 
 Examples
 ^^^^^^^^
@@ -37,25 +39,10 @@ Examples
 
 
   The ``--all`` flag means that the operation is applied to all documents that
-  match the query, rather than allowing you to pick one individual document to
+  match the query, rather than opening the picker to select individual documents to
   update.
 
-- Update a document automatically and interactively (searching by DOI in
-  Crossref or in other sources...):
-
-    .. code:: sh
-
-        papis update --auto "author:dyson"
-
-- Update your document from a DOI using the importer functionality as
-
-    .. code:: sh
-
-        papis update --from doi '10.1103/PhysRev.47.777' 'author:einstein'
-
-  For a list of all supported importers use ``papis update --list-importers``.
-
-- Add the ", Albert" to the author string of a documents matching 'Einstein':
+- Add ", Albert" to the author string of a documents matching 'Einstein':
 
     .. code:: sh
 
@@ -65,20 +52,19 @@ Examples
   formatter. Here, it is used to get the existing author "Albert" and then add
   the string ", Einstein" to end up with "Einstein, Albert".
 
-- Reset keys to their default values using:
+- Reset a field to its default value using:
 
     .. code:: sh
 
         papis update --reset ref Einstein
 
-  This will reset the value to the default value defined for the key. In the
-  above case the "ref" is set to the format described by :confval:`ref-format`.
-  Other supported keys are: "author" (gets updated from ``author_list`` and
-  :confval:`multiple-authors-format`), "notes" (using :confval:`notes-name`),
-  and "files" (using :confval:`add-file-name`). Other keys do not support this
-  as they do not have any well-defined default.
+  This will reset the value to the default value defined for the field. Here, the "ref"
+  is set to :confval:`ref-format`. Other fields that can be reset are: "author" (gets
+  updated from ``author_list`` and :confval:`multiple-authors-format`), "notes"
+  (using :confval:`notes-name`), and "files" (using :confval:`add-file-name`). Other
+  fields do not support this as they do not have any well-defined default.
 
-- The ``--append`` option can be used to append a string to an existing key:
+- The ``--append`` option can be used to append to a string value:
 
     .. code:: sh
 
@@ -99,9 +85,9 @@ Examples
     doesn't yet exist, it will be created. The new tag will only be appended to
     the list if the tag does not already exist (as an exact case-sensitive match).
 
-    The ``--append`` flag needs to know the type of the key it is appending to.
-    If the key exists in the document, then the value set in the document
-    determines the type. If the key doesn't exist in the document, the command
+    The ``--append`` flag needs to know the type of the field it is appending to.
+    If the field exists in the document, then the value set in the document
+    determines the type. If the field doesn't exist in the document, the command
     looks at the list of types defined in the :confval:`document-field-types`
     (and :confval:`document-field-types-extend`) configuration option. If the
     type cannot be determined in either of these two ways, the command will
@@ -114,7 +100,7 @@ Examples
 
         papis update --remove tags physics 'author:einstein'
 
-- To remove a key-value pair entirely, use ``--drop``. For example, to remove all
+- To remove a field entirely, use ``--drop``. For example, to remove all
   tags from documents use
 
     .. code:: sh
@@ -122,7 +108,9 @@ Examples
         papis update --drop tags 'author:einstein'
 
 - There is also a convenience option ``--rename`` if you want to rename
-  a list item. It's equivalent to doing ``--remove`` and ``--append``, but as
+  a list item. It's equivalent to doing ``--remove`
+and`
+--append``, but as
   a single operation:
 
     .. code:: sh
@@ -133,8 +121,8 @@ Examples
   not be added to the list if the original tag doesn't exist.
 
 - The ``--batch`` flag suppresses interactive prompts and skips any document
-  that cannot be fully updated. This is useful when applying the same operation
-  across many documents where some may not have the expected keys or values:
+  for which an update operation has failed. This is useful when applying the same
+  operation across many documents where some may not have the expected keys or values:
 
     .. code:: sh
 
@@ -147,42 +135,67 @@ Examples
 
   More specifically, ``--batch`` suppresses prompts for the following errors:
 
-  * An operation on a key fails (e.g. trying to ``--remove`` a value that is
+  * An operation on a field fails (e.g. trying to ``--remove`` a value that is
     not in the list, ``--append`` to a key of unknown type, or a type
     conversion error from ``--set``). The affected key retains its original
     value and the remaining keys for that document are still processed.
   * A file rename fails (e.g. the file on disk is missing or cannot be moved).
     The whole document is skipped.
 
-  In all cases the command exits with a non-zero status if any document was
-  skipped.
+  In all cases the command exits with a non-zero status (indicating a failure) if
+  any document was skipped.
 
-Advanced Examples
-^^^^^^^^^^^^^^^^^
+- Update a document automatically and interactively (searching by DOI in
+  Crossref or in other sources...):
 
-- When you update the ``files`` or ``notes`` keys, the corresponding files on
-  disk are also renamed to match the new value:
+    .. code:: sh
+
+        papis update --auto "author:dyson"
+
+- Update your document from a DOI using the importer functionality:
+
+    .. code:: sh
+
+        papis update --from doi '10.1103/PhysRev.47.777' 'author:einstein'
+
+  For a list of all supported importers use ``papis update --list-importers``.
+
+- When you update the ``files`
+or`
+notes`` fields, the corresponding files on
+  disk are renamed to match the new value:
 
     .. code:: sh
 
         papis update --set notes "my-new-notes.tex" Einstein
 
-  This renames the ``notes`` file on disk from its current name to
-  ``my-new-notes.tex`` and updates the ``info.yaml`` file accordingly. The same
-  applies when updating ``files``:
+  This updates the ``info.yaml`
+file and renames the`
+notes`` file on disk from
+  its current name to ``my-new-notes.tex``
+. The same applies when updating``
+files``:
 
     .. code:: sh
 
-        papis update --set files 0:einstein-new.pdf Einstein
+        papis update --rename files einstein-old.pdf einstein-new.pdf Einstein
 
-  This renames the first file in the document's file list to ``einstein-new.pdf``
-  on disk. If the rename fails (e.g. the file does not exist or there is a
-  permissions error), the metadata is not updated and an error is reported.
+  This renames the file ``einstein-old.pdf`
+to`
+einstein-new.pdf`` in the
+  ``info.yaml`` file and on disk. If the rename fails (e.g. the file does not exist
+  or there is a permissions error), the metadata is not updated and an error is
+  reported.
 
   Note that the ``--reset`` option also triggers a rename when used with
-  ``notes`` or ``files``. It renames the files to the names derived from the
+  ``notes`
+or`
+files``. It renames the files to the names derived from the
   configured :confval:`notes-name` and :confval:`add-file-name` formats
   respectively.
+
+Advanced Examples
+^^^^^^^^^^^^^^^^^
 
 - As an advanced feature, ``papis update`` also supports the parsing of Python
   expressions (such as lists or dictionaries). This can be used as follows:
@@ -194,15 +207,16 @@ Advanced Examples
   Because the above string is a valid Python expression, ``author_list`` is
   updated to a list that contains a dictionary.
 
-- You can use ``--set`` to set a value into a list at a given position:
+- You can use ``--set`` to set a value in a list at a given position:
 
     .. code:: sh
 
         papis update --set files 0:some-new-file.pdf Einstein
 
-  This special syntax for ``--set`` only works with keys that are known to be
-  lists (as before, based on :confval:`document-field-types` and
-  :confval:`document-field-types-extend`). If the key is not a list, then the
+  This command will rename the first entry in the list of files. This special
+  syntax for ``--set`` only works with keys that are known to be lists (as
+  determined based on :confval:`document-field-types` and
+  :confval:`document-field-types-extend`). If the field is not a list, then the
   value is treated as a string. This might be unexpected, so make sure you are
   using it for list keys only.
 
@@ -212,6 +226,7 @@ Command-line interface
 .. click:: papis.commands.update:cli
     :prog: papis update
 """
+
 from __future__ import annotations
 
 import ast
--- a/papis/commands/update.py
+++ b/papis/commands/update.py
@@ -9,8 +9,11 @@ expressions can be used. See below examples for more information. The command
 also tries to sanitise filenames so that they don't contain any problematic
 characters.
 
-Normally, ``papis update`` will abort on encountering an error. If you want to
-skip errors and apply as many changes as possible, use the ``--batch`` flag.
+Normally, ``papis update`` will prompt you when it encounters an error. If you
+want to continue errors without input and apply as many changes as possible,
+use the ``--batch`` flag. With this in mind, the command is mostly meant to be
+used to apply sweeping changes to a larger number of documents. If you just
+want to update a single document, ``papis edit`` might be more appropriate.
 
 Examples
 ^^^^^^^^
@@ -33,8 +36,9 @@ Examples
         papis update --all --set journal "Mass and Energy" "journal:'Energy and Mass'"
 
 
-  The ``--all`` flag means that the tag is applied to all documents that match the
-  query, rather than allowing you to pick one individual document to update.
+  The ``--all`` flag means that the operation is applied to all documents that
+  match the query, rather than allowing you to pick one individual document to
+  update.
 
 - Update a document automatically and interactively (searching by DOI in
   Crossref or in other sources...):
@@ -43,14 +47,13 @@ Examples
 
         papis update --auto "author:dyson"
 
-- Update your library from a BibTeX file, where many entries may be listed:
+- Update your document from a DOI using the importer functionality as
 
     .. code:: sh
 
-        papis update --from bibtex libraryfile.bib
+        papis update --from doi '10.1103/PhysRev.47.777' 'author:einstein'
 
-  Papis will try to look for documents in your library that match these
-  entries and will ask you for each entry whether you want to update it.
+  For a list of all supported importers use ``papis update --list-importers``.
 
 - Add the ", Albert" to the author string of a documents matching 'Einstein':
 
@@ -60,71 +63,148 @@ Examples
 
   The ``papis update`` command tries to format input strings using the configured
   formatter. Here, it is used to get the existing author "Albert" and then add
-  the string ", Einstein" to end up with "Einstein, Albert"
+  the string ", Einstein" to end up with "Einstein, Albert".
 
-- The above can also be achieved with the ``--append`` option:
+- Reset keys to their default values using:
+
+    .. code:: sh
+
+        papis update --reset ref Einstein
+
+  This will reset the value to the default value defined for the key. In the
+  above case the "ref" is set to the format described by :confval:`ref-format`.
+  Other supported keys are: "author" (gets updated from ``author_list`` and
+  :confval:`multiple-authors-format`), "notes" (using :confval:`notes-name`),
+  and "files" (using :confval:`add-file-name`). Other keys do not support this
+  as they do not have any well-defined default.
+
+- The ``--append`` option can be used to append a string to an existing key:
 
     .. code:: sh
 
         papis update --append author ", Albert" Einstein
 
-    This appends ", Albert" to the existing author string.
+  This appends ", Albert" to the existing author value. Note that it will be
+  appended to the existing value only if it does not already end with that
+  exact string (case-sensitive). This avoids appending duplicates by mistake, as
+  in the case of lists (below).
 
 - You can also append an item to a list:
 
     .. code:: sh
 
-        papis update --append tags physics
+        papis update --append tags physics 'author:einstein'
 
-    This adds the tag 'physics' to the existing list of tags. If the list
-    doesn't yet exist, it will be created. All duplicate items will be removed
-    from the list.
+    This adds the tag "physics" to the existing list of tags. If the list
+    doesn't yet exist, it will be created. The new tag will only be appended to
+    the list if the tag does not already exist (as an exact case-sensitive match).
 
-    As you might have guessed, the ``--append`` flag needs to know the type of
-    the key it is appending to. It does this by looking at the
-    :confval:`document-field-types` (and :confval:`document-field-types-extend`)
-    configuration options. If the key you are appending to is not in that list,
-    the command will fail.
+    The ``--append`` flag needs to know the type of the key it is appending to.
+    If the key exists in the document, then the value set in the document
+    determines the type. If the key doesn't exist in the document, the command
+    looks at the list of types defined in the :confval:`document-field-types`
+    (and :confval:`document-field-types-extend`) configuration option. If the
+    type cannot be determined in either of these two ways, the command will
+    fail.
 
-- To remove an item from a list, use ``--remove``:
-
-    .. code:: sh
-
-        papis update --remove tags physics
-
-    If the tag "physics" is in the list of tags, this command removes it.
-
-- To remove a key-value pair entirely, use ``--drop``:
+- To remove an item from a list, use ``--remove``. For example, to remove the
+  "physics" tag from the list of document tags, use
 
     .. code:: sh
 
-        papis update --drop tags
+        papis update --remove tags physics 'author:einstein'
 
-    This removes all tags.
+- To remove a key-value pair entirely, use ``--drop``. For example, to remove all
+  tags from documents use
+
+    .. code:: sh
+
+        papis update --drop tags 'author:einstein'
 
 - There is also a convenience option ``--rename`` if you want to rename
-  a list item. It's equivalent to doing ``--remove`` and ``--append`` sequentially:
+  a list item. It's equivalent to doing ``--remove`` and ``--append``, but as
+  a single operation:
 
     .. code:: sh
 
         papis update --rename tags physics philosophy
 
-  This renames the tag 'physics' to 'philosophy'. Note that this option being a
-  combination of ``--remove`` and ``--append``, it will append the desired values
-  even if the value to be removed didn't exist. Thus, the above command will add
-  the tag "philosophy" even if the tag "physics" didn't exist before the
-  operation.
+  This renames the tag "physics" to "philosophy". Note that the new tag will
+  not be added to the list if the original tag doesn't exist.
 
-- As an advanced feature, ``papis update`` also supports the parsing of python
+- The ``--batch`` flag suppresses interactive prompts and skips any document
+  that cannot be fully updated. This is useful when applying the same operation
+  across many documents where some may not have the expected keys or values:
+
+    .. code:: sh
+
+        papis update --all --batch --remove tags obsolete "tags:obsolete"
+
+  Without ``--batch``, if a document does not have "obsolete" in its tag list
+  (or has no ``tags`` key at all), the command pauses and asks whether to
+  continue. With ``--batch``, an error is logged, the document is skipped and the
+  command moves on to the next one.
+
+  More specifically, ``--batch`` suppresses prompts for the following errors:
+
+  * An operation on a key fails (e.g. trying to ``--remove`` a value that is
+    not in the list, ``--append`` to a key of unknown type, or a type
+    conversion error from ``--set``). The affected key retains its original
+    value and the remaining keys for that document are still processed.
+  * A file rename fails (e.g. the file on disk is missing or cannot be moved).
+    The whole document is skipped.
+
+  In all cases the command exits with a non-zero status if any document was
+  skipped.
+
+Advanced Examples
+^^^^^^^^^^^^^^^^^
+
+- When you update the ``files`` or ``notes`` keys, the corresponding files on
+  disk are also renamed to match the new value:
+
+    .. code:: sh
+
+        papis update --set notes "my-new-notes.tex" Einstein
+
+  This renames the ``notes`` file on disk from its current name to
+  ``my-new-notes.tex`` and updates the ``info.yaml`` file accordingly. The same
+  applies when updating ``files``:
+
+    .. code:: sh
+
+        papis update --set files 0:einstein-new.pdf Einstein
+
+  This renames the first file in the document's file list to ``einstein-new.pdf``
+  on disk. If the rename fails (e.g. the file does not exist or there is a
+  permissions error), the metadata is not updated and an error is reported.
+
+  Note that the ``--reset`` option also triggers a rename when used with
+  ``notes`` or ``files``. It renames the files to the names derived from the
+  configured :confval:`notes-name` and :confval:`add-file-name` formats
+  respectively.
+
+- As an advanced feature, ``papis update`` also supports the parsing of Python
   expressions (such as lists or dictionaries). This can be used as follows:
 
     .. code:: sh
 
         papis update --set author_list "[{'family': 'Einstein', 'given': 'Albert'}]"
 
-  Because the above string is a valid python expression, ``author_list`` is
-  updated to a set that contains a dictionary.
+  Because the above string is a valid Python expression, ``author_list`` is
+  updated to a list that contains a dictionary.
 
+- You can use ``--set`` to set a value into a list at a given position:
+
+    .. code:: sh
+
+        papis update --set files 0:some-new-file.pdf Einstein
+
+  This special syntax for ``--set`` only works with keys that are known to be
+  lists (as before, based on :confval:`document-field-types` and
+  :confval:`document-field-types-extend`). If the key is not a list, then the
+  value is treated as a string. This might be unexpected, so make sure you are
+  using it for list keys only.
 
 Command-line interface
 ^^^^^^^^^^^^^^^^^^^^^^
@@ -135,7 +215,9 @@ Command-line interface
 from __future__ import annotations
 
 import ast
-from typing import TYPE_CHECKING, Any
+import enum
+import os
+from typing import TYPE_CHECKING, Any, NamedTuple
 
 import click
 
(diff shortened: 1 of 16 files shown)


papis/commands/update.py
```diff
@@ -183,34 +243,49 @@ def run_set(
...
-                "rename any files on disk."
-            )
-        elif key == "files" and isinstance(value, list):
-            # TODO: handle renames/deletions of files on disk
-            document[key] = []
-            for file in value:
-                if isinstance(file, str):
-                    document[key].append(normalize_path(file))
-                else:
-                    document[key].append(value)
```
alexfikl, line 204: @jghauser I don't quite recall how this was supposed to work? It seems to just overwrite the whole files list + not sure what that `isinstance(file, str)` is supposed to do..

papis/commands/update.py
```diff
@@ -152,24 +153,83 @@
...
+    from papis.paths import _make_unique_file
+    dst = _make_unique_file(os.path.join(directory, dst))
+
+    import shutil
+    shutil.copyfile(src, dst)
+
+    return os.path.basename(dst)
+
+
+def _parse_file_name(document: DocumentLike, value: str) -> tuple[int, str] | None:
```
alexfikl: TODO: This needs to be a lot smarter about extensions, preferably using something from `papis.paths` 🤦

papis/commands/update.py
```diff
@@ -183,34 +243,49 @@ def run_set(
...
+        elif key == "files":
+            prev_files = document[key]
+            result = _parse_file_name(document, value)
+            if result is not None:
+                index, value = result
+                value = normalize_path(value)
+
+                if document_folder:
+                    value = _move_files(document_folder, prev_files[index], value)
+                document[key][index] = value
```
alexfikl: This implementation is not good at all, because it's moving files around without checking for successful changes anywhere. I'm working on something a bit nicer :grin:

jghauser:
Thanks a lot for working on this! :heart: 

Given that files are a list, maybe we should use `papis update --rename files OLD_NAME NEW_NAME`? It's a bit verbose but most in line with how `papis update` works otherwise. I never really saw `--set` to be useful for lists (except when using a python expression, but that's really not very nice). If file names could autocomplete this approach would even be somewhat ergonomic (and also, it wouldn't treat `files` as a special case). Hmm, now I wonder whether that's what you meant with having `files` as a special case... sorry if I missed that!

I wonder whether `--set` could just default to assuming a list of one item. So if you use set you don't have to specify the old name, but if there are multiple items in the set, it'll give you a warning (for `files` but also other lists). It does make `--set` functionality a bit harder to parse though.

Re defaults, I'm not sure. I'll think about it. I don't really like the `DEFAULT` approach. It would be nice if it worked like other commands where `DEFAULT` is assumed when nothing is specified. But we can't easily do this since the `<QUERY>` will be assumed to be the missing argument. Having query be the final argument (and without an explicit flag/option) makes solving this hard...

pmiam:
With the recent addition of an ipad to my reading workflow, I find 
myself in need of this feature. Long story. Excellent timing.

I see the ideal implementation as equally supporting two options for
updating metadata/files

I imagine the first similar to:

1. run `papis edit <QUERY>`
2. freely edit info.yaml (or just do this somehow, e.g. with sed or wgrep or whatever)
3. run `papis update`
4. changes to metadata propagate to all paths

Alex Fikl ***@***.***> writes:
> If a &quot;notes&quot; file already exists, it&#39;s also renamed in both cases. If it doesn&#39;t exist, this just sets the value in the `info.yaml` file, so that it can be created on the next `papis edit` or whatever.

I like this approach for handling nonexistant paths. It avoids surprises.

The second is naturally the command line driven option. As usual it is
stymied by a lack of certainty about how to support setting list elements

> something like
> ```
> # sets the first file in the files list to the given value/format pattern
> papis update --set files my-file-name.pdf &lt;QUERY&gt;
> # sets the first file in the files list to the default pattern from `add-files-name`
> papis update --set files DEFAULT &lt;QUERY&gt;
> # sets the n-th file in the file list to the give value
> papis update --set files n:my-file-name.pdf &lt;QUERY&gt;
> papis update --set files n:DEFAULT &lt;QUERY&gt;
> ```

I like this interface. I can't readily think of an improvement. My first
instinct was `papis update --set "list[0]" "string" <query>` but I think
that is clearly inferior.

> @jghauser Like you said, none of the flags in `papis update` allow easily setting a thing to some existing default, so this is not particularly nice. Do you have any better ideas?

Does it makes sense to allow a blank second argument to --set
i.e. `--set thing --set other-thing ...` as a way to ask for `--set thing "DEFAULT"`

Thanks for the great start. I will try to work on this over the next few months.

alexfikl:
> Does it makes sense to allow a blank second argument to --set i.e. `--set thing --set other-thing ...` as a way to ask for `--set thing "DEFAULT"`

Unfortunately, there's no way to tell click to parse 1 OR 2 arguments like that. It needs to be an explicit second argument. We can add some shorthand like `.` or `-` instead of `DEFAULT`, but otherwise..

We can also add a additional flag for "setting to default" like `--reset thing`? I kind of like how explicit that is + less typing. @jghauser @pmiam What do you think?

alexfikl:
FYI, this still needs a bit of work. The main thing that I haven't gotten to is error recovery: it now just renames and moves the files before seeing if the command itself succeeds.

I'm thinking of reworking the various `run_*` methods a bit to be better separate concerns:
1. Gather all the changes that need to be performed on the document from the CLI flags.
	* If anything fails, use `--batch` to decide if it's ok to exit early.
3. Apply changes to a copy of the document.
	* If anything fails, use `--batch` to decide if it's ok to exit early. 
4. If everything succeeded, apply file renames on disk.
5. If everything succeeded, replace the document with the modified copy. Ideally these last two would be a sort of "atomic operation", but not sure how to believably achieve that.

The main idea is to make it more like a database transaction. Like 83% of the code is there already, just needs a bit of moving around. :grin:

jghauser:
> We can also add a additional flag for "setting to default" like --reset thing? I kind of like how explicit that is + less typing. @jghauser @pmiam What do you think?

I like this. It's explicit and easy to explain. One difficulty is that I don't see how we could allow resetting e.g. a specific attached file. Reset would either have one argument and not support this (and then arguably reset both files) or two arguments (but then we'd need a dummy argument for when we don't need to select list items).

All these complications make me wonder if there's not a better way of specifying the fields on which we want to operate and how we'd like to operate on them. We could have had `papis update --key X --item Y --rename Z` and `papis update --key X --append Y` or something. In any case, this is a conversation for another day and possibly not worth the breaking changes :laughing:.

> I see the ideal implementation as equally supporting two options for
> updating metadata/files
> 
> I imagine the first similar to:
> 
> 1. run `papis edit <QUERY>`
> 2. freely edit info.yaml (or just do this somehow, e.g. with sed or wgrep or whatever)
> 3. run `papis update`
> 4. changes to metadata propagate to all paths

@pmiam, I don't think this would work unfortunately. If a user edits the names of two attached files, we have no way of knowing which was renamed from which.

alexfikl:
Ok, this latest push introduced all the changes I wanted from https://github.com/papis/papis/pull/1108#issuecomment-3703484079. All existing tests are passing with no changes!

Remaining issues:
* [x] Removed the `--set key n:value` syntax for lists. It could be nice to re-add that.
* [x] Renaming files is still not great:
	1. [x] Need to copy the renamed files to `/tmp` or something and the copy them back to the document folder to make sure they don't overwrite each other.
	2. [x] Not sure how to recover from errors during renaming? e.g. 3 documents renamed just fine, but the 4th one got an `OSError` of some sort. Would we need to roll back the previous renamings?
* [x] Add separate `Rename` command (semantics aren't the same as remove+append).
* [x] Still need to add some tests for the `--reset` functionality.
* [x] Still need to add some tests for the file renaming.
* [x] Still need to add some tests for the `--set files n:filename.pdf` syntax.

A nice (from my point of view :grin:) new addition is that the order of the command-line flags now matters for `papis update` when operating on the same key. For example,
```
papis update --rename files some-file.pdf some-other-file.pdf --remove files some-other-file.pdf
```
no longer fails, because the file will exist by the time `--remove` gets called. In the current implementation, this was dependent on the order of the flags in the code.

alexfikl:
@jghauser @pmiam This is finally ready for a serious look, I think. It passes all the tests and adds a bunch more + it seems to work like I expect it :grin:

Pretty much all the functionality should be the same as before, I think. There might be some differences in error handling. For the actual renaming functionality, it can now do
```
papis update --reset files QUERY
```
to update all the files in the document to their default `add-file-name` format and rename them on disk to match that. You should definitely back up files if you're playing with it at the moment, but hopefully the tests cover most obvious situations.

Some small remaining items to do:
* [x] Add some examples with the `--reset` flag.
* [x] Add an example with the `--set tags n:my_new_tag` thing, assuming we want to keep that.

papis/commands/update.py
```diff
@@ -50,7 +54,9 @@
         papis update --from bibtex libraryfile.bib
 
   Papis will try to look for documents in your library that match these
-  entries and will ask you for each entry whether you want to update it.
+  entries and will ask you for each entry whether you want to update it. If
+  you are working with BibTeX file, the ``papis bibtex`` command may be more
```
jghauser:
```suggestion
  you are working with a BibTeX file, the ``papis bibtex`` command may be more
```

papis/commands/update.py
```diff
@@ -60,69 +66,92 @@
...
 
     As you might have guessed, the ``--append`` flag needs to know the type of
     the key it is appending to. It does this by looking at the
     :confval:`doctor-key-type-keys` (and :confval:`doctor-key-type-keys-extend`)
-    configuration options. If the key you are appending to is not in that list,
-    the command will fail.
+    configuration options. If the key does not exist in the document nor in that
+    list of key types, then the command will fail (we don't know what you meant).
+
+- You can instead use ``--set``` to isert a value into a list at a given position:
```
jghauser:
```suggestion
- You can use ``--set`` to insert a value into a list at a given position:
```
'instead' seems wrong since `--set` overwrites the value while `--append` adds a new item?

jghauser: I'm still not sure I'm convinced we need this. I think just using `--rename` could achieve the same? Also, if we do want to keep this feature, I'd suggest not speaking of 'inserting' a value. To me this can also be interpreted as adding a new value at the position (without overwriting an existing value).

EDIT: Thinking about this a little more I do see some use in it. With `--rename` you need to know the filename, with the `0:...` syntax, you only need to know the position. And if you're sure that there's just e.g. 1 file attached, this can actually work very well. I'd still not use it because sometimes I do have more files attached, but I see the use. However, I still think 'insert' isn't the right word to describe this. Using 'set' or 'replace' seems less ambiguous.

papis/commands/update.py
```diff
@@ -60,69 +66,92 @@
...
 
         papis update --drop tags
 
-    This removes all tags.
+  This removes all tags.
 
 - There is also a convenience option ``--rename`` if you want to rename
-  a list item. It's equivalent to doing ``--remove`` and ``--append`` sequentially:
+  a list item. It's equivalent to doing ``--remove`` and ``--append``, but as
+  an "atomic" operation:
```
jghauser:
```suggestion
  a single operation:
```

papis/commands/update.py
```diff
@@ -446,44 +812,40 @@ def run(
...
-    "to_set",
+    "-s", "--set", "to_set",
     help="Set the key to the given value.",
     multiple=True,
-    type=(str, papis.cli.FormatPatternParamType()),
+    type=(papis.cli.KeyParamType(), papis.cli.FormatPatternParamType()),
+)
+@click.option(
+    "--reset", "to_reset",
+    help="Reset keys to their default values (e.g. 'ref' is set to 'ref-format').",
```
jghauser:
```suggestion
    help="Reset keys to their default values.",
```
I would keep examples to the docs (soon to be in the man pages :smiley:)

papis/commands/update.py
jghauser: Also, this is a bit confusing to me. Why do we give a bibtex example and then say they should use the `papis bibtex` command? Maybe we should use a different example or explain when to use `papis bibtex` instead of `papis update`.

papis/commands/update.py
```diff
@@ -60,69 +66,92 @@
...
+
+    .. code:: sh
+
+        papis update --reset ref Einstein
+
+  This will reset the reference to the format described by :confval:`ref-format`.
+  Most keys do not support this, as they do not have any well-defined default.
+  Other supported keys are: "author" (gets updated from ``author_list`` and
+  :confval:`multiple-authors-format`), "notes" (using :confval:`notes-name`),
+  "files" (using :confval:`add-file-name`).
```
jghauser:
```suggestion
  This will reset the value to the default value defined for the key. In the above
  case the "ref" is set to the format described by :confval:`ref-format`. Other
  supported keys are: "author" (gets updated from ``author_list`` and
  :confval:`multiple-authors-format`), "notes" (using :confval:`notes-name`),
  and "files" (using :confval:`add-file-name`). Other keys do not support this
  as they do not have any well-defined default.
```

papis/commands/update.py
```diff
@@ -60,69 +66,92 @@
...
 
 - The above can also be achieved with the ``--append`` option:
```
jghauser:
```suggestion
- The ``--append`` option can be used to append a string to an existing string:
```
('above' isn't correct anymore since there's now the `--reset` operation)

papis/commands/update.py
```diff
@@ -60,69 +66,92 @@
...
     .. code:: sh
 
         papis update --append tags physics
 
-    This adds the tag 'physics' to the existing list of tags. If the list
-    doesn't yet exist, it will be created. All duplicate items will be removed
-    from the list.
+    This adds the tag "physics" to the existing list of tags. If the list
+    doesn't yet exist, it will be created. The new tag will only be appended to
+    the list if it does not already exist.
```
jghauser:
```suggestion
    the list if the tag does not already exist.
```

papis/commands/update.py
```diff
@@ -60,69 +66,92 @@
...
 
     As you might have guessed, the ``--append`` flag needs to know the type of
     the key it is appending to. It does this by looking at the
     :confval:`doctor-key-type-keys` (and :confval:`doctor-key-type-keys-extend`)
-    configuration options. If the key you are appending to is not in that list,
-    the command will fail.
+    configuration options. If the key does not exist in the document nor in that
+    list of key types, then the command will fail (we don't know what you meant).
```
jghauser:
```suggestion
    The ``--append`` flag needs to know the type of the key it is appending to.
    If the key exists in the document, then value set in the document determines
    the type. If the key doesn't exist in the document, the command looks at the
    list of types defined in the :confval:`doctor-key-type-keys` and 
    :confval:`doctor-key-type-keys-extend` configuration options. If the type
    cannot be determined in either of these two ways, the command will fail.
```

papis/commands/update.py
```diff
@@ -60,69 +66,92 @@
...
+
+- You can instead use ``--set``` to isert a value into a list at a given position:
+
+    .. code:: sh
+
+        papis update --set files 0:some-new-file.pdf Einstein
+
+  This special syntax for ``--set`` also only works with keys that are known to
```
jghauser:
```suggestion
  This special syntax for ``--set`` only works with keys that are known to
```

papis/commands/update.py
```diff
@@ -60,69 +66,92 @@
...
+  be lists (as before, based on :confval:`doctor-key-type-keys`). If the key is
+  not a list, then the value is treated as a string. This might be unexpected,
+  so make sure you are using it for list keys only.
```
jghauser:
```suggestion
  be lists (as before, based on :confval:`doctor-key-type-keys` and
  :confval:`doctor-key-type-keys-extend`). If the key is not a list, then the
  value is treated as a string. This might be unexpected, so make sure
  you are using it for list keys only.
```

papis/commands/update.py
```diff
@@ -60,69 +66,92 @@
...
 
         papis update --rename tags physics philosophy
 
-  This renames the tag 'physics' to 'philosophy'. Note that this option being a
-  combination of ``--remove`` and ``--append``, it will append the desired values
-  even if the value to be removed didn't exist. Thus, the above command will add
-  the tag "philosophy" even if the tag "physics" didn't exist before the
-  operation.
+  This renames the tag "physics" to "philosophy". Note that, if the original
+  tag doesn't exist, we will not add it to the list.
```
jghauser:
```suggestion
  This renames the tag "physics" to "philosophy". Note that the new tag will
  not be added to the list if the original tag doesn't exist.
```

jghauser (commented):
I finally got the time to check this out. This really improves things a lot. I've only managed to go through the docs part of things and left a bunch of comments. One bigger point is that the file renaming functionality isn't explained at all. I think we should explicitly say when what happens (for which options, when will files be renamed, when deleted, etc).

I do still need to thoroughly test especially the file renaming functionality, which I'll do next. I'm currently super busy with things, so it might take a little longer.

One oddity I've already encountered is the following:

```
❯ papis --lib test update --rename files "bliblibli.pdf" test.pdf raleigh                 nix-shell-env
[INFO] commands.update: [1/1], Gathering metadata changes for document: Clarifying the Opacity of Neural Networks - Raleigh, Thomas and Knoks, Aleks.
[INFO] commands.update: [1/1] Applying metadata changes for document: Clarifying the Opacity of Neural Networks - Raleigh, Thomas and Knoks, Aleks.
[INFO] commands.update: Running doctor auto-fixers on document: 'Clarifying the Opacity of Neural Networks - Raleigh, Thomas and Knoks, Aleks'.
[INFO] commands.doctor: [FIX] Removing file from document: 'test.pdf'.
[INFO] commands.doctor: Auto-fixed 1 / 1 errors!
```
The 'bliblibli.pdf' file exists:
```
❯ ls raleigh-2025-clarifying-the-opacity-of
bliblibli.pdf  info.yaml
```
After the command, the info.yaml doesn't list any attached files anymore and the file on disk is left as is. I expected the file to be renamed.

alexfikl:
> I do still need to thoroughly test especially the file renaming functionality, which I'll do next. I'm currently super busy with things, so it might take a little longer.

Thanks a lot for taking a look! No worries about taking a while too.. I'm also a bit busy for the next couple of weeks, so that's perfectly fine :grin:

papis/commands/update.py
alexfikl: Hm. this example seems to have been here pretty much from the beginning. My guess would be that it's meant to highlight the "importer" functionality. Maybe it would make more sense to use something else (e.g. `--from doi`)?

In the meantime, fixed the typo!

papis/commands/update.py
alexfikl: I'm perfectly fine with removing this! My initial motivation for adding it came from looking through the `--set key some_list` code and thinking "well, I would really like just having an index here".. i.e. no real usecase.

I'll probably remove it and add a little issue about it to see if anyone is interested or has a usecase for something like this?

papis/commands/update.py
alexfikl: Fixed in [d01649a](https://github.com/papis/papis/pull/1108/commits/d01649a81bd4dc32232657a0dfa7f297040e7082).

papis/commands/update.py
alexfikl: Agreed! Fixed in [d01649a](https://github.com/papis/papis/pull/1108/commits/d01649a81bd4dc32232657a0dfa7f297040e7082).

papis/commands/update.py
alexfikl: Fixed in [d01649a](https://github.com/papis/papis/pull/1108/commits/d01649a81bd4dc32232657a0dfa7f297040e7082).

papis/commands/update.py
alexfikl: Fixed in [d01649a](https://github.com/papis/papis/pull/1108/commits/d01649a81bd4dc32232657a0dfa7f297040e7082).

papis/commands/update.py
alexfikl: Fixed in [d01649a](https://github.com/papis/papis/pull/1108/commits/d01649a81bd4dc32232657a0dfa7f297040e7082).

papis/commands/update.py
alexfikl: Fixed in [d01649a](https://github.com/papis/papis/pull/1108/commits/d01649a81bd4dc32232657a0dfa7f297040e7082).

papis/commands/update.py
alexfikl: Fixed in [d01649a](https://github.com/papis/papis/pull/1108/commits/d01649a81bd4dc32232657a0dfa7f297040e7082).

papis/commands/update.py
alexfikl: Fixed in [d01649a](https://github.com/papis/papis/pull/1108/commits/d01649a81bd4dc32232657a0dfa7f297040e7082).

papis/commands/update.py
alexfikl: Fixed in [d01649a](https://github.com/papis/papis/pull/1108/commits/d01649a81bd4dc32232657a0dfa7f297040e7082).

papis/commands/update.py
alexfikl: I've also updated `--append key string` to not actually append if the string already ends with the value from the command-line. Good idea?

See [70e75c5](https://github.com/papis/papis/pull/1108/commits/70e75c5cfdef45df02efa4c48b3d489263e26e42).

alexfikl:
> I finally got the time to check this out. This really improves things a lot. I've only managed to go through the docs part of things and left a bunch of comments. One bigger point is that the file renaming functionality isn't explained at all. I think we should explicitly say when what happens (for which options, when will files be renamed, when deleted, etc).

Agreed that the renaming stuff needs a lot more documentation. I'll try to add some in the coming days.

> One oddity I've already encountered is the following:

I could reproduce that! It should be fixed in the newest version [347959d](https://github.com/papis/papis/pull/1108/commits/347959d99bfc6727dee9aa314a98297cecdf3df5). 

The issue was that the lists were getting modified in place and they were just shallow copies of the document :facepalm: so when it got to renaming `document[files]` and `data[files]` were actually the same (and pointed to the new names).

alexfikl:
@jghauser I had some more time to look at this and I think it's in a good state now. It does more or less all the things that I wanted, has some documentation and a bunch of tests and everything seems to be working. The text below summarizes a bit what the current state is :grin:

Some changes to the workflow (mostly on how it handles errors):
1. If an error occurs with the document keys (missing key, wrong type, etc.)
    * with `--batch`: it reverts it to the original state and moves on to the next key.
    * without `--batch`: it asks the user if it's ok to continue.
2. If an error occurs while moving files around / saving the document
    * the files are all rolled back, even if half of them got renamed successfully and half haven't.
    * with `--batch`: no changes are applied and it moves on to the next document.
    * without `--batch`: it asks the user if it's ok to continue to the next document or exit.

This seems to me like it would result in the least amount of broken documents. Even with `--batch`, if it fails to update a key coherently, it just keeps the one in the document as is (no destructive changes).

For the file renaming, you can now do:
```
papis update --set notes my-new-notes.md QUERY
papis update --set files 0:my-new-file.pdf QUERY
papis update --set files '["my-new-file.pdf"]' QUERY
papis update --rename files old-file-name.pdf my-new-file.pdf QUERY
papis update --reset notes QUERY
papis update --reset files QUERY
```
to set the files to a desired value or to their `add-file-name` format or whatever. All of these cases correctly rename the files on disk (assuming the document already has a file). You can also do
```
papis update --append files my-new-file.pdf QUERY
```
but this won't do anything on the disk, i.e. it assumes the user knows this file exists in the document folder. This seems like enough for most cases that I could think of and it's not doing anything too fancy.

I don't know how many people use this more advanced functionality of `papis update`, so I'm inclined to merge it even with a minimal review and fix issues later :grin:

papis/commands/update.py
```diff
@@ -501,98 +984,143 @@ def cli(
...
     _all: bool,
     sort_field: str | None,
     sort_reverse: bool,
-    to_set: list[tuple[str, str]],
-    to_drop: list[str],
-    to_append: list[tuple[str, str]],
-    to_remove: list[tuple[str, str]],
-    to_rename: list[tuple[str, str, str]],
 ) -> None:
     """Update document metadata."""
```
jghauser, line 999: Should we say that `update` renames files according to changed metadata?

papis/commands/update.py
```diff
@@ -9,8 +9,11 @@
 also tries to sanitise filenames so that they don't contain any problematic
 characters.
 
-Normally, ``papis update`` will abort on encountering an error. If you want to
```
jghauser: I went through the whole doc text and made some changes. Because any other way of adding this here seems impossible with the github UI, I'm just posting the whole thing here. Let's see how that fucks with the comment box ;)

```rst
"""
This command allows you to update the document metadata stored in the
``info.yaml`` file. With it, you can either set individual fields' values manually
or update a document with information automatically retrieved from a variety
of sources. The command also renames files and notes on disk when needed.

Papis formatting patterns and Python expressions can be used. See below examples
for more information. The command also sanitises filenames so that they don't
contain any problematic characters (see :confval:`doc-paths-extra-chars`).

By default, ``papis update`` will prompt you when it encounters an error. If you
want to skip errors automatically and apply as many changes as possible,
use the ``--batch`` flag.

This command is mostly meant to be used to apply sweeping changes across a large
number of documents. If you just want to update a single document, ``papis edit``
might be more appropriate.


Examples
^^^^^^^^

- Search among papers with the tag "classics" and update the author to
  "Einstein, Albert":

    .. code:: sh

        papis update --set author "Einstein, Albert" "tags:classics"

  This will open the picker containing all documents that match the query from
  where you can select the document you want to update.

- Update the journal to "Mass and Energy" for all documents with the journal
  "Energy and Mass":

    .. code:: sh

        papis update --all --set journal "Mass and Energy" "journal:'Energy and Mass'"


  The ``--all`` flag means that the operation is applied to all documents that
  match the query, rather than opening the picker to select individual documents to
  update.

- Add ", Albert" to the author string of a documents matching 'Einstein':

    .. code:: sh

        papis update --set author "{doc[author]}, Albert" Einstein

  The ``papis update`` command tries to format input strings using the configured
  formatter. Here, it is used to get the existing author "Albert" and then add
  the string ", Einstein" to end up with "Einstein, Albert".

- Reset a field to its default value using:

    .. code:: sh

        papis update --reset ref Einstein

  This will reset the value to the default value defined for the field. Here, the "ref"
  is set to :confval:`ref-format`. Other fields that can be reset are: "author" (gets
  updated from ``author_list`` and :confval:`multiple-authors-format`), "notes"
  (using :confval:`notes-name`), and "files" (using :confval:`add-file-name`). Other
  fields do not support this as they do not have any well-defined default.

- The ``--append`` option can be used to append to a string value:

    .. code:: sh

        papis update --append author ", Albert" Einstein

  This appends ", Albert" to the existing author value. Note that it will be
  appended to the existing value only if it does not already end with that
  exact string (case-sensitive). This avoids appending duplicates by mistake, as
  in the case of lists (below).

- You can also append an item to a list:

    .. code:: sh

        papis update --append tags physics 'author:einstein'

    This adds the tag "physics" to the existing list of tags. If the list
    doesn't yet exist, it will be created. The new tag will only be appended to
    the list if the tag does not already exist (as an exact case-sensitive match).

    The ``--append`` flag needs to know the type of the field it is appending to.
    If the field exists in the document, then the value set in the document
    determines the type. If the field doesn't exist in the document, the command
    looks at the list of types defined in the :confval:`document-field-types`
    (and :confval:`document-field-types-extend`) configuration option. If the
    type cannot be determined in either of these two ways, the command will
    fail.

- To remove an item from a list, use ``--remove``. For example, to remove the
  "physics" tag from the list of document tags, use

    .. code:: sh

        papis update --remove tags physics 'author:einstein'

- To remove a field entirely, use ``--drop``. For example, to remove all
  tags from documents use

    .. code:: sh

        papis update --drop tags 'author:einstein'

- There is also a convenience option ``--rename`` if you want to rename
  a list item. It's equivalent to doing ``--remove`` and ``--append``, but as
  a single operation:

    .. code:: sh

        papis update --rename tags physics philosophy

  This renames the tag "physics" to "philosophy". Note that the new tag will
  not be added to the list if the original tag doesn't exist.

- The ``--batch`` flag suppresses interactive prompts and skips any document
  for which an update operation has failed. This is useful when applying the same
  operation across many documents where some may not have the expected keys or values:

    .. code:: sh

        papis update --all --batch --remove tags obsolete "tags:obsolete"

  Without ``--batch``, if a document does not have "obsolete" in its tag list
  (or has no ``tags`` key at all), the command pauses and asks whether to
  continue. With ``--batch``, an error is logged, the document is skipped and the
  command moves on to the next one.

  More specifically, ``--batch`` suppresses prompts for the following errors:

  * An operation on a field fails (e.g. trying to ``--remove`` a value that is
    not in the list, ``--append`` to a key of unknown type, or a type
    conversion error from ``--set``). The affected key retains its original
    value and the remaining keys for that document are still processed.
  * A file rename fails (e.g. the file on disk is missing or cannot be moved).
    The whole document is skipped.

  In all cases the command exits with a non-zero status (indicating a failure) if
  any document was skipped.

- Update a document automatically and interactively (searching by DOI in
  Crossref or in other sources...):

    .. code:: sh

        papis update --auto "author:dyson"

- Update your document from a DOI using the importer functionality:

    .. code:: sh

        papis update --from doi '10.1103/PhysRev.47.777' 'author:einstein'

  For a list of all supported importers use ``papis update --list-importers``.

- When you update the ``files`` or ``notes`` fields, the corresponding files on
  disk are renamed to match the new value:

    .. code:: sh

        papis update --set notes "my-new-notes.tex" Einstein

  This updates the ``info.yaml`` file and renames the ``notes`` file on disk from
  its current name to ``my-new-notes.tex``. The same applies when updating ``files``:

    .. code:: sh

        papis update --rename files einstein-old.pdf einstein-new.pdf Einstein

  This renames the file ``einstein-old.pdf`` to ``einstein-new.pdf`` in the
  ``info.yaml`` file and on disk. If the rename fails (e.g. the file does not exist
  or there is a permissions error), the metadata is not updated and an error is
  reported.

  Note that the ``--reset`` option also triggers a rename when used with
  ``notes`` or ``files``. It renames the files to the names derived from the
  configured :confval:`notes-name` and :confval:`add-file-name` formats
  respectively.

Advanced Examples
^^^^^^^^^^^^^^^^^

- As an advanced feature, ``papis update`` also supports the parsing of Python
  expressions (such as lists or dictionaries). This can be used as follows:

    .. code:: sh

        papis update --set author_list "[{'family': 'Einstein', 'given': 'Albert'}]"

  Because the above string is a valid Python expression, ``author_list`` is
  updated to a list that contains a dictionary.

- You can use ``--set`` to set a value in a list at a given position:

    .. code:: sh

        papis update --set files 0:some-new-file.pdf Einstein

  This command will rename the first entry in the list of files. This special
  syntax for ``--set`` only works with keys that are known to be lists (as
  determined based on :confval:`document-field-types` and
  :confval:`document-field-types-extend`). If the field is not a list, then the
  value is treated as a string. This might be unexpected, so make sure you are
  using it for list keys only.
```

jghauser (commented):
Finally getting around to looking at this. It's looking really nice! I only have some small nitpicks/questions. I have to admit I didn't go through the code in detail, but I think you're much better than me at doing that part well anyway haha. Posting this now so that we can discuss these points at least, but I'll try take another look soon if you think it's needed.

## Somewhat confusing `Continue` prompt

When trying to rename a non-existent file, I'm prompted `Updating document failed with the above error. Continue? (Y/n):` but it wasn't clear to me what we're continuing with. When doing `Y`, the messages say that metadata is applied, but I don't think it is (or rather it is but I guess the metadata that is applied is the empty set)? When doing `n`, then no messages are shown. I'm guessing that if there had been other metadata changes in this invocation of `papis update`, there would have been differences. Maybe we could make the message a bit more obvious so the user knows what we're continuing with and/or check if there aren't any metadata changes to be applied?

```
papis update-rename-files*​ ≡
❮ papis --lib test update --rename files my-new-file.pdf bli.pdf shea                     nix-shell-env
[INFO] commands.update: [1/1] Gathering metadata changes for document: The Ethical Standard for End-of-Life Decisions for Unrepresented Patients - Shea, Matthew.
[ERROR] commands.update: Failed 'Rename' operation on key 'files': key 'files' does not contain 'my-new-file.pdf' (doc: The Ethical Standard for End-of-Life Decisions for Unrepresented Patients - Shea, Matthew).
Updating document failed with the above error. Continue? (Y/n):
[INFO] commands.update: [1/1] Applying metadata changes to document: The Ethical Standard for End-of-Life Decisions for Unrepresented Patients - Shea, Matthew.
[INFO] commands.update: Running doctor auto-fixers on document: 'The Ethical Standard for End-of-Life Decisions for Unrepresented Patients - Shea, Matthew'.

papis update-rename-files*​ ≡40s 
❮ papis --lib test update --rename files my-new-file.pdf bli.pdf shea                     nix-shell-env
[INFO] commands.update: [1/1] Gathering metadata changes for document: The Ethical Standard for End-of-Life Decisions for Unrepresented Patients - Shea, Matthew.
[ERROR] commands.update: Failed 'Rename' operation on key 'files': key 'files' does not contain 'my-new-file.pdf' (doc: The Ethical Standard for End-of-Life Decisions for Unrepresented Patients - Shea, Matthew).
Updating document failed with the above error. Continue? (Y/n): n
```

## Somewhat unexpected `--batch` functionality

I think the code currently applies all changes to a doc as long as all operations for a given key are successful. This is a bit unexpected since `--batch` usually applies *all* successful operations. For instance, I'd expect `papis update --remove tags physics --append tags philosophy` to succeed adding 'philosophy' even if the tag 'physics' doesn't exist. Though I'm not sure we invite unpredictabilities of other kind if users expect a certain order of successful operations (though I can't really see why they would do that). Either way, we should be explicit about it in the docs.

papis/commands/update.py
alexfikl: Meh, I think it's fine for this to be vague, since it's just a small text that shows on `--help`.

papis/commands/update.py
alexfikl: Chug this in there: [18108b2](https://github.com/papis/papis/pull/1108/commits/18108b2639c05ed28619e9f578cd9e7f094efe75).

For adding these things when Github doesn't want to work with you, I've seen people just copy-paste the patch itself (`git diff > changes.patch`) in here :grin:

jghauser: Good idea, I'll do that next time!

alexfikl:
@jghauser I've updated this a bit.

> Somewhat confusing Continue prompt

This now shows `Updating document key '{key}' failed. Skip the failed change and continue updating the document?`. That sounds clearer to me. What do you think? :grin:

It also doesn't show the "Applying changes" if there are no changes to apply + added a little message at the end to let the user know it's done. Hopefully it looks a bit better now.

> Somewhat unexpected --batch functionality

Yeah, I was torn between making `--batch` leave keys alone if the updates fail, so that it doesn't leave things in a "middle state", and your suggestion. I've changed it to your suggestion now.

jghauser:
Seems to work very well! Only small nit I have is that I get a `Updating document key 'files' failed. Skip the failed change and continue updating the document?` (or similar) even when there are no other updates to apply. It would be more elegant if we just returned. Definitely not a blocker though!

alexfikl:
> Seems to work very well! Only small nit I have is that I get a `Updating document key 'files' failed. Skip the failed change and continue updating the document?` (or similar) even when there are no other updates to apply.

When is that happening? Definitely want to fix that :D

jghauser:
I think it happens whenever I try to do a single operation which fails, e.g.

```
❯ papis --lib test update --rename files some-file.pdf bli.pdf shea
[INFO] commands.update: [1/1] Gathering metadata changes for document: The Ethical Standard for End-of-Life Decisions for Unrepresented Patients - Shea, Matthew.
[ERROR] commands.update: Failed '--rename' operation on key 'files': key 'files' does not contain 'some-file.pdf' (doc: The Ethical Standard for End-of-Life Decisions for Unrepresented Patients - Shea, Matthew).
Updating document key 'files' failed. Skip the failed change and continue updating the document? (Y/n):
```

Here, the `some-file.pdf` doesn't exist, which is why the operation fails. There is nothing else to update.

Actually, I'm just realising that the 'doc: XYZ' part of the error message might be redundant given that there's an info message that already informs the user about it. Though I guess maybe being explicit has its advantages too... And while we're nitpicking, maybe 'key' should be 'field' (after all, the key itself is left alone)? :smile:

alexfikl:
> I think it happens whenever I try to do a single operation which fails, e.g.

Ah, I see! :facepalm: I'm not sure what the best way to fix that is though. The issue is that "continue" here means "keep the changes we've made so far and continue", so if you say no the whole document will be skipped.

Maybe these should be two separate messages? 
1. One to go through the different operations and stop when the user had enough (current message, rephrased maybe).
2. One to go "This document had errors. Continue with it?".

Or just do `if is_last: (2) else: (1)`?

> I'm just realising that the 'doc: XYZ' part of the error message might be redundant given that there's an info message that already informs the user about it. 

The reason I added that there was because I was thinking that if I have 10 errors or so, the document name will be high up enough that I lose track of it. But yeah, not sure I'm happy with some of this output.. it's probably too noisy when there's a lot of errors anyway :\

> And while we're nitpicking, maybe 'key' should be 'field' (after all, the key itself is left alone)? 😄

Yeah, fair enough. I'm still a bit confused with this nomenclature though (mostly mismatch with Python's key+value). In this message, which ones would be fields? All of them?
```
[ERROR] commands.update: Failed '--rename' operation on key 'files': key 'files' does not contain 'some-file.pdf'.
```

jghauser:
> > I think it happens whenever I try to do a single operation which fails, e.g.
> 
> Ah, I see! 🤦 I'm not sure what the best way to fix that is though. The issue is that "continue" here means "keep the changes we've made so far and continue", so if you say no the whole document will be skipped.

Gotcha! Yeah, in this instance this is a bit confusing because no changes have been made to the document at all. Only one operation was attempted and failed. Directly returning is what I would have expected.

> Maybe these should be two separate messages?
> 
>     1. One to go through the different operations and stop when the user had enough (current message, rephrased maybe).
> 
>     2. One to go "This document had errors. Continue with it?".
> 
> 
> Or just do `if is_last: (2) else: (1)`?

I wonder how fine-grained the prompting should be. Should we just have a single prompt at the end of all operations on a doc that says something like: This document has errors, apply what's worked or discard changes. And if there's no changes that worked, don't even show the message?

We can prompt for each operation, but maybe that's just overcomplicating things? I'm not sure...

> > I'm just realising that the 'doc: XYZ' part of the error message might be redundant given that there's an info message that already informs the user about it.
> 
> The reason I added that there was because I was thinking that if I have 10 errors or so, the document name will be high up enough that I lose track of it. But yeah, not sure I'm happy with some of this output.. it's probably too noisy when there's a lot of errors anyway :\

Yeah probably will get hard to parse at that point anyway. I think the more common case is probably at most a few errors per doc, which should make it easy-ish to identify the doc in question.

> Yeah, fair enough. I'm still a bit confused with this nomenclature though (mostly mismatch with Python's key+value). In this message, which ones would be fields? All of them?
> 
> ```
> [ERROR] commands.update: Failed '--rename' operation on key 'files': key 'files' does not contain 'some-file.pdf'.
> ```

So my understanding of this is (and I think in line with e.g. bibtex's talk of fields) is that a field is simply a key+value pair. It's a bit unfortunately that field isn't really a term in Python and that in Papis we're currently using a bit of both :sweat_smile:. The way I was thinking about it is that in Papis we usually operate on fields by changing the value associated with a key. This is what the message above is about: we want to change the 'files' field by changing the value associated with the key 'files'. I guess on that understanding, a key cannot 'contain' a filename, at most the value does (or to be really pedantic, the field contains the filename while the value *is* the filename haha). Sometimes we do change keys (which turns one field into another), maybe by converting an 'author' into an 'editor' or something.

alexfikl:
> We can prompt for each operation, but maybe that's just overcomplicating things? I'm not sure...

Yeah, went with that. We can add finer control if someone complains :grin:
 
> Yeah probably will get hard to parse at that point anyway. I think the more common case is probably at most a few errors per doc, which should make it easy-ish to identify the doc in question.

Removed!
 
> So my understanding of this is (and I think in line with e.g. bibtex's talk of fields) is that a field is simply a key+value pair.

I went through it and update all the uses I could find. I think it's in line with this now. Let me know if you spot anything!

alexfikl:
@jghauser Do you want to have another look at this? It seems to be in a reasonable state, so I would be inclined to merge it and fix things as they come :grin:

jghauser:
Played with it for a bit and it looks good to me! :tada:

alexfikl:
> Played with it for a bit and it looks good to me! 🎉

Awesome! Thanks a lot for all the help with it :grin: Can't wait to go and do a little `papis update --reset ref` through my whole library, haha.

Issue #1095, opened 2025-11-10
FR: shell completions for entries


retroflexivity:
The finder-centered approach works well but is somewhat unorthodox for shell programs. I guess many people (me including) would rather press tab at the prompt and get the list of entries. 

Given that papis already picks the entry automatically in case of a unique match, all that is needed is to supply unique completions (e.g. folder names) for all entries on db sync. The only problem is to support the vast number of shells, but I could at least do that for fish.

alexfikl:
I'm not sure I follow exactly what you're suggesting here. Something like
```
papis list Einst<TAB>
```
and it automatically shows all the documents that match in the shell's completion system? I'm not quite sure how that would work, but if you have some ideas for fish, I think we can at least include them in `contrib`, if you want to share :grin:

tuurep:
Yeah I imagine there's difficulty in this because the query to match the document can be pretty much anything,

see the option [match-format](https://papis.readthedocs.io/en/latest/configuration.html#confval-match-format)

But I can see something in your suggestion that I really love as an idea:

I most commonly query documents using a short ref string `<lastname><year>`

If this was tab completable it would be a really nice way skip the picker phase

But in relation to the whole system, doesn't seem as simple as this

Issue #903, opened 2024-09-01
Allow git-style subcommand aliasing


tuurep:
Discussed in https://github.com/papis/papis/discussions/898

When using git I heavily use aliases like this:

`~/.gitconfig`

```gitconfig
[alias]
        s = status
	a = add
	c = commit
        ca = commit --amend
        p = push
        d = diff
        ds = diff --staged
        rs = restore --staged
        r = remote
        b = branch
        co = checkout
        l = log
```

Would really wish to do the same in papis :)

Turns out that the aliasing in git can get very flexible and complex, which I didn't know.

But for now I'm basically interested in only simple substitutions such as in my above example.

Also important to know that I learned from Alex is that it's already possible to shorten any subcommand as long as it's an unambiguous prefix of some command, for example:

`$ papis op` works the same as `$ papis open`

(Looks like it doesn't accept one-letter prefix in any case, however?)

Still I would find the git-style aliasing a great feature because of how frequently you need certain subcommand plus flag pairs, similarly to when using git, something like:

`$ papis add --from doi`

Pointers by @alexfikl:

> For the git-style aliases it should probably go in the `get_command` method of the `ScriptLoaderGroup` here:
> https://github.com/papis/papis/blob/95e0b5456afef60ee4f54c4aaec624244093c1e0/papis/commands/default.py#L66-L78
>        
> You can see there that we're trying to load entrypoints (that have `script.plugin`) and from various external scripts (either in your path or in `~/.config/papis/scripts`). You could probably extend that piece of code to also load the config section `[alias]` and try to run those instead.

alexfikl:
For how the current automatic aliasing is done see the `AliasGroup` class here:
https://github.com/papis/papis/blob/95e0b5456afef60ee4f54c4aaec624244093c1e0/papis/commands/__init__.py#L15-L23

tuurep:
I've been hating two things about the automatic prefix aliases:

1. Doesn't allow 1 char, e.g. `papis o` for open, only `op`
2. It keeps yelling at me, though I use it intentionally (a lot):
    - `Command 'op' is unknown! I suppose you meant: 'open'`

But it has still been useful enough that I haven't prioritized this issue yet. Before full-blown git-style command aliases, could give a quick patch for the above two points, if there's interest.

Dilemma: I'd like it to be completely silent, but it's legitimately useful for new users who don't know it works that way.

Strong opinion: it should be silent anyways, and document the behavior somewhere else.

tuurep:
[...]

PR #1112, opened 2025-12-19
add document autocompletions

retroflexivity:
this addresses #1095, adding shell completions with click (so works in bash, zsh and fish).

currently, the completions are hard-coded to be the folder names. this works ok because folder names are unique. however, it requires match_pattern to contain {doc.subfolder}.

this could be turned into a format pattern and made customizable, but Document currently has no property for the folder name.

image

### DIFF
--- a/doc/source/default_settings.rst
+++ b/doc/source/default_settings.rst
@@ -139,8 +139,11 @@ General settings
 
 .. papis-config:: completion-format
 
-    Format for document autocompletion. It is triggered by pressing Tab when
-    entering a search query in the command line, e.g. `papis addto ein<TAB>`.
+    Format for document autocompletion. It is used to construct the completion
+    value that is given to the shell when pressing Tab after entering a
+    search query in the command-line. For example, writing ``papis addto ein<TAB>``
+    will query the database for the ``ein`` string, but the completion will
+    be constructed using ``completion-format`` from the matching documents.
 
 .. papis-config:: completion-help-format
 
--- a/doc/source/shell_completion.rst
+++ b/doc/source/shell_completion.rst
@@ -60,14 +60,22 @@ Document completions
 --------------------
 
 Commands that expect a document query (``edit``, ``addto``, etc.) support
-document completions.
+document completions. These are mainly controlled by the :confval:`completion-format`
+configuration setting and can be be triggered on e.g.
 
-Normally, all documents that match the incomplete query via
-:confval:`match-format` are returned, for the shell to filter them. If the shell
-expects completions already filtered to those that have the incomplete string as
-a prefix, :confval:`prefix-only-completions` can be used.
+.. code:: bash
 
-For example, ``zsh`` replaces the incomplete string with the longest common
-prefix of the suggested completions, which is not guaranteed to be longer than
-or equal to the incomplete string. To prevent losing the incomplete string,
+   papis edit ein<TAB>
+
+Normally, all documents that match the *incomplete* query via :confval:`match-format`
+are returned and the user can filter them further using the shell's capabilities.
+
+To force the results to only include matches that have the *incomplete* query as a
+prefix, use :confval:`prefix-only-completions`. For example, ``zsh`` automatically
+replaces the *incomplete* string with the longest common prefix of the suggested
+completions, which can break your workflow. To prevent losing the incomplete string,
 enable :confval:`prefix-only-completions`.
+
+Note that the document completion only sets the value of
+:confval:`completion-format` for the selected document as the query. This can
+still bring up the Papis picker if the selection is not sufficiently unique.
--- a/papis/cli.py
+++ b/papis/cli.py
@@ -79,18 +79,29 @@ def bool_flag(*args: Any, **kwargs: Any) -> DecoratorCallable:
 
 
 def _clean_completions(s: str) -> str:
-    """Remove certain characters that are treated as completion breaks.
-    See http://tiswww.case.edu/php/chet/bash/FAQ, E13 for the colon.
-    The comma causes the same problem in fish for a reason."""
+    """Remove certain characters that are treated as completion breaks."""
+
+    # NOTE: See http://tiswww.case.edu/php/chet/bash/FAQ, E13 for the colon.
+    # NOTE: The comma causes the same problem in fish for a reason.
     return re.sub(r"[:,]", "", s)
 
 
-# NOTE: default completion format (`doc.ref`) is not guaranteed to be unique
-# TODO: prevent a completion for A from matching B when fmt(A) is infix of fmt(B)
 def _query_shell_complete(ctx: click.Context,
                           param: click.Parameter,
                           incomplete: str) -> list[CompletionItem]:
     """Return completion items that match a given incomplete document query."""
+
+    # NOTE: some known quirks of the current completion setup:
+    # - we select a query based on `completion-format`, which need not be unique
+    #   so the picker can still open after the selection.
+    # - the query is searched in the database using `match-format`, but only
+    #   `completion-format` is sent to the shell.
+    # - see other discussion in https://github.com/papis/papis/pull/1112
+
+    # TODO:
+    # - prevent a completion for A from matching B when fmt(A) is infix of fmt(B)
+    # - allow selecting based on the document folder (need to allow access to it
+    #   in the formatting)
 
     fmt = papis.config.getformatpattern("completion-format")
     help_fmt = papis.config.getformatpattern("completion-help-format")
@@ -99,32 +110,26 @@ def _query_shell_complete(ctx: click.Context,
     lib = ctx.parent.params.get("lib") if ctx.parent else None
     query = (
         incomplete if incomplete
+        # return all documents on empty query
         else papis.config.getstring("default-query-string")
     )
 
-    # suppress logging of "Setting path {} as the main library folder"
-    # when completions requested for an unnamed library
-    with papis.logging.quiet("papis.config", level=logging.ERROR):
+    # NOTE: suppress all logging to avoid spamming the screen during completion
+    with papis.logging.quiet("papis", level=logging.ERROR):
         comps_and_helps = (
             (
                 _clean_completions(format(fmt, doc)),
                 _clean_completions(format(help_fmt, doc))
-            ) for doc in
-            handle_doc_folder_or_query(
-                # return all documents on empty query
-                query,
-                None,
-                library_name=lib
-            )
+            ) for doc in handle_doc_folder_or_query(query, None, library_name=lib)
         )
+
     return [
         CompletionItem(
             comp,
             help=help
         )
         for comp, help in comps_and_helps
-        # if prefix_only, only include those completions
-        # that start with the query
+        # if prefix_only, only include matches that start with the query string
         if not prefix_only or comp.startswith(incomplete)
     ]
 


jghauser:
Thanks, I really like this! Papis shell completion has been in need of some attention for a while!

I would appreciate if this could be made configurable and maybe default to the `ref`? That should be unique as well (if the library is in good order) and (at least to me) is a more standard way to refer to a document.

retroflexivity:
> I would appreciate if this could be made configurable and maybe default to the `ref`? That should be unique as well (if the library is in good order) and (at least to me) is a more standard way to refer to a document.

great idea! the only problem is that `match-format` doesn't include `ref` by default. if it's ok to change this, i would prefer this, too

jghauser:
I'm ok with that, what do you think @alexfikl?

papis/cli.py
```diff
@@ -75,12 +76,29 @@ def bool_flag(*args: Any, **kwargs: Any) -> DecoratorCallable:
         **kwargs)
 
 
+def query_shell_complete(ctx: click.Context,
```
alexfikl:
```suggestion
def _query_shell_complete(ctx: click.Context,
```
? I.e. make this private looking.

doc/source/default_settings.rst
```diff
@@ -137,6 +137,10 @@ General settings
     non-intrusive way. Preferably, this should be a short string that allows
     easily identifying which document is being referenced.
 
+.. papis-config:: completion-help-format
+
+    Format for the tip displayed next to document completions in the command line upon pressing Tab. Not all shells support this.
```
alexfikl: Wrap this a bit to 80-ish characters to match the rest of the document?

Also, maybe give some examples of where this is used? e.g. something like "These completions appear when using e.g. `papis addto ein<TAB>`".

papis/cli.py
```diff
@@ -75,12 +76,29 @@ def bool_flag(*args: Any, **kwargs: Any) -> DecoratorCallable:
...
+                         param: click.Parameter,
+                         incomplete: str) -> list[CompletionItem]:
+
+    docs = handle_doc_folder_or_query(incomplete, None)
```
alexfikl: What happens if there's no completion here, i.e. just `papis addto <TAB>`? It feels like it may need to fall back to `default-query-string` or something in that case.

jghauser:
> Generally `refs` won't be unique though. Would that still work?

Aren't they supposed to be unique? As in, duplicate `ref`s are something that `papis doctor` warns about I think; `ref`s *should* be unique. I think `ref`s are a much more intuitive way to refer to documents than folder names (importantly they are used when inserting references into documents). If having duplicate `ref`s just leads to opening up the picker to disambiguate, then I'd still prefer it. The user should really just fix their library in that case.

alexfikl:
> Aren't they supposed to be unique? As in, duplicate `ref`s are something that `papis doctor` warns about I think

The `duplicate-keys` check isn't enabled by default (at least my reasoning for avoiding that was that it's quite slow for a bigger library), so that's not enforced. Also, a unique `ref` seems to mostly be a requirement for BibTeX (and Typst?) exports, not necessarily a general thing (we have the `papis_id` as a unique identifier for when that's needed?).

Even if the `ref` was unique, if we allow the user to provide a configurable format pattern, it's not guaranteed that will be unique, so we need to handle that case somehow in the completions regardless.

papis/cli.py
```diff
@@ -75,12 +76,29 @@ def bool_flag(*args: Any, **kwargs: Any) -> DecoratorCallable:
...
+    help_fmt = papis.config.getformatpattern("completion-help-format")
+
+    return [
+        CompletionItem(
+            doc.get_main_folder_name(),
+            help=format(help_fmt, doc)
```
alexfikl: This may need to be truncated? 

I just tried the completions on a lot of documents and they looked very wonky (i.e. the help part was missing).. my guess is that happened because the "query + (format)" didn't fit on on line of my terminal.

retroflexivity:
> Any feel for whether this slows down the completions at all? They're already quite slow, so I'm always weary of that :\

this is all lazy, and the function should only be run on pressing tab inside a subcommand like `papis edit`, so it shouldn't affect other completions. i haven't noticed any slowdowns either

retroflexivity:
> Generally `refs` won't be unique though. Would that still work?

what this does is just fill the query. the query is evaluated as usual: the finder is opened when it is not unique.

the finder opening wastes the point of completions somewhat of course, but the case is uncommon and the effect isn't horrible. also, any combination of fields can be set as the completion format, up to equal to the match format

retroflexivity:
> Typing `papis edit carlson<TAB>` (in zsh) gets confused and switches to `papis edit 20<TAB>` instead. Not quite sure how to make it work nicely..

seems like papis returned entries with "carlson" from the database, then zsh filled the prompt with the longest common prefix. i guess it can be configured in zsh

papis/cli.py
retroflexivity: could you share a screenshot please? fish for instance handles trimming itself

alexfikl: Normally it looks like this, i.e. the help text appears and is properly trimmed
Screenshot_20260101_104635

But sometimes (when there's a lot more entries?) it looks like this
Screenshot_20260101_104558

As far as I can tell, I have one entry that has a very long folder name and it just pushes everything outside. Maybe we can truncate the folder name at least? Does fish do that too?

I'm using zsh (with quite a few random plugins) and fzf, so it's quite likely this is just a weird setting/bug I haven't hit yet.

alexfikl:
> what this does is just fill the query. the query is evaluated as usual: the finder is opened when it is not unique.

What I was mostly thinking is that if using e.g. a non-unique ref, one of the entries will be shadowed (i.e. not show up in the completion suggestions). This could be confusing and make you think that that entry just doesn't exist.

EDIT: Nevermind, actually tested this and they both just show up, so it seems to be fine, as you said :grin:

> seems like papis returned entries with "carlson" from the database, then zsh filled the prompt with the longest common prefix. i guess it can be configured in zsh

Yeah, that seems to be the case. I'll look into it to see if there's some configuration.

papis/cli.py
alexfikl: For what it's worth, I'm getting no suggestions at all with `papis addto <TAB>`. `papis addto .<TAB>` seems to show everything though (because `.` is the default query string for the `papis` backend).

papis/cli.py
alexfikl: For what it's worth, this is not an issue for me when using the `ref` as a match, since those are generally a lot shorter (I use a `FirstAuthorYear` format).

retroflexivity:
> What I was mostly thinking is that if using e.g. a non-unique ref, one of the entries will be shadowed (i.e. not show up in the completion suggestions).

seems to be the case. not sure what to do with it apart from adding a note in the wiki

retroflexivity:
another problem is that there's no guarantee of a unique match even if refs are unique.
like, if ref a is contained in match pattern b, selecting completion a will bring up the picker with both a and b

i know that there's the field:query syntax, so we can limit some unexpected matches by setting the completion pattern to `ref:{{ doc.ref }}`. still, uniqueness would fail with refs like `author2000` and `author2000a` (which is quite a common case). do we have something like `exact:query` syntax?

papis/cli.py
retroflexivity: > Maybe we can truncate the folder name at least

this would mean that the inserted string is truncated as well, which is not what we want, i guess

alexfikl: Hm, that's a good point. I keep forgetting there shown string and the search string are the same :(

I'll try to look if there's some way to configure this in zsh/fzf.

alexfikl:
> seems to be the case. not sure what to do with it apart from adding a note in the wiki

I don't know about the wiki, but I think it's worth adding some short notes to the code to make sure we remember most of the choices / shortcomings we found so far?

Otherwise, I'm ok with merging this. @jghauser Any thoughts?

> do we have something like `exact:query` syntax?

Afraid not, the default search syntax is pretty basic :( I think the whoosh backend has some fancier regex support, but I haven't used it in a while.

papis/cli.py
alexfikl: @retroflexivity Have you had a chance to look at this?

retroflexivity:
should I set the default to `ref:{{ doc.ref }}` then?

advantages:
- disambiguation in some (pretty uncommon though) cases when ref is matched by some other fields of another item
- ~no need to change the default `match_pattern`~ **UPD**: just realized that removing ref from match_pattern breaks suggestion generation

disadvantages:
- uglier completions

alexfikl:
> should I set the default to `ref:{{ doc.ref }}` then?

I can't say I have a good feel for that, so whatever feels more natural to you (as an actual user of this) sounds good! We can always come back and update it after you (and others) play with it a bit more.

papis/cli.py
retroflexivity: agree!

retroflexivity:
> I can't say I have a good feel for that, so whatever feels more natural to you (as an actual user of this) sounds good! We can always come back and update it after you (and others) play with it a bit more.

let's stick with the first option then

alexfikl (approved):
:+1: This looks good to me! Lets give @jghauser a chance to take a look and then we can merge it.

tuurep:
Hey these completions are super neat, I'm giving the PR a try,

Here are some things I immediately notice:

---

For most of my docs I get the author listed as this:

image

It may be that it's on me to set the format setting, I'll take a look. I've been setting `author` as that for the purpose of autogenerating the ref from the name.

**Solved:** yes, what I want to do here is to set `completion-help-format` to the same as, or similar, as either my `header-format` or `fzf-header-format` where I untangle that author field situation. 

---

There's some ref where when I try to complete it, it instead clears the string:

`papis open berg<Tab>`

`berg` is an ambiguous prefix for a ref, here's an example doc:

```yaml
author: '{au[family]}, {au[given]}'
author_list:
- family: Berg
  given: Christopher
files:
- mastering-guitar-technique-process-and-essence.pdf
isbn: 1-6106-5058-1
isbn-13: 978-16106-5058-8
notes: mastering-guitar-technique-process-and-essence.md
papis_id: 5961c8b34ebdbbc8dcffc2259301bf9b
publisher: Mel Bay Publications, Inc.
ref: berg2008
time-added: 2025-12-21-05:21:30
title: 'Mastering Guitar Technique: Process and Essence'
year: '2008'
```

**Edit**: looks like I also reproduce it with most very short prefixes of 1 or 2 characters (but not all), I think there is something in certain docs that makes it bail?

**Edit**: finding the cause for this is confusing but I'm trying. Let me know if anyone can reproduce for example: `papis open a<Tab>` just erases the prefix. Suspecting some particularly malformed document.

**Edit**: looks like in the case of `berg`, it wants to simultaneously complete it to `berg20`, and `schedlberger2024`. Is this the cause, and maybe also what you were discussing with the infix matching problem?

Yeah the TODO in here seems related: https://github.com/papis/papis/pull/1112/changes/92fa1ca103771777ae0826656c5312e121a7f141

**Edit**: see https://github.com/papis/papis/pull/1112#issuecomment-3706917051

---

The completions don't seem to respect `-l somelibrary` to only complete entries in the given library. I think this would be very useful.

Suggested fix: https://github.com/papis/papis/pull/1112#discussion_r2659252716

tuurep:
Is it intentional that I can complete a prefix (or infix) of a title into a refname, something like this?

https://github.com/user-attachments/assets/5942c4d2-7dcf-412f-9aad-b40256030de7

tuurep:
Nitpick & probably out of scope: I often (mis?)use the shorter names for commands like:

`papis op`
`papis ed`

In these cases the completions don't work

But I've been meaning to look into the state of the command abbreviation more in #903

retroflexivity:
> Is it intentional that I can complete a prefix (or infix) of a title into a refname, something like this?
> demo.mp4

it is intentional. suggestions are generated by matching on the usual `match_pattern`. if you think this is counterintuitive, we can consider adding an option to match on `completion_pattern` / making the pattern customizable (might also be useful on low-end machines)

retroflexivity:
> **Edit**: looks like in the case of `berg`, it wants to simultaneously complete it to `berg20`, and `schedlberger2024`. Is this the cause, and maybe also what you were discussing with the infix matching problem?
> 
> Yeah the TODO in here seems related: [92fa1ca](https://github.com/papis/papis/commit/92fa1ca103771777ae0826656c5312e121a7f141)

no, the todo is another thing, about how completing doesn't guarantee that the picker won't open. your case is the same as https://github.com/papis/papis/pull/1112#pullrequestreview-3620581554.

the problematic thing is imho that the shell replaces the string with the longest common prefix of matches, not that infix completion get generated. but if it's a common case, let's add a "dumb" completion mode that matches prefixes of `completion_pattern`only

tuurep:
Thank you

Yeah honestly the string clearing case is so frequent to me that it's going to condition me to avoid using the completions at all :smile: with many commonly used 3-character prefixes

> let's add a "dumb" completion mode that matches prefixes of `completion_pattern` only

Could be that this was my expectation of what should be happening

On https://github.com/papis/papis/pull/1112#issuecomment-3706904855:

On intuition I'd have wanted "starting str" to complete into "Starting strength - basic barbell training" (doc.title), but this seems complicated in many ways: requires multiple types of completion item and could bloat the result lists too much?

retroflexivity:
turns out customizing matching rules is harder than i thought, so here's a very primitive implementation with a list filter. not particularly efficient but does the job.

as a side note, zsh-autocomplete seemingly allows configuring this https://github.com/marlonrichert/zsh-autocomplete?tab=readme-ov-file#first-insert-the-common-substring

retroflexivity:
> On intuition I'd have wanted "starting str" to complete into "Starting strength - basic barbell training" (doc.title), but this seems complicated in many ways: requires multiple types of completion item and could bloat the result lists too much?

i guess this could be implemented. generate the list of titles, authors, refs, join them together, cache them, access with some duplicate elimination. but i believe this is unnecessary when the shell properly allows picking a completion

tuurep:
Thanks, using `prefix-only-completions = true` feels great as a solution to my issue and that description on the docs is very nice

> zsh-autocomplete seemingly allows configuring this

This got me also looking at some options for the `:complete:` widget at `man zshcompsys` https://man.archlinux.org/man/zshall.1.en#insert~7 but finding it quite hard to understand.

tuurep:
Similarly to command abbreviations (https://github.com/papis/papis/pull/1112#issuecomment-3706615456), I notice completions don't work for a custom command:

https://papis.readthedocs.io/en/latest/scripting.html

That I find to be a bigger problem at the moment

papis/cli.py
```diff
@@ -75,12 +76,44 @@ def bool_flag(*args: Any, **kwargs: Any) -> DecoratorCallable:
...
+
+    comps_and_helps = (
+        (format(fmt, doc), format(help_fmt, doc)) for doc in
+        handle_doc_folder_or_query(
+            # return all documents on empty query
+            incomplete if incomplete
+            else papis.config.getstring("default-query-string"),
+            None
+        )
+    )
```
tuurep, line 115: Here's a way to make it match only documents in the given library when used with `papis -l somelibrary open <Tab>`
```suggestion

    lib = ctx.parent.params.get("lib")

    from papis.database import get_database
    try:
        db = get_database(lib)
    except RuntimeError:
        # nonexistent library name, return empty completions without noise
        return []

    query = (
        # return all documents on empty query
        incomplete if incomplete
        else papis.config.getstring("default-query-string")
    )

    comps_and_helps = (
        (format(fmt, doc), format(help_fmt, doc)) for doc in db.query(query)
    )

```

Any caveats with this?

tuurep: Needs error handling if given a library name that doesn't exist: spits a lengthy error on completion

**Edit:** updated suggestion with that

tuurep: Not well thought out enough using an unnamed library (giving an existing dir as `-l`):

image

(note: `pap` is my alias for `papis`, trying to avoid showing it in demos but it's slippin')

doc/source/default_settings.rst
```diff
@@ -137,6 +137,21 @@ General settings
...
+
+.. papis-config:: completion-help-format
+
+    Format for the tip displayed next to document completions in the command line
+    upon pressing Tab. Not all shells support this.
+
+.. papis-config:: prefix-only-completions
+    :type: bool
+
+    Whether to only suggest completions that start with the inserted query. Setting this to true will only suggest e.g. `einstein1905` when the query is `ein` but neither `wittgenstein1921` nor `goethe1780` with the title *Ein Gleiches*. This can be useful if your shell trims the query to the longest common prefix, like zsh.
```
alexfikl: Are we sure this is a good idea? I understand what it's trying to work around, but it feels like that's the shell's job and we should just provide it with a list of possible completions.

My main worry is that this makes the papis completions work differently than usual completions.

papis/cli.py
alexfikl: I think you mentioned this somewhere, but having played with it a bit more, maybe we want to do a
```
papis.config.set("match-format", papis.config.getformatpattern("completion-format"))
```
here? It's unintuitive to me to have things match that aren't in the `completion-format`.

doc/source/default_settings.rst
retroflexivity: > Are we sure this is a good idea? I understand what it's trying to work around, but it feels like that's the shell's job and we should just provide it with a list of possible completions.

well, since it's a *workaround* for certain idiosyncrasies of a shell and off by default, I don't feel bad about it

> My main worry is that this makes the papis completions work differently than usual completions.
not quite sure about this. e.g. check an [example](https://click.palletsprojects.com/en/stable/shell-completion/#custom-type-completion) from the click docs

```
class EnvVarType(ParamType):
    name = "envvar"

    def shell_complete(self, ctx, param, incomplete):
        return [
            CompletionItem(name)
            for name in os.environ if name.startswith(incomplete)
        ]

@click.command()
@click.option("--ev", type=EnvVarType())
def cli(ev):
    click.echo(os.environ[ev])
```

it uses `startswith`, too

papis/cli.py
retroflexivity: `prefix-only-completions` does this (should we split it into two options?). otherwise, cf. e.g. github that matches on commit titles, not just hashes 

image

doc/source/default_settings.rst
alexfikl: I guess, but this can be controlled from the shell hopefully. E.g. in `zsh` you can do
```sh
zstyle ':completion:*' matcher-list 'm:{a-z}={A-Z}'
```
to remove any fancy substring matching and just do a case insensitive prefix match. For what it's worth, my current default is
```sh
> zstyle -L ':completion:*' matcher-list
zstyle ':completion:*' matcher-list 'm:{[:lower:]}={[:upper:]}' 'm:{[:upper:]}={[:lower:]}' 'r:|[._-]=* r:|=*' 'l:|=* r:|=*'
```
which does substring matching (with `r:` and `l:`) if the prefix matching fails.

papis/cli.py
alexfikl: I meant that `handle_doc_folder_or_query` should just use the `completion-format` and not `match-format` in this function. We can force it to do that by using `papis.config.set`, which should be orthogonal to what `prefix-only-completions` does.

alexfikl: Also not quite sure how this works with the `whoosh` backend. That ignores `match-format` completely anyway.

retroflexivity: > I meant that `handle_doc_folder_or_query` should just use the `completion-format` and not `match-format` in this function. We can force it to do that by using `papis.config.set`, which should be orthogonal to what `prefix-only-completions` does.

i was saying that i don't think this is the optimal behavior. i would like to get suggestions based on title etc, not just the ref (similarly to what github does)

alexfikl: Ah, I see, so the desired behavior is:
1. Get a query string (i.e. `incomplete`).
2. Match it against the whole `match-format`.
3. Only show the shell the `completion-format` (which is in no way guaranteed to contain any of the substrings in `match-format`) text for further matching?

I gladly admit I have very little experience in how fancy completions actually work, haha, but that seems confusing to me.

doc/source/default_settings.rst
retroflexivity: well, I'm not using it either so I'll go with any decision. @tuurep seems to enjoy it, though

jghauser: I also run into @tuurep's problem regarding clearing of completion in zsh (it's super common, happening almost every time). I think having some easy way of fixing this at least with a setting would be useful. Even better would of course be to not cause such problems with common options in common shells (e.g. bash, zsh, fish). The clearing issue definitely makes Papis' completions work differently from usual completions :sweat_smile:.

I'm not quite such I understand @alexfikl's suggestion. Should these zsh completion settings somehow avoid the problem? It didn't seem to do anything for me.

alexfikl: > I'm not quite such I understand @alexfikl's suggestion. Should these zsh completion settings somehow avoid the problem? It didn't seem to do anything for me.

It didn't? :facepalm: The internet must have lied to me.. I'll actually try it out and report back if there's anything to it :\ Might need a [`compinit`](https://zsh.sourceforge.io/Doc/Release/Completion-System.html#Completion-System-Configuration) or something to regenerate the completions.

That being said.. I wouldn't be eager to modifying that, since it might mess with other completion muscle memory :(

jghauser: I'm not sure if I'm doing something wrong but I get the same issues whether I have:

`zstyle ':completion:*'                 matcher-list 'm:{a-z}={A-Z}'` 

(my default) or your setting. I also had

`zstyle ':completion:*:correct:*'       insert-unambiguous true`

which I thought might cause the problem, but setting that to `false` didn't fix things :thinking:. I ran `autoload -U compinit && compinit` in between trials.

alexfikl: Well that's unfortunate.. I was hoping something like that would work :\

retroflexivity:
> Similarly to command abbreviations ([#1112 (comment)](https://github.com/papis/papis/pull/1112#issuecomment-3706615456)), I notice completions don't work for a custom command:
> 
> https://papis.readthedocs.io/en/latest/scripting.html
> 
> That I find to be a bigger problem at the moment

@tuurep could you please send an example of a script you'd like to support completions?

tuurep:
Yeah!

This is the main one I use (so far)

https://github.com/tuurep/dotfiles/blob/master/.config/papis/scripts/papis-note

`papis note`: essentially does both `papis open` and `papis edit -n` on a document

I was wondering if it's something I'm supposed to put on the script but didn't investigate yet

alexfikl:
@tuurep I don't think our ad-hoc "command abbreviations" have a chance of working with this right now because those are just errors as far as `click` is concerned (that get automatically expanded to the real thing). If we introduce actual aliasing, we can let click know about those and it should be able to include them in the completions.

No idea why the custom commands don't work though. :grin:

retroflexivity:
> No idea why the custom commands don't work though. 😁

well, `external_cli` doesn't have the `query_argument` decorator, and the script file is only evaluated on execution, so its click options are ignored for completion. adding `@papis.cli.query_argument()` to `external_cli` (before `@click.argument("flags", nargs=-1)`) does the job.

(i'm not sure whether we want to enable doc completions for external scripts by default, but anything else would require significant rethinking of external execution, and also there are no completions at all otherwise so this will do no harm i guess)

alexfikl:
> adding @papis.cli.query_argument() to external_cli (before @click.argument("flags", nargs=-1)) does the job.

If this is the case, we're effectively requiring all external scripts to take "query" as the first positional argument. Is that reasonable? There aren't many people making external scripts, so maybe it's not a big deal, but even the included commands don't all have a query argument.

> (i'm not sure whether we want to enable doc completions for external scripts by default, but anything else would require significant rethinking of external execution, and also there are no completions at all otherwise so this will do no harm i guess)

This seems very ambitious indeed, haha.

How does something like git handle this? I just have [git-bug](https://github.com/git-bug/git-bug) installed at the moment and that seems to ship its own completions. Those only seem to work (unless I'm doing something wrong) when calling `git-bug` but not for `git bug`.

EDIT: I was doing something wrong :grin: But git has some fancy manual completion.. I don't know if we can do that automatically through `click`.

papis/cli.py
tuurep: Resolved in [e03be83](https://github.com/papis/papis/pull/1112/commits/e03be83bc25c4b840fdb17ee7294387a43a54041), [5a2de20](https://github.com/papis/papis/pull/1112/commits/5a2de2046f14b4709606167217891349ad0178ae)

Tested also with a legitimate unnamed library (containing Papis docs), it works. Thanks!

retroflexivity:
> If this is the case, we're effectively requiring all external scripts to take "query" as the first positional argument.

well yes, but it's no big deal since we compose the arguments and just send them to the script together. it's up to the script how to parse them, so it need not expect a query. the following works with the only nuance being out-of-place completions

```python
#!/usr/bin/env python3

import papis.api
import papis.cli
import click

@click.command()
@click.argument("string")
def main(string: str) -> None:
    print(string)

if __name__ == "__main__":
	main()
```

alexfikl:
> the following works with the only nuance being out-of-place completions

Assuming the script takes a positional argument at all.

I don't know, this seems a bit hacky. I would prefer the git way of doing it: the script defines its flags and arguments so it can also define its completions. However, then we would need to add some code to our actual completion files to work with external completions. Not sure if that's even possible with the auto-generated stuff from `click`..

Anyway, we can keep that, but as soon as someone comes and says their script broke, I'm deleting it :grin:

alexfikl:
Is there anything else that needs to be done here?

My main worry now is that it's quite finicky. Any way we could add some tests? Not necessarily for the shell completion itself, but at least for what `_query_shell_complete` returns with various parameters.

tuurep:
I'm currently satisfied! As long as I set the `prefix-only-completions = true`, if we accept that's a good workaround for whatever the actual problem was there

retroflexivity:
> Assuming the script takes a positional argument at all.

oh right.

well, bash scripts won't care, and python ones can be fixed by adding a sink... need to update the docs as well.

I agree that this is all quite ugly and we could do better, but I'm not enthusiastic enough

jghauser:
I'm happy with how this works, though given how common the zsh issues are without `prefix-only-completions = true`, I would like this workaround to be documented in [doc/source/shell_completion.rst](https://github.com/papis/papis/blob/6107ef53c4976c8e694c7ac7d796ea235b95c6f5/doc/source/shell_completion.rst).

Thanks a lot for all the work -- it looked tedious :sweat_smile:!

tuurep:
A bit of a weirdness I see with these settings however:

```
completion-format = {{ doc["title"] }}
completion-help-format = {{ doc["ref"] }}
prefix-only-completions = true
```

image

First doc:

- title: `Modern C`
- ref: `gustedt2024`

Second doc:

- title: `Modern Vim: Craft Your Development Environment with Vim 8 and Neovim`
- ref: `neil2018`

Completion goes partially into the help part, the `:` in the title looks related

**Edit:** Better view of the damage of the colon:

image

**Edit:** Maybe I have some control over it in the jinja template

With an empty `completion-help-format`, I think it's sane:

image

(they're on one line only because `Modern C` is such a short title)

retroflexivity:
seems to be a frequent problem in bash and the family, see E13 in http://tiswww.case.edu/php/chet/bash/FAQ. added cleaning of the problematic characters

tuurep:
Great, thanks:

image

I wouldn't have quite known where to look.

doc/source/shell_completion.rst
```diff
@@ -54,3 +54,14 @@ Alternatively, the completion can be generated on-the-fly by running (see more i
...
+
+====================
+Document completions
+====================
+
+Commands that expect a document query (``edit``, ``addto``, etc.) support document completions.
+
+Normally, all documents that match the incomplete query via ``match-pattern`` are returned, for the shell to filter them. if the shell expects completions already filtered to those that have the incomplete string as a prefix, ``prefix-only-completions`` can be used.
+
+For example, ``zsh`` replaces the incomplete string with the longest common prefix of the suggested completions, which is not guaranteed to be longer or equal then the incomplete string. to prevent losing the incomplete string, enable ``prefix-only-completions``.
```
tuurep:
```suggestion
Document completions
--------------------

Commands that expect a document query (``edit``, ``addto``, etc.) support
document completions.

Normally, all documents that match the incomplete query via
:confval:`match-format` are returned, for the shell to filter them. If the shell
expects completions already filtered to those that have the incomplete string as
a prefix, :confval:`prefix-only-completions` can be used.

For example, ``zsh`` replaces the incomplete string with the longest common
prefix of the suggested completions, which is not guaranteed to be longer or
equal then the incomplete string. To prevent losing the incomplete string,
enable :confval:`prefix-only-completions`.
```

tuurep (commented):
Some doc improvements:

- h2
- few typo fixes
- link to config options referenced
- format long lines to textwidth=~~88~~ [80](https://github.com/papis/papis/pull/1112#discussion_r2662900350)

doc/source/shell_completion.rst
tuurep: Question: does the link to `match-format` make sense?

(just making sure specifically `match-pattern` wasn't important)

tuurep: Edit about textwidth: looks like 88 is for py and on docs it's.. about 80, but not always consistent

Edited with 80 which looks to be the most polite doc textwidth

doc/source/default_settings.rst
tuurep:
```suggestion
    Whether to only suggest completions that start with the inserted query.
    Setting this to true will only suggest e.g. `einstein1905` when the query is
    `ein` but neither `wittgenstein1921` nor `goethe1780` with the title *Ein
    Gleiches*. This can be useful if your shell trims the query to the longest
    common prefix, like zsh.
```

Nit: textwidth 80

doc/source/shell_completion.rst
retroflexivity: > Question: does the link to `match-format` make sense?
> 
> (just making sure specifically `match-pattern` wasn't important)

yeah I'm just dumb

jghauser:
Looks all good to me, thanks for adding the docs!

doc/source/shell_completion.rst
```diff
@@ -54,3 +54,20 @@ Alternatively, the completion can be generated on-the-fly by running (see more i
...
+document completions.
+
+Normally, all documents that match the incomplete query via
+:confval:`match-format` are returned, for the shell to filter them. If the shell
+expects completions already filtered to those that have the incomplete string as
+a prefix, :confval:`prefix-only-completions` can be used.
+
+For example, ``zsh`` replaces the incomplete string with the longest common
+prefix of the suggested completions, which is not guaranteed to be longer or
+equal then the incomplete string. To prevent losing the incomplete string,
```
tuurep:
```suggestion
prefix of the suggested completions, which is not guaranteed to be longer than
or equal to the incomplete string. To prevent losing the incomplete string,
```

Sorry.. I meant to include this grammar fix but it didn't end up in the final suggestion

This is the last one! 😄

tuurep:
Yes thanks so much everyone

See also https://github.com/papis/papis/issues/901 which could be relevant for some things discussed here

alexfikl:
I added a few more notes to the code and linked to the discussion here. Also rephrased the documentation a bit and mentioned that the picker could still be shown (in case someone expected the completion to just open the selected document).

Hopelly didn't introduce any silly typos :-s 

Thanks a lot for all the work on this! I'll go ahead and merge this when the CI passes :rocket:

PR #1113, opened 2025-12-20
Add ruff rules for flake8-simplify

cclauss:
https://docs.astral.sh/ruff/rules/#flake8-simplify-sim

% `ruff check --select=SIM --ignore=SIM102,SIM117 --statistics`
```
5	SIM201	[ ] negate-equal-op
4	SIM108	[ ] if-else-block-instead-of-if-exp
4	SIM300	[*] yoda-conditions
2	SIM105	[ ] suppressible-exception
1	SIM118	[ ] in-dict-keys
1	SIM905	[*] split-static-string
1	SIM910	[*] dict-get-with-none-default
Found 18 errors.
```


alexfikl:
Any reason in particular you want these in? In my experience the SIM rules are quite opinionated, so I wasn't very inclined to turn them on.

cclauss:
SIM rules are primarily designed to simplify readability when transitioning from one project to another.  If they do not correspond to this project's needs, then this PR can be closed.

examples/scripts/papis-abbrev.py
```diff
@@ -175,10 +175,7 @@ def run(query: str, all_: bool = True) -> int:
         search=query
     )
 
-    if all_:
-        picked_documents = documents
-    else:
-        picked_documents = list(papis.api.pick_doc(documents))
+    picked_documents = documents if all_ else list(papis.api.pick_doc(documents))
```
alexfikl, line 178: Can you disable `SIM108`? This might be personal preference, but I've never liked putting the condition in the middle (likely because of all the normal ifs).

examples/scripts/papis-check.py
```diff
@@ -92,9 +92,8 @@ def run(keys: Sequence[str],
                 result.append({"doc": document, "key": key, "msg": "Missing key"})
             elif not document[key] and document[key] is not False:
                 result.append({"doc": document, "key": key, "msg": "Empty key"})
-            elif key == "files":
-                if not check_files(document):
-                    result.append({"doc": document, "key": key, "msg": "Missing files"})
```
alexfikl, line 97: Can you disable `SIM102`? This was most written as a "switch on each [KEY] type and do [THING]", so combing the ifs is a bit confusing.

examples/scripts/papis-check.py
```diff
@@ -122,7 +121,7 @@ def cli(query: str, keys: list[str]) -> None:
             )
         )
 
-    if not len(troubled_docs) == 0:
+    if len(troubled_docs) != 0:
```
alexfikl, line 124: Isn't the pythonic way to do with `if not troubled_docs`?

papis/web/citations.py
```diff
@@ -26,17 +26,16 @@ def render(doc: Document,
...
+                method="POST"), t.div(cls="btn-group", role="group"):
+        with t.button(cls="btn btn-success",
+                      type="submit"):
+            wh.icon_span("cloud-bolt", "Fetch citations")
+        if "doi" in doc:
+            quoted_doi = urllib.parse.quote(doc["doi"], safe="")
+            with t.a(cls="btn btn-primary",
+                     target="_blank",
+                     href=ads_fmt.format(doi=quoted_doi)):
+                wh.icon_span("globe", "ads")
```
alexfikl, line 38: I think the reasoning here (to abuse the nesting) was to match the HTML tags better. I'm not sure it's an improvement to group them?

alexfikl:
@cclauss Are you still interested in adding these linting fixes?

examples/scripts/papis-check.py
cclauss: There is a different ruff rule for that.

alexfikl: Ah, yes, I know. I was trying to say that we have that rule (PLC1802) enabled already, so this should fix it. It'll probably show up once we run the CI though.

alexfikl:
I'll go ahead and close this. Feel free to re-open if interested in addressing the points I mentioned.

PR #1121, opened 2026-01-03
doctor: Add pointer to BibTeX manual

hseg:
The Reason given for biblatex-key-convert is unclear unless one has previously read the discussion on the topic in the manual.
The location of the discussion appears in the code -- show it to the user for clarity.
Am unsure of the best shape of the message, in particular:
- whether the `(see manual)` should appear at the end of the line or before the colon
- if we shouldn't strengthen the language to `'issue' should be 'number'`

### DIFF
--- a/papis/commands/doctor.py
+++ b/papis/commands/doctor.py
@@ -683,7 +683,8 @@ def biblatex_key_convert_check(doc: Document) -> list[Error]:
         fix_action = None
         if key == "issue":
             if is_number_like(value):
-                msg = f"Document key 'issue' looks like a 'number': '{value}'"
+                msg = (f"Document key 'issue' looks like a 'number'"
+                       f" (see BibLaTeX manual §2.3.11): '{value}'")
                 fix_action = issue_to_number_fixer
 
         if fix_action is None:


alexfikl:
Hm, can't say I feel strongly about this, so whichever way sounds / looks more helpful is good enough for me. 

> if we shouldn't strengthen the language to 'issue' should be 'number'

I went with that loose text at the time because my reading of various forums was that this is quite subjective. Setting the issue to a number seems to not be recommended, but it's less clear where other types of values/codes go. Also, usage in the wild does not seem to care much, haha.

hseg:
> Hm, can't say I feel strongly about this, so whichever way sounds / looks more helpful is good enough for me.

I'd say then that if number values tend to be under ~6 chars, it makes sense to put the ref after (to get to the point). Otherwise, I'd put the static parts together.

> if we shouldn't strengthen the language to 'issue' should be 'number'
>
> I went with that loose text at the time because my reading of various forums was that this is quite subjective. Setting the issue to a number seems to not be recommended, but it's less clear where other types of values/codes go. Also, usage in the wild does not seem to care much, haha.

(-: my understanding of the manual was that these should be styled differently, so there is some weight to the difference. But if people you spoke to were ambivalent, I don't feel strongly about strengthening the language.

To avoid all the above concerns, we could also just put the language in the manpage.

alexfikl:
> (-: my understanding of the manual was that these should be styled differently, so there is some weight to the difference.

Yeah, the manual is pretty clear that they could/should be styled differently, agreed! My point was more that people do not seem generally aware there even is a difference, much less where to put what :grin:

That said, I'm fine with strengthening the language, since we are trying to say that the choice we made is preferred.

papis/commands/doctor.py
```diff
@@ -683,7 +683,8 @@ def is_number_like(value: Any) -> bool:
         fix_action = None
         if key == "issue":
             if is_number_like(value):
-                msg = f"Document key 'issue' looks like a 'number': '{value}'"
+                msg = (f"Document key 'issue' looks like a 'number'"
+                       f" (see BibTeX manual §2.3.11): '{value}'")
```
jghauser: This should be `BibLaTex manual`

alexfikl:
@hseg This looks good to me besides the little typo that @jghauser mentioned. Can you fix that? I can update it if not, no problem at all.

hseg:
Hrmph. The Github interface makes pushing amended commits uglier, repushing the initial commit and the amendment. Not rebasing for clarity's sake.

alexfikl:
Looks great! In it goes :rocket:

PR #1125, opened 2026-01-06
Use sphinxlint to enforce some documentation rules

alexfikl:
This adds a [sphinx-lint](https://github.com/sphinx-contrib/sphinx-lint) to the CI to check that documentation also stays at a max line length of 88 (and other small checks).

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -48,6 +48,35 @@
               };
             };
 
+            sphinx-lint = final.buildPythonPackage rec {
+              pname = "sphinx-lint";
+              version = "1.0.2";
+              pyproject = true;
+              build-system = [
+                final.hatchling
+                final.hatch-vcs
+              ];
+
+              dependencies = [
+                final.polib
+                final.regex
+              ];
+
+              src = final.fetchPypi {
+                inherit version;
+                pname = "sphinx_lint";
+                sha256 = "sha256-Tn/BL0T3ULAAbqrSN9fbmx2KupKt2pyDiviRZUs3HTU=";
+              };
+
+              doCheck = false;
+
+              meta = with pkgs.lib; {
+                homepage = "https://github.com/sphinx-contrib/sphinx-lint";
+                description = "Linter for Sphinx documentation";
+                license = licenses.psfl;
+              };
+            };
+
             papis =
               let
                 # Returns an attribute set that can be passed to `buildPythonPackage`.


alexfikl:
@jghauser Good idea to add some very basic linting to the docs?

Also, I've messed around with the `flake.nix` file to add `sphinx-lint`. It would be very appreciated if you could check that it's not doing anything weird :grin:

jghauser:
> @jghauser Good idea to add some very basic linting to the docs?

Yes, definitely good!
 
> Also, I've messed around with the `flake.nix` file to add `sphinx-lint`. It would be very appreciated if you could check that it's not doing anything weird 😁

Doesn't seem to be doing anything weird, but is there a reason you defined `sphinx-lint` yourself? It seems to be in `nixpkgs`. Ah, I see it's not in the python packages set and that makes the flake fail without your explicit package definition. I think the better solution would be to use an overlay to put it where it's expected. I'll look into it!

flake.nix
```diff
@@ -48,6 +48,35 @@
...
+
+              doCheck = false;
+
+              meta = with pkgs.lib; {
+                homepage = "https://github.com/sphinx-contrib/sphinx-lint";
+                description = "Linter for Sphinx documentation";
+                license = licenses.psfl;
+              };
+            };
+
```
jghauser:
```suggestion
            inherit (pkgs) sphinx-lint;
```

tuurep:
Hey regarding text width 88 vs. 80 from #1112,

The way I arrived at 80 was by setting a colorcolumn in nvim and eyeballing the doc files, where it was pretty clear they've at some point been formatted at 80 cols.

Is it going to be 88 for docs, anyways?

88 as I understand is a convention from the Black formatter https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length (just as context)

flake.nix
alexfikl: > Doesn't seem to be doing anything weird, but is there a reason you defined sphinx-lint yourself?

Because I don't know what I'm doing in nix :grin: Thank you so much!

alexfikl:
> Is it going to be 88 for docs, anyways?

I don't see any reason to have a different line length in the docs and the code. Is that a thing?

Plus, I think enforcing 80 line length is going to be very annoying with `sphinx-lint`. It has no local ignores and we have a lot of links in the docs that go over it (even 88 was a bit of a pain in a few places).

tuurep:
Yes, using the same textwidth for docs seems like a better idea :+1: 

Just thinking about what it currently is and what it's going to be changed into.

Issue #1096, opened 2025-11-11
Have `man papis` page contain something in it, hopefully to lead into config options


tuurep:
Hi I have a suggestion to make after noticing myself expect certain things:

1. I needed to check a certain papis setting
2. Instinct: run `man papis` and start searching
3. Friction: `No manual entry for papis`
    - What I wanted was instead `man papis-config-settings` but this is very hard to figure out

I understand that the manpages are auto-generated from our docs, and that generally works pretty well.

Would it make sense to make `man papis` lead directly to `papis-config-settings` manpage (or be a copy of it)?

tuurep:
I've also seen this convention of placing man pages about configs in section 5, like:

`dunst(5)` which is accessed by `man 5 dunst`

Some discussion about that: https://unix.stackexchange.com/a/736087

What do you think of `papis(5)` as a copy of `papis-config-settings`?

tuurep:
Oh yeah,

`papis-default` is the one for the main command, `papis`, in a sense.

Suggestion (if we have control over these things):

- `man papis` => `man papis-default`
- `man 5 papis` => `man papis-config-settings`

alexfikl:
You should be able to get some good results there with a bit of work. You can add text only to the man pages in the rst docs using
```rst
.. only:: man

   This text will appear **only** in the Sphinx man page output.
```
so that they link better to each other. You can also add them to more appropriate sections using https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-man_pages. They're currently all added to section 1 in here:
https://github.com/papis/papis/blob/cecfc5364d1bdee8f623a546b9278b928ac94a57/doc/source/conf.py#L280-L289
Section 7 should also be more appropriate for the library structure description and whatnot.

tuurep:
That sounds promising, thanks

I'll get to it at some point :)

PR #1126, opened 2026-01-08
Add `man papis` and other manpage improvements

tuurep:
Draft for #1096

### Motivation

- make the manpages discoverable, currently `papis(1)` doesn't exist, which easily makes it seem there are no manpages
- get to config options as easy as possible:
    - `man 5 papis` instead of `man papis-config-settings`
    
### Todos/notes    

I think it would be best to rework `papis-default(1)` into `papis(1)`

Could take example from `git(1)`

Or have essentially the `papis --help` message but in manpage form with additional linkages to other pages?

To be investigated:

https://github.com/papis/papis/issues/1096#issuecomment-3515334987

> Section 7 should also be more appropriate for the library structure description and whatnot.

If there are many pages that could go into section 7, should they be combined into one?

**Edit:** probably not, look at how git does the section 7 pages:

image

I think adding a `SEE ALSO` section at the bottom just like that would be great, with also the `papis(5)` (config page) mentioned there.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -283,8 +283,10 @@ latex_documents = [
 # One entry per manual page. List of tuples
 # (source start file, name, description, authors, manual section).
 man_pages = [
-    ("configuration", "papis-config-settings", "Papis configuration",
+    ("commands/default", "papis", "Papis",
      [papis.__author__], 1),
+    ("configuration", "papis", "Papis configuration",
+     [papis.__author__], 5),
     ("library_structure", "papis-library-structure", "Papis library structure",
      [papis.__author__], 1),
     ("info_file", "papis-info-file", "Papis info.yaml file",


alexfikl:
> Or have essentially the papis --help message but in manpage form with additional linkages to other pages?

There'a already a `Command-line Interface` section that's essentially just `papis --help` slightly reformatted. Is that not enough? Or?

> I think adding a `SEE ALSO` section at the bottom just like that would be great, with also the `papis(5)` (config page) mentioned there.

I'm definitely in favor of that! Using `.. only:: man` should make it rather painless.

> If there are many pages that could go into section 7, should they be combined into one?

It's also easier to just have separate files for each one :grin: I would prefer that + a nice `SEE ALSO` section (this would also nicely match the online docs, which are separate).

EDIT: Was looking at `man git` and that SYNOPSIS section looks pretty nifty. No idea how to get it though..

tuurep:
Oops sorry for wiping the message, I was about to move the todos around. I edited all of them back :smile:

tuurep:
It's like `papis-default` does have all the correct info in it, but in a bit of an awkward order for the manpage and this:

image

Looks so much worse than this:

image

Agree on the SYNOPSIS, could start with just substituting this for it:

```
papis [OPTIONS] COMMAND [ARGS]...
```

if getting it just like that is too hard to achieve

alexfikl:
> Looks so much worse than this:

Yeah, it kind of does :( I've never played with Sphinx manpage generation though, so I'm not sure how much of that can be styled differently. If you make any progress, that would be amazing! :grin:

EDIT: The styling might also come from `sphinx-click`.

tuurep:
Yeah it will take some experimentation

I was also thinking should we link to the subcommands' pages similar to this:

image

Which would only make sense for a manpage, so a large section would have to be written separately for docs/manpage :thinking:

jghauser:
Always happy to see work on the docs :heart:! More than happy to help with this!

> Which would only make sense for a manpage, so a large section would have to be written separately for docs/manpage 🤔

What do you mean by this? That we'd have to write a page detailing how to find all the subcommand specific pages? That doesn't seem so bad to me...

alexfikl:
> What do you mean by this? That we'd have to write a page detailing how to find all the subcommand specific pages? That doesn't seem so bad to me...

Yeah, agreed, that doesn't seem too bad.. it would just be a one time thing. Maybe we can even automate it somehow using a special directive + some magic in `papis/sphinx_ext.py`.

tuurep:
Yep I was just saying if we want the list of commands to double as refs to their manpages:

- `papis-add(1)`

Only makes sense in the manpage version

But that was just inspiration from the git manpage, those refs can be on a separate list also

But yeah it's just that one word/representation that has to vary that's true

But this is thinking a bit too far ahead :smile:

> That we'd have to write a page detailing how to find all the subcommand specific pages?

With this approach, they could just be generated into the `SEE ALSO` part or have like one `SEE ALSO` for all subcommands pages, and another for the misc pages

Or a general note saying "The subcommands man pages are in `papis-<command>(1)`"

PR #1134, opened 2026-01-11
Update regex for string-cleaner

### LINKED ISSUE #1123: doctor: string-cleaner(author) regex doesn't match multi-initials
* papis version (`$ papis --version` or commit number): 79dd5d7e

# how to reproduce the issue
https://sociologica.unibo.it/article/view/8350 has as author `Schmidt, Johannes F.K.`, which the regex considers to have initials without a following period.
Inserting a space between the initials silences this, but clearly the correct solution is fixing the regex.

alexfikl:
Fixes #1123.


alexfikl:
@hseg I'll go ahead and merge this. Feel free to re-open if there are some cases it doesn't handle well enough.

PR #1137, opened 2026-01-23
Add upper bound for Sphinx on the CI

alexfikl:
The new Sphinx 9.1.0 started using Python 3.12+ only syntax. Luckily, this seems to only be in `TYPE_CHECKING` blocks, so it only breaks `mypy`.

For now, this modifies the `ci-install.sh` script to keep it to `<9.1.0`, so that `mypy` works properly on all versions.

### DIFF
--- a/tools/ci-install.sh
+++ b/tools/ci-install.sh
@@ -2,7 +2,10 @@
 
 set -o errexit -o noglob -o pipefail
 
-# NOTE: setuptools is needed as a runtime dependency of isbnlib
-python -m pip install --upgrade setuptools
 python -m pip install --upgrade pip hatchling wheel
 python -m pip install --editable '.[develop,docs,optional]'
+
+# NOTE: setuptools is needed as a runtime dependency of isbnlib
+python -m pip install --upgrade setuptools
+# NOTE: sphinx>=9.1.0 uses 3.12 syntax for typing and breaks mypy on the CI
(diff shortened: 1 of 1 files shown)


alexfikl:
cc @hseg This will probably also break the `makepkg` check once Arch upgrades to the latest Sphinx version. I'm not sure if there is any better fix..

hseg:
Will it? Arch only supports python 3.14 anyway, so shouldn't this pass?

23.01.2026 19:59:26 Alex Fikl ***@***.***>:

alexfikl:
> Will it? Arch only supports python 3.14 anyway, so shouldn't this pass?

@hseg The issue is that we're setting the Python version for `mypy` to be 3.10 in `pyproject.toml`, i.e. it's not checking using the installed version. You might be able to work around it by forcing it using `--python-version '3.14'` in the `PKGBUILD`.

hseg:
Understood, testing and updating. Thanks for the heads'-up!

hseg:
Hrm. Not sure why, but doing this breaks the mypy check in a clean chroot (it can't infer the types for python-doi), but on my installed system it works fine.

hseg:
And my mypy config is _more_ strict than the default

hseg:
Nm, problem solved - for some reason was installing a python 3.13 python-doi to the chroot

PR #1147, opened 2026-03-01
Make isbnlib an optional dependency

alexfikl:
This officially makes `isbnlib` an optional dependency, since `pkg_resources` was officially removed in the latest `setuptools`, so using `isbnlib` should downright fail when using it.


alexfikl:
For what it's worth, our `isbnlib` tests don't get to call into `pkg_resources`, so they actually still work (AFAICT). Even so, this adds a bound on `setuptools` since the runtime will not work.

PR #1163, opened 2026-04-11
Rename doctor-key-type-keys to document-field-types

alexfikl:
This moves `doctor-key-type-keys` to a more central location in `papis.document`. The config option is being used in more places to check things have expected types, so it deserves to be promoted.

cc @jghauser

### DIFF
--- a/doc/source/default_settings.rst
+++ b/doc/source/default_settings.rst
@@ -236,8 +236,8 @@ General settings
 
 .. papis-config:: document-field-types
 
-   A list of strings ``key:type`` items defining expected types for document
-   keys. The type should be a builtin Python type. For example, this can be
+   A list of ``key:type`` strings defining expected types for document
+   fields. The type should be a builtin Python type. For example, this can be
    ``["year:int", "tags:list"]`` to check that the year is an integer and the
    tags are given as a list in a document. These values are used in the ``papis
    doctor`` check ``key-type``, in the ``papis update`` command, etc.
@@ -249,8 +249,8 @@ General settings
 .. papis-config:: document-field-types-extend
     :type: :class:`~typing.List` [:class:`str`]
 
-    A list of keys that extend the default ones from :confval:`document-field-types`.
-    This list extends instead of overwriting the given keys.
+    A list of ``key:type`` strings that extend the default ones from :confval:`document-field-types`.
+    By setting these, you can extend the defaults rather than overwriting them.
 
     **Note**: This configuration option was previously available as
     ``doctor-key-type-keys-extend``. This version is deprecated (starting with
--- a/doc/source/default_settings.rst
+++ b/doc/source/default_settings.rst
@@ -240,7 +240,7 @@ General settings
    fields. The type should be a builtin Python type. For example, this can be
    ``["year:int", "tags:list"]`` to check that the year is an integer and the
    tags are given as a list in a document. These values are used in the ``papis
-   doctor`` check ``key-type``, in the ``papis update`` command, etc.
+   doctor`` check ``field-type``, in the ``papis update`` command, etc.
 
    **Note**: This configuration option was previously available as
    ``doctor-key-type-keys`` and ``doctor-key-type-check-keys``. Both of
@@ -692,10 +692,10 @@ Doctor options
     :confval:`doctor-html-tags-keys`. This list extends instead of overwriting
     the given keys.
 
-.. papis-config:: doctor-key-type-separator
+.. papis-config:: doctor-field-type-separator
     :type: str
 
-    A separator used by the ``key-type`` check fixer. When converting from
+    A separator used by the ``field-type`` check fixer. When converting from
     :class:`str` to :class:`list`, it is used to split the string into a list,
     and when converting from :class:`list` to :class:`str`, it is used to join
     list items. The split will ignore additional whitespace around the separator
--- a/doc/source/default_settings.rst
+++ b/doc/source/default_settings.rst
@@ -184,7 +184,7 @@ General settings
     based on the Python formatter. These will need to all be specified explicitly
     if another formatter is chosen.
 
-    **Note** The older (misspelled) version ``"formater"`` is deprecated.
+    **Note**: The older (misspelled) version ``"formater"`` is deprecated.
 
 .. papis-config:: doc-paths-lowercase
     :type: bool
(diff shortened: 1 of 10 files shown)


doc/source/default_settings.rst
```diff
@@ -249,6 +249,28 @@ General settings
    and are just used to add style to the resulting citation (e.g. marking titles
    as italic).
 
+.. papis-config:: default-key-types
```
alexfikl: Naming?

papis/document.py
```diff
@@ -292,6 +292,64 @@ def split_authors_name(authors: str | list[str],
...
+    # parse configuration options
+    import builtins
+
+    result: dict[str, type] = {}
+    for value in keys:
+        if ":" not in value:
+            logger.error("Invalid (key, type) pair: '%s'. Must be 'key:type'.",
+                         value)
+            continue
+        key, cls_name = value.split(":")
```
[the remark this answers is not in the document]
alexfikl: Good point, fixed!

doc/source/default_settings.rst
jghauser: maybe `document-schema`? it's a bit more descriptive of what it is

alexfikl: I guess. One worry I have is that we don't actually impose the types anywhere and `document-schema` sounds very "this is what your document will be from now on".

I'm not sure what a good name would be.. Will give it a day or two..

jghauser: maybe `document-field-types`?

alexfikl: That sounds good to me. It's renamed now :grin: Let me know if anything else looks fishy.

doc/source/default_settings.rst
```diff
@@ -234,6 +234,28 @@ General settings
     ``papis --pick-lib export --all``. The format takes a dictionary named
     ``library`` with the keys *name*, *dir*, and *paths*.
 
+.. papis-config:: document-field-types
+
+   A list of strings ``key:type`` items defining expected types for document
+   keys. The type should be a builtin Python type. For example, this can be
```
jghauser:
```suggestion
   fields. The type should be a builtin Python type. For example, this can be
```

doc/source/default_settings.rst
```diff
@@ -234,6 +234,28 @@ General settings
...
+   doctor`` check ``key-type``, in the ``papis update`` command, etc.
+
+   **Note**: This configuration option was previously available as
+   ``doctor-key-type-keys`` and ``doctor-key-type-check-keys``. Both of
+   these are deprecated (starting with Papis 0.16) and will be removed.
+
+.. papis-config:: document-field-types-extend
+    :type: :class:`~typing.List` [:class:`str`]
+
+    A list of keys that extend the default ones from :confval:`document-field-types`.
```
jghauser:
```suggestion
    A list of ``key:type`` strings that extend the default ones from :confval:`document-field-types`.
```

doc/source/default_settings.rst
```diff
@@ -234,6 +234,28 @@ General settings
     ``papis --pick-lib export --all``. The format takes a dictionary named
     ``library`` with the keys *name*, *dir*, and *paths*.
 
+.. papis-config:: document-field-types
+
+   A list of strings ``key:type`` items defining expected types for document
```
jghauser:
```suggestion
   A list of ``key:type`` strings defining expected types for document
```

doc/source/default_settings.rst
```diff
@@ -234,6 +234,28 @@ General settings
...
+
+   **Note**: This configuration option was previously available as
+   ``doctor-key-type-keys`` and ``doctor-key-type-check-keys``. Both of
+   these are deprecated (starting with Papis 0.16) and will be removed.
+
+.. papis-config:: document-field-types-extend
+    :type: :class:`~typing.List` [:class:`str`]
+
+    A list of keys that extend the default ones from :confval:`document-field-types`.
+    This list extends instead of overwriting the given keys.
```
jghauser:
```suggestion
    By setting these, you can extend the defaults rather than overwriting them.
```

jghauser (commented):
Left some small nitpicks in the docs, but otherwise looks good! The only thing a bit confusing is that the `papis doctor` check is now not called in a consistent way (the option is about `field`s, but the check about `key`s). I think `field` is actually accurate here and will allow us to eventually slow move Papis towards some consistency even if for now it's leading to the opposite of that 😅. So yeah, maybe eventually we can move the `papis doctor` check name as well. Not a big issue though...

doc/source/default_settings.rst
jghauser: Here and below I'm also happy with 'string items' if you prefer that :)

alexfikl:
> The only thing a bit confusing is that the `papis doctor` check is now not called in a consistent way (the option is about `field`s, but the check about `key`s)

@jghauser That's actually a good point. We pretty consistently use `key` when referring to the various entries in a document (probably because the document is just a dictionary and those are called keys in Python). I'm not sure we have a good reason to switch to `field` then?

From a quick grep, `field` is used 
* In the `--sort-field` command line option, for some reason.
* When referring to some other software / data format, e.g. BibTeX "fields", or in database parts.
* Probably a few other places that snuck in because we didn't really pay attention :grin:

I would prefer `document-key-types` or even `default-document-key-types`. I initially added the `default` because of the usage in `papis update`, where it's used to initialize some values for e.g. `--append`. I'm thinking using it in `Document.__missing__` would also make sense (with the same meaning) to not always return an empty string.

jghauser:
> @jghauser That's actually a good point. We pretty consistently use `key` when referring to the various entries in a document (probably because the document is just a dictionary and those are called keys in Python). I'm not sure we have a good reason to switch to `field` then?

I thought that technically, the key is just the key of a key-value pair/field, and what we're setting here is the type that the value that belongs to the given key has to have. But I'm not even sure whether this is just the terminology in my head or if that's what these terms consistently mean -- and also it's a bit pedantic haha. So, I'm happy with anything that will get us to a consistent place somewhat soon :).

> I would prefer `document-key-types` or even `default-document-key-types`. I initially added the `default` because of the usage in `papis update`, where it's used to initialize some values for e.g. `--append`. I'm thinking using it in `Document.__missing__` would also make sense (with the same meaning) to not always return an empty string.

I don't think `default` really adds that much to understanding the option. It's true we don't always check the type and users can pretty much do whatever (if they don't use doctor, which really they should :smile:), but I don't think adding `default` really makes that clear (and makes the option pretty long too).

alexfikl:
> I thought that technically, the key is just the key of a key-value pair/field, and what we're setting here is the type that the value that belongs to the given key has to have

Ah :man_facepalming: I completely misunderstood.. in my head the key kind of was the value, haha. I'm not sure what's the more common terminology here :\ But I'm fine with keeping it `field` then!

jghauser:
Haha no worries! Ok, let's keep field then and make other things consistent as we come across them

alexfikl:
@jghauser Went ahead and also renamed the doctor check from `key-type` to `field-type` (+ everything that was using it), so at least that part should be pretty consistent now.

This is good to go now from my end, so let me know if there's anything else that I missed :grin:

jghauser:
looks good!

PR #1165, opened 2026-04-12
Port from pyparsing to lark for query parsing

### LINKED ISSUE #907: NOT in query doest not work on fields
* papis version (`$ papis --version` or commit number):

# how to reproduce the issue
I have a paper  with 'ellenbogen' in the author :
```
❯ papis --log INFO list "author:ellenbogen"
/home/alex/papers/4Y3UKB9D
```
Trying to filter for author without this author does not show any result
```
❯ papis --log INFO list "author:(NOT ellenbogen)"
[WARNING] commands.list: No documents retrieved

❯ papis --log INFO list "NOT author:ellenbogen"
[WARNING] commands.list: No documents retrieved

```
Same for tags. Am I missing something ? Either way, it would be cool to have the syntax in the quickstart at it's a bit buried in the doc.
Thanks :)

### LINKED ISSUE #1002: Does papis support languages other than english?
I have some references in Chinese and Japnaese. I found that I could query with these languages.

### LINKED ISSUE #1004: Non-ascii characters (ä, ö, ...) in `ref` keys problems
From https://github.com/papis/papis/issues/1002#issuecomment-2817809874, where I thought this was the same issue as something else:

>> Hey, I remember having a similar problem querying a string with the >> Finnish `ä` and `ö` characters.
>>
>> Looked at it again, `title` can be queried with `ä` and `ö`, but not >> `ref` (i.e. the citation key).
>>
>> Maybe it's something about UTF-8 support?
> 
> This sounds unrelated. Might be because we're generally "slugifying" references (since standard BibTeX doesn't support Unicode to my knowledge?). Can you open a separate issue with some example that doesn't work? (i.e. an `info.yml` file or something)

Here's an example with the primary author name *Käkönen*:

`$ papis add --from doi 10.1002/cncr.11132`

`info.yaml` produced with the enormous `citations` field removed:

```yaml
author: K�k�nen, Sanna-Maria and Mundy, Gregory R.
author_list:
- affiliation: []
  family: K�k�nen
  given: Sanna-Maria
- affiliation: []
  family: Mundy
  given: Gregory R.
doc_url: https://onlinelibrary.wiley.com/doi/full/10.1002/cncr.11132
doi: 10.1002/cncr.11132
files: []
issue: S3
journal: Cancer
language: en
month: 2
pages: 834--839
papis_id: c9266fc224159175277144cc36eb9111
publisher: Wiley
ref: kknen2003
time-added: 2025-04-21-20:33:00
title: Mechanisms of osteolytic bone metastases in breast carcinoma
type: article
url: https://doi.org/10.1002/cncr.11132
volume: '97'
year: 2003
```

Notice it has those � characters in the author fields, this is something I saw for the first time with this example.

My `ref` is generated with this jinja2 template:

```
ref-format = {%% if doc["author_list"] %%}{{ doc["author_list"][0]["family"].lower() }}{%% else %%}{{ doc["publisher"].lower() }}{%% endif %%}{{ doc["year"] }}
```

(TL;DR: `<1st author family><year>`)

Fixing the �s by hand we get this "correct" info.yaml:

```yaml
author: Käkönen, Sanna-Maria and Mundy, Gregory R.
author_list:
- affiliation: []
  family: Käkönen
  given: Sanna-Maria
- affiliation: []
  family: Mundy
  given: Gregory R.
doc_url: https://onlinelibrary.wiley.com/doi/full/10.1002/cncr.11132
doi: 10.1002/cncr.11132
files: []
issue: S3
journal: Cancer
language: en
month: 2
pages: 834--839
papis_id: c9266fc224159175277144cc36eb9111
publisher: Wiley
ref: käkönen2003
time-added: 2025-04-21-20:33:00
title: Mechanisms of osteolytic bone metastases in breast carcinoma
type: article
url: https://doi.org/10.1002/cncr.11132
volume: '97'
year: 2003
```

And this is the primary issue, document can't be queried with

`$ papis edit käkönen2003`

But this works:

`$ papis edit nen2003`

I made sure to check that I'm able to use an actual bib citekey like `käkönen2003` in a tex document, since I wasn't sure about this:

> (since standard BibTeX doesn't support Unicode to my knowledge?)

Maybe specific to my setup.

### LINKED ISSUE #1128: doc/feat: Query syntax undocumented, restrictive
* papis version (`$ papis --version` or commit number): 79dd5d7e

# how to reproduce the issue
Firstly, the builtin query syntax is underdocumented. Refering to papis/docmatcher.py, it appears to consist of repeated `[KEY :] VALUE` strings, with the `VALUE` permitted to be quoted, an arbitrary amount of space around the colons, and `KEY` and `VALUE` otherwise matching
```
KEY ~ /[a-zA-Z0-9-._/]+/
VALUE ~ /[a-zA-Z0-9-._/()]+/
```
Note, though, that this is much more restrictive than [YAML
allows](https://yaml.org/spec/1.2.2/#rule-ns-plain-one-line) (in
particular, I can't search for `VALUE` by sigils (`!@#$%`) which I could be
using for internal markers), and furthermore prevents searching for `KEY:` to
list documents that have any value for `KEY:`.
It's probably simpler to just say that the query string is just split into quoted substrings, and then chunked left-to-right into `KEY : VALUE` pairs, with both `KEY` and `VALUE` being optional.

alexfikl:
This improves the query parsing for the default `papis` database backend quite a bit:
1. Lark is supposedly a bit faster (need to test this).
2. Lark automatically has unicode support.
3. Adds support for `AND`, `OR` and `NOT` as keywords for the queries.

TODO:
* [x] Improve documentation.
* [x] Add to changelog.
* [x] Test a bit more on an actual library.

Superseeds #602.
Fixes #1128.
Fixes #1004.
Fixes #1002.
Fixes #907.


alexfikl:
@supersambo I went ahead and finished this up. It should do everything in #602 and hopefully leaves room for future extensions, if necessary. 

Any additional testing or review would be very welcome, whenever you have the time! :grin:

alexfikl:
For the performance, I'm not seeing a big change. For my library of about 2800 documents and a query that returns about 400, I'm getting something like the results below. However, `lark` is also doing unicode, so this is definitely worth it as far as I'm concerned. :grin:

#### Before

```
Benchmark 1: papis --lib papers list -a fluid
  Time (mean ± σ):     359.1 ms ±  29.4 ms    [User: 324.0 ms, System: 32.9 ms]
  Range (min … max):   327.7 ms … 410.4 ms    10 runs
```

#### After

```
Benchmark 1: papis --lib papers list -a fluid
  Time (mean ± σ):     347.0 ms ±  23.9 ms    [User: 312.9 ms, System: 31.8 ms]
  Range (min … max):   321.3 ms … 399.0 ms    10 runs
```

alexfikl:
@supersambo I'll go ahead and merge this. If you have a chance to play with it later, I'll be happy to do a follow up PR to solve any remaining issues :grin:

PR #1171, opened 2026-04-18
Rename normalize_path to normalize_path_part

alexfikl:
`normalize_path` did not work on paths (either as strings or as `pathlib.Path`), so the name was quite confusing. This adds a little `_part` to it and updates the documentation to hopefully make it clearer.

cc @jghauser

### DIFF
--- a/doc/source/default_settings.rst
+++ b/doc/source/default_settings.rst
@@ -197,14 +197,13 @@ General settings
 
     By default document paths in Papis libraries can contain only a limited set
     of characters. This is mainly to exclude characters that are invalid for
-    file paths on any operating system or possibly unprintable. Allowed
-    characters are:
+    file paths on any operating system or possibly unprintable. Currently, the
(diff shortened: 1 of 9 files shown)


jghauser:
Thanks, this looks good to me! Out of curiosity, what's your policy on when to add a deprecation warning? (I suppose having a nice official API will make that easier too)

jghauser (approved):

alexfikl:
> Thanks, this looks good to me! Out of curiosity, what's your policy on when to add a deprecation warning? (I suppose having a nice official API will make that easier too)

I try to add a deprecation warning every time when breaking something, usually with a one release window. I doubt anyone is actually using these things (since they're pretty new anyway), but just in case.

I think the only "official" API is the one in `papis.api` and that hasn't changed in a long time. There's a bunch of semi-official stuff like `Document` and `papis.database` that gets used by plugins, but besides that..

PR #1173, opened 2026-04-21
New FastAPI server

jghauser:
As discussed in various places in the past (#1020, #425 and probably elsewhere), Papis is being held back by the lack of a proper client/server architecture with a clear API. There are various benefits for this, for instance:

- TUIs (and other interfaces) can get event messages when the database state changes
- Various plugins etc can use a stable API (our current API is very very limited)
- TUI and web interface can use the same API
- A Papis server can run on a host different from the client

This PR implements the first part of this architecture: a FastAPI app that provides a (hopefully) consistent API that enables all of Papis's main functionality.

To try it out, get this branch, make sure to have the new dependencies installed, and then run `uvicorn papis.server.app:app --reload`. You can then access `http://127.0.0.1:8000/redoc` and `http://127.0.0.1:8000/docs` where the API is documented. The `docs` version is particularly useful as it's interactive and allows you to try out the API.

Much of Papis's main functionality has been implemented. We are still missing two main things:

- [ ] `papis config`
- [ ] `papis doctor`

as well as various arguments such as `--git`, `--link`, etc (I still need to make a list of what exactly is needed). We might also want to implement paging, getting only specific fields of a document, async operations, and so on. We also do not yet implement a websocket with which clients could get updates about database changes.

For now, I'd be very glad to get feedback from users on any issues you see with the current API design. I'm sure there are things that could be done more elegantly and use cases I've somehow not considered.

Note also that this is the first piece of work funded by NLnet :heart:

PS: Sorry about the big commits. I was trying to keep things parceled up into neat little bundles, but things got out of hand as I kept adjusting a million things... If it's very hard to navigate let me know, and I can give separating things another go :sweat_smile: 

EDIT: Here's a high level overview of the API design: https://mis.julianhauser.com/s/bM4sY8KnfdcBKKP. Feel free to leave notes and comments.

### DIFF
--- a/papis/server/exceptions.py
+++ b/papis/server/exceptions.py
@@ -1,0 +1,163 @@
+"""API exception classes and error codes.
+
+All exceptions inherit from :class:`APIError` (which extends FastAPI's
+:class:`HTTPException`).
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from fastapi import FastAPI, HTTPException, Request
+from fastapi.responses import JSONResponse
+
+import papis.logging
+from papis.server.models import ErrorDetail, ErrorResponse
+
+logger = papis.logging.get_logger(__name__)
+
+
+class APIError(HTTPException):
+    """Base class for API errors.
+
+    All subclasses produce a structured ``detail`` dict (an :class:`ErrorDetail`).
+    Subclasses declare their HTTP status code. Each instance carries a
+    machine-readable ``code`` string that clients use to distinguish error
+    conditions.
+    """
+
+    status_code: int
+    description: str = "Error"
+
+    def __init__(
+        self,
+        message: str = "",
+        *,
+        code: str,
+        context: dict[str, Any] | None = None,
+    ) -> None:
+        detail = ErrorDetail(
+            code=code,
+            message=message,
+            context=context,
+        ).model_dump(exclude_none=True)
+        super().__init__(
+            status_code=self.status_code,
+            detail=detail,
+        )
+
+    @classmethod
+    def responses(
+        cls, codes: list[str] | None = None
+    ) -> dict[str | int, dict[str, Any]]:
+        description = cls.description
+        if codes:
+            description += ". ``code`` is one of: " + ", ".join(codes)
+        return {
+            cls.status_code: {
+                "model": ErrorResponse,
+                "description": description,
+            }
+        }
+
+
+class ErrorCode:
+    # 400
+    INVALID_JSON = "invalid_json"
+    PATH_ESCAPE = "path_escape"
+    FILES_IN_METADATA_NOT_ALLOWED = "read_only_field"
+    READ_ONLY_FIELD = "read_only_field"
+    INVALID_URI = "invalid_uri"
+    IMPORTER_NO_DATA = "importer_no_data"
+    MUTUALLY_EXCLUSIVE = "mutually_exclusive"
+    UNKNOWN_EXPORT_FORMAT = "unknown_export_format"
+    CITATION_NO_DOI = "citation_no_doi"
+    CITATION_FETCH_EMPTY = "citation_fetch_empty"
+    LOCAL_MODE_REQUIRED = "local_mode_required"
+    # 404
+    LIBRARY_NOT_FOUND = "library_not_found"
+    DOCUMENT_NOT_FOUND = "document_not_found"
+    FILE_NOT_FOUND = "file_not_found"
+    NOTE_NOT_FOUND = "note_not_found"
+    IMPORTER_NOT_FOUND = "importer_not_found"
+    CHECK_NOT_FOUND = "check_not_found"
+    # 409
+    NOTES_EXIST = "notes_exist"
+    # 412
+    NOT_A_GIT_REPOSITORY = "not_a_git_repository"
+    # 502
+    UPSTREAM_ERROR = "upstream_error"
+    # 500 (internal, only used in the catch-all handler)
+    INTERNAL_SERVER_ERROR = "internal_server_error"
+
+
+class BadRequestError(APIError):
+    """Bad request (HTTP 400)."""
+
+    status_code = 400
+    description = "Bad request"
+
+
+class ResourceNotFoundError(APIError):
+    """Resource not found (HTTP 404)."""
+
+    status_code = 404
+    description = "Resource not found"
+
+
+class ConflictError(APIError):
+    """Conflict (HTTP 409)."""
+
+    status_code = 409
+    description = "Conflict"
+
+
+class PreconditionFailedError(APIError):
+    """Precondition failed (HTTP 412)."""
+
+    status_code = 412
+    description = "Precondition failed"
+
+
+class InternalServerError(APIError):
+    """Internal server error (HTTP 500)."""
+
+    status_code = 500
+    description = "Internal server error"
+
+
+class UpstreamError(APIError):
+    """Upstream error (HTTP 502)."""
+
+    status_code = 502
+    description = "Upstream error"
+
+
+def register_exception_handlers(app: FastAPI) -> None:
+    """Register custom exception handlers on the FastAPI application.
+
+    This installs a catch-all handler for all exceptions that aren't explicitly
+    handled by the FastAPI server.
+    """
+
+    @app.exception_handler(Exception)
+    async def unhandled_exception_handler(
+        request: Request, exc: Exception
+    ) -> JSONResponse:
+        logger.error(
+            "Unhandled exception in request %s %s: %r",
+            request.method,
+            request.url,
+            exc,
+            exc_info=True,
+        )
+        return JSONResponse(
+            status_code=500,
+            content={
+                "detail": InternalServerError(
+                    "Internal server error. The database may be in a bad state."
+                    " Try running 'papis doctor' to diagnose.",
+                    code=ErrorCode.INTERNAL_SERVER_ERROR,
+                ).detail
+            },
+        )
--- a/papis/server/exceptions.py
+++ b/papis/server/exceptions.py
@@ -1,0 +1,200 @@
+"""API exception classes and error codes.
+
+All exceptions inherit from :class:`APIError` (which extends FastAPI's
+:class:`HTTPException`).
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from fastapi import FastAPI, HTTPException, Request
+from fastapi.exceptions import RequestValidationError
+from fastapi.responses import JSONResponse
+
+import papis.logging
+from papis.server.models import ProblemResponse
+
+logger = papis.logging.get_logger(__name__)
+
+
+class APIError(HTTPException):
+    """Base class for API errors.
+
+    Errors are returned as `RFC 7807 <https://tools.ietf.org/html/rfc7807>`_ problem
+    detail objects. Subclasses declare their HTTP status code and a short ``title``.
+    Each instance carries a machine-readable ``type`` that clients may use to
+    distinguish error conditions.
+    """
+
+    status_code: int
+    title: str = "Error"
+
+    def __init__(
+        self,
+        detail: str = "",
+        *,
+        type: str,
+        context: dict[str, Any] | None = None,
+    ) -> None:
+        self.problem = ProblemResponse(
+            type=type,
+            title=self.title,
+            status=self.status_code,
+            detail=detail,
+            context=context,
+        )
+        super().__init__(
+            status_code=self.status_code,
+            detail=self.problem.model_dump(exclude_none=True),
+        )
+
+    @classmethod
+    def responses(
+        cls, types: list[str] | None = None
+    ) -> dict[str | int, dict[str, Any]]:
+        title_str = cls.title
+        if types:
+            title_str += ". ``type`` is one of: " + ", ".join(types)
+        return {
+            cls.status_code: {
+                "model": ProblemResponse,
+                "description": title_str,
+            }
+        }
+
+
+class ErrorCode:
+    # 400
+    INVALID_JSON = "/errors/invalid-json"
+    PATH_ESCAPE = "/errors/path-escape"
+    IMMUTABLE_FIELD = "/errors/immutable-field"
+    INVALID_URI = "/errors/invalid-uri"
+    IMPORTER_NO_DATA = "/errors/importer-no-data"
+    MISSING_FIELD = "/errors/missing-field"
+    MUTUALLY_EXCLUSIVE = "/errors/mutually-exclusive"
+    UNKNOWN_EXPORT_FORMAT = "/errors/unknown-export-format"
+    CITATION_NO_DOI = "/errors/citation-no-doi"
+    CITATION_FETCH_EMPTY = "/errors/citation-fetch-empty"
+    LOCAL_MODE_REQUIRED = "/errors/local-mode-required"
+    VALIDATION_ERROR = "/errors/validation-error"
+    # 404
+    LIBRARY_NOT_FOUND = "/errors/library-not-found"
+    DOCUMENT_NOT_FOUND = "/errors/document-not-found"
+    FILE_NOT_FOUND = "/errors/file-not-found"
+    NOTE_NOT_FOUND = "/errors/note-not-found"
+    IMPORTER_NOT_FOUND = "/errors/importer-not-found"
+    CHECK_NOT_FOUND = "/errors/check-not-found"
+    # 409
+    NOTES_EXIST = "/errors/notes-exist"
+    FOLDER_EXISTS = "/errors/folder-exists"
+    FOLDER_INSIDE_DOCUMENT = "/errors/folder-inside-document"
+    FILE_EXISTS = "/errors/file-exists"
+    # 412
+    NOT_A_GIT_REPOSITORY = "/errors/not-a-git-repository"
+    VERSION_MISMATCH = "/errors/version-mismatch"
+    # 502
+    UPSTREAM_ERROR = "/errors/upstream-error"
+    # 500 (internal, only used in the catch-all handler)
+    INTERNAL_SERVER_ERROR = "/errors/internal-server-error"
+
+
+class BadRequestError(APIError):
+    """Bad request (HTTP 400)."""
+
+    status_code = 400
+    title = "Bad request"
+
+
+class ResourceNotFoundError(APIError):
+    """Resource not found (HTTP 404)."""
+
+    status_code = 404
+    title = "Resource not found"
+
+
+class ConflictError(APIError):
+    """Conflict (HTTP 409)."""
+
+    status_code = 409
+    title = "Conflict"
+
+
+class PreconditionFailedError(APIError):
+    """Precondition failed (HTTP 412)."""
+
+    status_code = 412
+    title = "Precondition failed"
+
+
+class InternalServerError(APIError):
+    """Internal server error (HTTP 500)."""
+
+    status_code = 500
+    title = "Internal server error"
+
+
+class UpstreamError(APIError):
+    """Upstream error (HTTP 502)."""
+
+    status_code = 502
+    title = "Upstream error"
+
+
+def register_exception_handlers(app: FastAPI) -> None:
+    """Register custom exception handlers on the FastAPI application.
+
+    Installs an :class:`APIError` handler that returns ``application/problem+json``
+    (RFC 7807), normalises ``RequestValidationError`` to the same format, and adds a
+    catch-all for unhandled exceptions.
+    """
+
+    @app.exception_handler(APIError)
+    async def api_error_handler(_request: Request, exc: APIError) -> JSONResponse:
+        return JSONResponse(
+            status_code=exc.status_code,
+            content=exc.problem.model_dump(exclude_none=True),
+            media_type="application/problem+json",
+        )
+
+    @app.exception_handler(RequestValidationError)
+    async def validation_error_handler(
+        _request: Request, exc: RequestValidationError
+    ) -> JSONResponse:
+        errors = exc.errors()
+        detail = "; ".join(
+            f"{'.'.join(str(loc) for loc in err['loc'])}: {err['msg']}"
+            for err in errors
+        )
+        return JSONResponse(
+            status_code=422,
+            content=ProblemResponse(
+                type=ErrorCode.VALIDATION_ERROR,
+                title="Validation error",
+                status=422,
+                detail=detail,
+                context={"errors": errors},
+            ).model_dump(exclude_none=True),
+            media_type="application/problem+json",
+        )
+
+    @app.exception_handler(Exception)
+    async def unhandled_exception_handler(
+        request: Request, exc: Exception
+    ) -> JSONResponse:
+        logger.error(
+            "Unhandled exception in request %s %s: %r",
+            request.method,
+            request.url,
+            exc,
+            exc_info=True,
+        )
+        return JSONResponse(
+            status_code=500,
+            content=InternalServerError(
+                "Internal server error. The database may be in a bad state."
+                " Try running 'papis doctor' to diagnose.",
+                type=ErrorCode.INTERNAL_SERVER_ERROR,
+            ).problem.model_dump(exclude_none=True),
+            media_type="application/problem+json",
+        )
--- a/papis/server/models.py
+++ b/papis/server/models.py
@@ -1,0 +1,328 @@
+"""Pydantic models for API request and response bodies."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+from pydantic import BaseModel, ConfigDict, Field, create_model
+
+from papis.document import get_document_field_types
+from papis.id import ID_KEY_NAME
+
+if TYPE_CHECKING:
+    from papis.document import Document
+
+
+class ErrorDetail(BaseModel):
+    """Structured error detail returned by the API."""
+
+    code: str = Field(description="Machine-readable error code.")
+    message: str = Field(description="Human-readable explanation of the error.")
+    context: dict[str, Any] | None = Field(
+        default=None,
+        description=(
+            "Optional key/value pairs identifying the resource(s) involved (e.g."
+            " ``library``, ``id``, ``file``)."
+        ),
+    )
+
+
+class ErrorResponse(BaseModel):
+    """Top-level error response matching FastAPI's ``{\"detail\": ...}`` convention."""
+
+    detail: ErrorDetail
+
+
+def _document_field_definitions() -> dict[str, Any]:
+    """Field definitions for the document models.
+
+    Each field is optional (``T | None``) and unset by default so that partial
+    updates and sparse creations are allowed. Document metadata keys mirror
+    the ``info.yaml`` fields.
+    """
+    key_types = get_document_field_types()
+    return {name: (field_type | None, None) for name, field_type in key_types.items()}
+
+
+if TYPE_CHECKING:
+
+    class PapisRequestDocument(BaseModel):
+        model_config = ConfigDict(extra="allow")
+
+    class PapisResponseDocument(BaseModel):
+        model_config = ConfigDict(extra="allow")
+        papis_id: str
+
+else:
+    PapisRequestDocument = create_model(
+        "DocumentRequest",
+        **_document_field_definitions(),
+        __config__=ConfigDict(extra="allow"),
+        __doc__=(
+            "Document metadata for create/update requests.\n\n"
+            "Only the known fields listed below are type-checked. Additional"
+            " fields are accepted. All fields are optional."
+        ),
+    )
+    PapisResponseDocument = create_model(
+        "DocumentResponse",
+        **_document_field_definitions(),
+        **{ID_KEY_NAME: (str, ...)},
+        __config__=ConfigDict(extra="allow"),
+        __doc__=(
+            "Document metadata returned by read endpoints.\n\n"
+            "Mirrors :class:`DocumentRequest` but requires the ``ID_KEY_NAME`` field"
+            " (i.e. ``papis_id``)."
+        ),
+    )
+
+
+def document_to_response_model(doc: Document) -> PapisResponseDocument:
+    """Convert a papis Document to the API Document model."""
+
+    data = {k: v for k, v in doc.items() if k != ID_KEY_NAME}
+    return PapisResponseDocument(**{ID_KEY_NAME: doc[ID_KEY_NAME]}, **data)
+
+
+class DocumentResponse(BaseModel):
+    """A single document with its folder path."""
+
+    document: PapisResponseDocument = Field(description="The document.")
+    folder: str | None = Field(
+        default=None,
+        description="Document folder path relative to the library root.",
+    )
+
+
+class DocumentsResponse(BaseModel):
+    """List of documents with their folder paths."""
+
+    documents: list[PapisResponseDocument] = Field(description="List of documents.")
+    folders: dict[str, str] = Field(
+        description=(
+            "Mapping of document ``papis_id`` to folder path relative to the"
+            " library root."
+        ),
+    )
+
+
+class DocumentRequest(BaseModel):
+    """Request body for ``PATCH /documents/{id}``."""
+
+    data: PapisRequestDocument | None = Field(
+        default=None,
+        description="Key/value pairs to merge into the document metadata.",
+    )
+    folder: str | None = Field(
+        default=None,
+        description=(
+            "New folder path or format pattern (``null`` for the"
+            " ``add-folder-name`` default)."
+        ),
+    )
+
+
+class LibraryResponse(BaseModel):
+    """Information about a library.
+
+    The library path is only included in local mode.
+    """
+
+    name: str = Field(description="Name of the library.")
+    path: str | None = Field(default=None, description="Library path.")
+
+
+class LibrariesResponse(BaseModel):
+    """List of libraries."""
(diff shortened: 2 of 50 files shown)


alexfikl:
> For now, I'd be very glad to get feedback from users on any issues you see with the current API design. I'm sure there are things that could be done more elegantly and use cases I've somehow not considered.

3k lines of code are a lot to review, haha. I have some starting questions:
1. Is there some central place I can look at to see what the "API design" is?
2. Is there an example / command on how to use this stuff? Currently it doesn't seem to be wired into anything.
4. "There are various benefits for this, for instance:" -- Do you have some rough idea of how any of these things will look like? e.g. How would this help plugins?

jghauser:
> > For now, I'd be very glad to get feedback from users on any issues you see with the current API design. I'm sure there are things that could be done more elegantly and use cases I've somehow not considered.
> 
> 3k lines of code are a lot to review, haha. I have some starting questions:

Yeah, sorry about that! I can parcel it up by route (one commit per route, so 'documents', 'cache', 'files', etc), that should work quite nicely for the most part. Though the total amount of lines of code wouldn't really change ;). In my defense, most of it is probably tests...

>     1. Is there some central place I can look at to see what the "API design" is?

I have a doc with my own notes. I'll tidy it up and upload it somewhere. For now, I think checking out the automatic docs (mentioned in my first post) should give a good overview.

>     2. Is there an example / command on how to use this stuff? Currently it doesn't seem to be wired into anything.

Yes, it's not wired into anything at the moment. I think the general structure of how to use the API can be seen in the tests. Basically, consumers send GET/POST/etc requests to the endpoints (sometimes with json payloads) and get some response back (also json). My idea is that all the database/info.yaml/files stuff should be handled centrally and clients effect changes via the API.

>     3. "There are various benefits for this, for instance:" -- Do you have some rough idea of how any of these things will look like? e.g. How would this help plugins?

- For plugins: here the advantage really is just that they can have a dependable API and do no longer need to use various internal bits of Papis
- For TUI/Web interface: I want the TUI to get notified if the database state changes (e.g. if I add a document with `papis add` CLI, I want the TUI to know that the database has changed).
- My papis.nvim plugin: I'm actually now thinking to use the API as well (rather than sqlite directly). It's easier to implement in neovim (since it speaks json well) and safer
- The web interface can also be modernised using the new API (it implements its own API at the moment, but it's not very consistent)
- If we have Papis running on some server, we will be able to access it not only via the web interface, but also via e.g. the CLI (or, here I'm dreaming, with an Android app on the tablet).

jghauser:
@alexfikl, I've added a short design doc to my first post.

alexfikl:
> Yeah, sorry about that! I can parcel it up by route

No, I don't think that's worth it. It's just a big chunk of code so breaking up the commits probably doesn't help.
 
> I have a doc with my own notes. I'll tidy it up and upload it somewhere. For now, I think checking out the automatic docs (mentioned in my first post) should give a good overview.

Thanks! We should probably have these somewhere. Maybe using [sphinx-openapi](https://sphinxcontrib-openapi.readthedocs.io/)? Or `sphinx-redoc`? Not sure which one of these are better maintained these days.

> the automatic docs (mentioned in my first post)

:facepalm: Yeah, that's perfect! I just missed it, sorry :D

> * For plugins: here the advantage really is just that they can have a dependable API and do no longer need to use various internal bits of Papis

I'm not sure this is a convincing argument, since we can just add more stuff to `papis.api` and make a promise to not break it. All this does is add a serialization and deserialization step in between, no?

> * For TUI/Web interface: I want the TUI to get notified if the database state changes (e.g. if I add a document with `papis add` CLI, I want the TUI to know that the database has changed).

This is a "new" more persistent TUI, not the current version, right?

> * My papis.nvim plugin: I'm actually now thinking to use the API as well (rather than sqlite directly). It's easier to implement in neovim (since it speaks json well) and safer

So going forward we want to recommend users just start the server on startup? I assume we'd like more things to go through it, since it should be faster (e.g. doesn't need to reload things from disk).

> * If we have Papis running on some server, we will be able to access it not only via the web interface, but also via e.g. the CLI (or, here I'm dreaming, with an Android app on the tablet).

We'd probably need a `papis-client` sort of things to migrate CLI commands to using the server? Any plans there?

papis/server/models.py
```diff
@@ -0,0 +1,90 @@
...
+        **{ID_KEY_NAME: (str, ...)},
+        __config__=ConfigDict(extra="allow"),
+    )
+
+
+class LibraryResponse(BaseModel):
+    """Information about a library."""
+
+    name: str
+    # path: str  # NOTE: for security reasons, we don't return this
```
alexfikl: This seems like a weird comment given that we can get document folders? Are those also relative to the library? From a quick skim of `routes.documents`, it doesn't seem so?

pyproject.toml
```diff
@@ -235,6 +237,7 @@ lib = "papis.importer.library:FromLibraryImporter"
 pdf2arxivid = "papis.importer.arxiv:ArxivFromPDFImporter"
 pdf2doi = "papis.importer.doi:DOIFromPDFImporter"
 pmid = "papis.importer.pubmed:PubMedImporter"
+url = "papis.downloaders:WebImporter"
```
alexfikl, line 239: This should probably go into a separate small PR? :grin:

papis/server/models.py
jghauser: All paths returned should indeed be relative to the library/document. Check how e.g. `GET /libraries/{library}/documents/{id}/folder` returns the doc folder path relative to the library path. Or did I miss it somewhere?

alexfikl: Ah, yeah, you're right! I just looked at the `get_folder` function :facepalm:

papis/server/routes/libraries.py
```diff
@@ -0,0 +1,41 @@
...
+
+def get_library_info(library: str) -> tuple[str, pathlib.Path]:
+    """Get library information from the library name."""
+    try:
+        lib = papis.config.get_lib_from_name(library)
+    except papis.exceptions.InvalidLibraryError as e:
+        raise api_e.ResourceNotFoundError(f"Library '{library}' not found") from e
+    return lib.name, pathlib.Path(
+        lib.paths[0]
+    )  # NOTE: simplify once we only support single path
```
alexfikl: I think it's fine to make a separate PR that just makes it one path per library? We can do a little Github search (or through our own issues), but I'm not sure I've ever seen multiple paths..

alexfikl: Maybe it's used for subfolders? But `papis.utils.get_folders` is already recursive..

jghauser:
> Thanks! We should probably have these somewhere. Maybe using [sphinx-openapi](https://sphinxcontrib-openapi.readthedocs.io/)? Or `sphinx-redoc`? Not sure which one of these are better maintained these days.

Yes, totally. From what I see sphinx-openapi seems to be maintained these days.
 
> > * For plugins: here the advantage really is just that they can have a dependable API and do no longer need to use various internal bits of Papis
> 
> I'm not sure this is a convincing argument, since we can just add more stuff to `papis.api` and make a promise to not break it. All this does is add a serialization and deserialization step in between, no?

I meant more that at least this way we finally have a decent API ;). It's true that just for plugins this could have been done differently.

> > * For TUI/Web interface: I want the TUI to get notified if the database state changes (e.g. if I add a document with `papis add` CLI, I want the TUI to know that the database has changed).
> 
> This is a "new" more persistent TUI, not the current version, right?

Yes, totally.

> > * My papis.nvim plugin: I'm actually now thinking to use the API as well (rather than sqlite directly). It's easier to implement in neovim (since it speaks json well) and safer
> 
> So going forward we want to recommend users just start the server on startup? I assume we'd like more things to go through it, since it should be faster (e.g. doesn't need to reload things from disk).

I would say so. I think we should just auto-start the server in the background on the first invocation of a Papis command (of course we should make this configurable).

> > * If we have Papis running on some server, we will be able to access it not only via the web interface, but also via e.g. the CLI (or, here I'm dreaming, with an Android app on the tablet).
> 
> We'd probably need a `papis-client` sort of things to migrate CLI commands to using the server? Any plans there?

I suppose there are various options here. I thought to just make the current Papis the client. So the current commands would check if a server is running (or if a remote server is defined) and connect to that. Why do you think we need a separate `papis-client`?

papis/server/routes/libraries.py
jghauser: Haha, yeah, I agree. I don't quite understad this either. Also, what even happens when you add a document and have multiple `dirs` defined? :thinking:

alexfikl:
> I meant more that at least this way we finally have a decent API ;). It's true that just for plugins this could have been done differently.

I think that's the argument in general. This approach is only better if
1. You can't call the Python Papis library (assuming a similar API is stabilized), e.g. in a Lua neovim plugin :grin:
2. Your Papis library is remote somewhere.

Otherwise, you're just incurring additional costs of serializing and deserializing things all the time. Did you have some other use cases in mind?

> I suppose there are various options here. I thought to just make the current Papis the client. So the current commands would check if a server is running (or if a remote server is defined) and connect to that. Why do you think we need a separate papis-client?

I mostly suggested a separate `papis-client` to allow slowly migrating things over to the server architecture. I assume doing a one shot port of all the commands will be quite difficult. Especially since not all the functionality is there, so most things will be a mix between calling the server and calling the older functions.

Long run, I agree, it should be merged into just one thing.

EDIT: Although maybe that will just ensure that they never get merged back together? :\ Since the two will diverge and just become harder to maintain.. ok, might need some more careful planning given what we'll actually have time for :grin:

papis/server/routes/libraries.py
alexfikl: We carefully and thoughtfully choose the best one :grin:
```python
    # create a nice folder name for the new document
    if base_path is None:
        base_path = os.path.expanduser(papis.config.get_lib_dirs()[0])
```

jghauser:
> I think that's the argument in general. This approach is only better if
> 
>     1. You can't call the Python Papis library (assuming a similar API is stabilized), e.g. in a Lua neovim plugin 😁
> 
>     2. Your Papis library is remote somewhere.
> 
> 
> Otherwise, you're just incurring additional costs of serializing and deserializing things all the time. Did you have some other use cases in mind?

The way I see it is that FastAPI with http endpoints is simple, popular, cross-platform, and works for the local and remote use-case. There are other options, but they all are either more complex or can't do one or the other thing. It's true that this means that in the pure python local case, there is additional overhead. But from my research (and given what Papis does), I do not think this is significant. What slows down Papis are database calls and (currently) having no persistent server.

Basically, my point is that we want to support the remote case (and a web interface + neovim lua plugins) and that constrains our options. In order to keep things as simple as possible, we should then use this option (fastapi) for the other clients too since doing so won't slow things down (in any significant way). For maintainability, it makes sense to converge on one API, especially given that the relevant tooling is easy to understand and well supported.
 
> I mostly suggested a separate `papis-client` to allow slowly migrating things over to the server architecture. I assume doing a one shot port of all the commands will be quite difficult. Especially since not all the functionality is there, so most things will be a mix between calling the server and calling the older functions.
> 
> Long run, I agree, it should be merged into just one thing.
> 
> EDIT: Although maybe that will just ensure that they never get merged back together? :\ Since the two will diverge and just become harder to maintain.. ok, might need some more careful planning given what we'll actually have time for 😁

Yeah, I agree. This needs to be carefully planned. It may make sense to change each command so that it supports both the current local functionality and the API. And then eventually, we might want to phase out the local case. There's also the complication that some commands/arguments can't work or don't make sense over the API (at least in the remote case).

alexfikl:
> What slows down Papis are database calls and (currently) having no persistent server.

I also assume that's true, but haven't really benchmarked it. On my machine, a `papis list -a QUERY` takes about 120ms with the `sqlite` backend. I think the `papis` backend was at about 300ms and the whoosh one is slower.

Last time I checked, most of the remaining time is spent around `click` and loading plugin entrypoints. With the server in the background, some of the plugin loading can probably be nicely cached, but otherwise it should be pretty similar.

> Basically, my point is that we want to support the remote case (and a web interface + neovim lua plugins) and that constrains our options. In order to keep things as simple as possible, we should then use this option (fastapi) for the other clients too since doing so won't slow things down (in any significant way). For maintainability, it makes sense to converge on one API, especially given that the relevant tooling is easy to understand and well supported.

I agree that for the remote case this is pretty much a requirement.. and, since duplication isn't great, we should converge around that. I'm offering a bit of pushback because personally I probably won't be using the server stuff unless it becomes mandatory (no remote usage, don't have a tablet, etc. haha). So my worry is that this will just lay dormant like the current `papis/web` stuff :-s (which I assume works, but rarely try it out).

jghauser:
> > What slows down Papis are database calls and (currently) having no persistent server.
> 
> I also assume that's true, but haven't really benchmarked it. On my machine, a `papis list -a QUERY` takes about 120ms with the `sqlite` backend. I think the `papis` backend was at about 300ms and the whoosh one is slower.
> 
> Last time I checked, most of the remaining time is spent around `click` and loading plugin entrypoints. With the server in the background, some of the plugin loading can probably be nicely cached, but otherwise it should be pretty similar.

Yeah totally. I think FastAPI should only add single digit ms overhead, though I haven't tested it yet. Also because FastAPI is very popular, there are many ways of further improving performance.

> I agree that for the remote case this is pretty much a requirement.. and, since duplication isn't great, we should converge around that. I'm offering a bit of pushback because personally I probably won't be using the server stuff unless it becomes mandatory (no remote usage, don't have a tablet, etc. haha). So my worry is that this will just lay dormant like the current `papis/web` stuff :-s (which I assume works, but rarely try it out).

I for one will be using this functionality. Definitely for my neovim plugin and the new persistent TUI. But I'm also interested in the remote server connecting to my laptop and tablet over tailscale. Many cool possibilities :). And I think if the webapp gets tidied up a bit, it might be useful to a few people :).

alexfikl:
> I for one will be using this functionality. Definitely for my neovim plugin and the new persistent TUI. But I'm also interested in the remote server connecting to my laptop and tablet over tailscale. Many cool possibilities :). And I think if the webapp gets tidied up a bit, it might be useful to a few people :).

Fair enough! As long as it's gonna see usage and maintenance, it's all good :grin:

I'll try to go over the code during the weekend, but the general API endpoints look quite reasonable, so it's a great start.

papis/server/routes/cache.py
```diff
@@ -0,0 +1,19 @@
...
+
+from typing import Annotated
+
+from fastapi import APIRouter, Path, Response
+
+import papis.api
+from papis.server.api import API_V1
+from papis.server.routes.libraries import get_library_info
+
+router = APIRouter(prefix=API_V1)
```
alexfikl: Is it common to create one of these as a global variable in each module?

papis/server/routes/cache.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+@router.delete("/libraries/{library}/cache", tags=["Cache"])
+def clear_cache(library: Annotated[str, Path(description="Library name")]) -> Response:
+    """Clear the document cache for a library."""
+    _, _ = get_library_info(library)
+    papis.api.clear_lib_cache(library)
```
alexfikl: Is this meant to just clear, not rebuild? i.e. this doesn't `db.initialize`, so the next query will likely be against an empty cache.

papis/server/routes/documents.py
```diff
@@ -0,0 +1,280 @@
...
+
+if TYPE_CHECKING:
+    from papis.database.base import Database
+
+router = APIRouter(prefix=API_V1)
+
+
+def get_db(library: str) -> Database:
+    """Get database for a library name."""
+    _, _ = get_library_info(library)
```
alexfikl:
```suggestion
```
? not quite sure I understand the point of this wrapper function.

papis/server/exceptions.py
```diff
@@ -0,0 +1,92 @@
+from __future__ import annotations
+
+from typing import Any
+
+from fastapi import HTTPException
+
+
+class ApiError(HTTPException):
```
alexfikl:
```suggestion
class APIError(HTTPException):
```
? Feels weird to just capitalize the first one.

papis/server/routes/documents.py
```diff
@@ -0,0 +1,280 @@
...
+    return papis.database.get(library)
+
+
+def get_doc(db: Database, id: str) -> Document:
+    """Get a document by ID, raising 404 if not found."""
+    doc = db.find_by_id(id)
+    if not doc:
+        raise api_e.ResourceNotFoundError("Document not found")
```
alexfikl:
```suggestion
        raise api_e.ResourceNotFoundError(f"Document not found: {id}")
```
? Not quite sure where this error ends up, but it's usually nice to get more info about what failed. Maybe even some info about the database it tried to search? (same in other places this is used)

papis/server/routes/documents.py
```diff
@@ -0,0 +1,280 @@
...
+    if not folder or not pathlib.Path(folder).exists():
+        raise api_e.ResourceNotFoundError("Document folder not found")
+    return pathlib.Path(folder)
+
+
+def _normalize_doc_path(path: str) -> pathlib.Path:
+    """Splits the path into components, normalizes each one, then rejoins."""
+    parts = pathlib.Path(path).parts
+    normalized = [normalize_path_part(c) for c in parts if c and c != "."]
+    return pathlib.Path(*normalized) if normalized else Path()
```
alexfikl:
```suggestion
    return pathlib.Path(*normalized)
```
? This should be the same as far as I know. Also, not sure where this is used, but is it safe to return `"."` as the normalized doc path?

papis/server/routes/exporters.py
```diff
@@ -0,0 +1,71 @@
...
+router = APIRouter(prefix=API_V1)
+
+_EXPORT_CONTENT_TYPES: dict[str, str] = {
+    # Special default case
+    "json": "application/json",
+    # Other formats
+    "bibtex": "text/x-bibtex",
+    "yaml": "application/yaml",
+    "csl": "application/vnd.citationstyles.style+xml",
+    "typst": "text/vnd.typst",
```
alexfikl: These seem to be used for exporting the documents? For that:
* `csl`: this is just text, the exporter doesn't write out XML or anything like that (that's just used to define the styles), i.e. you get something like
```
Greenstadt, J., 1972. A Quasi-Newton Method With No Derivatives. Mathematics of Computation, 26(117), pp.145–145.
```
* `typst`: the export we do is to YAML (for [hayagriva](https://github.com/typst/hayagriva)), not as a typst source file.

papis/server/routes/documents.py
```diff
@@ -0,0 +1,280 @@
...
+    "/libraries/{library}/documents",
+    tags=["Documents"],
+    response_model=list[DocumentResponse],
+    response_model_exclude_none=True,
+    responses={**DOCUMENT_RESPONSES, **api_e.UnknownExportFormat.responses()},
+)
+def get_documents(
+    library: Annotated[str, Path(description="Library name")],
+    q: str | None = Query(None, description="Query string to filter documents"),
+    format: str = Query(default="json", description="Export format"),
```
alexfikl: Is it a good idea to bundle the "get documents" with "get documents in format"? I would sort of expect a separate export API.

papis/server/models.py
```diff
@@ -0,0 +1,90 @@
...
+
+    importer: str
+    data: dict[str, Any]
+
+
+def document_to_output_model(doc: Document) -> DocumentResponse:
+    """Convert a papis Document to the API Document model."""
+
+    data = {k: v for k, v in doc.items() if k != ID_KEY_NAME}
+    papis_id = doc.get(ID_KEY_NAME, "")
```
alexfikl: Is the `papis_id` ever allowed to be missing here? That only happens when you create a new document from scratch.

papis/server/routes/documents.py
```diff
@@ -0,0 +1,280 @@
...
+) -> DocumentResponse | Response:
+    """Get a specific document by ID."""
+    db = get_db(library)
+    doc = get_doc(db, id)
+
+    validate_format(format)
+
+    if format != "json":
+        content, content_type = export_documents([doc], format)
+        filename = f"{doc.get_main_folder_name()}.{format}"
```
alexfikl: `get_main_folder_name` is an absolute path. Good idea to use it here?

papis/server/routes/documents.py
```diff
@@ -0,0 +1,280 @@
...
+    doc_data = {k: v for k, v in data.model_dump().items() if v is not None}
+    _reject_files_notes(doc_data)
+
+    doc_path_norm = lib_path / _normalize_doc_path(folder)
+    doc_path = ensure_within_root(doc_path_norm, lib_path)
+    doc_path = pathlib.Path(_make_unique_folder(doc_path))
+    doc_path.mkdir(parents=True, exist_ok=True)
+    doc = Document(folder=str(doc_path), data=doc_data)
+    db.maybe_compute_id(doc)
+    doc.save()
```
alexfikl: The fact that this doesn't use anything from `papis.commands.add` is a bit weird. We have a bunch of logic there for how we create document folders and things. For example, this is ignoring `add-folder-name`. Is it expected that the caller handle that?

papis/server/routes/documents.py
```diff
@@ -0,0 +1,280 @@
...
+    """Move a document to a new location within the library."""
+    _, lib_path = get_library_info(library)
+    db = get_db(library)
+    doc = get_doc(db, id)
+    doc_folder = get_folder(db, id)
+
+    doc_path_norm = lib_path / _normalize_doc_path(folder)
+    doc_path_new = ensure_within_root(doc_path_norm, lib_path)
+    doc_path_new = pathlib.Path(_make_unique_folder(doc_path_new))
+    shutil.move(doc_folder, doc_path_new)
```
alexfikl: This should also not use its own logic for handling the `doc_folder`. Needs to do whatever `papis mv` ends up doing. Or at least `papis.document.move`?

papis/server/routes/documents.py
```diff
@@ -0,0 +1,280 @@
...
+import pathlib
+import shutil
+from typing import TYPE_CHECKING, Annotated, Any
+
+from fastapi import APIRouter, Path, Query, Response
+
+import papis.database
+from papis.document import Document
+from papis.paths import _make_unique_folder, normalize_path_part
+from papis.server import exceptions as api_e
```
alexfikl: Why not just import the exceptions here? `api_e` is a weird name for `papis.server.exceptions`.

papis/server/routes/exporters.py
```diff
@@ -0,0 +1,71 @@
...
+    "csv": "text/csv",
+}
```
alexfikl: Would be nice to let the exporters tell us what mimetype they export? This would just become out of date as new exporters are added.

papis/server/routes/files.py
```diff
@@ -0,0 +1,186 @@
...
+
+if TYPE_CHECKING:
+    from papis.document import Document
+
+router = APIRouter(prefix=API_V1)
+
+
+def _get_file_path(
+    doc: Document,
+    doc_folder: pathlib.Path,
```
alexfikl, line 59: Is `doc_folder` expected to be different than `doc.get_main_folder`? Why pass it as a separate argument?

papis/server/routes/files.py
```diff
@@ -0,0 +1,186 @@
...
+@router.get("/libraries/{library}/documents/{id}/files", tags=["Files"])
+def list_files(
+    library: Annotated[str, Path(description="Library name")],
+    id: Annotated[str, Path(description="Document ID")],
+) -> list[FileName]:
+    """List files in a document."""
+    db = get_db(library)
+    doc = get_doc(db, id)
+
+    return doc.get("files") or []
```
alexfikl:
```suggestion
    return doc.get("files", [])
```
?

papis/server/routes/files.py
```diff
@@ -0,0 +1,186 @@
...
+    },
+)
+def add_file(
+    library: Annotated[str, Path(description="Library name")],
+    id: Annotated[str, Path(description="Document ID")],
+    upload: UploadFile = File(...),  # noqa: B008
+) -> FileName:
+    """Add a file to a document."""
+    if not upload.filename:
+        # NOTE: this is only reached if function isn't called over HTTP routes
```
alexfikl: When would that happen?

papis/server/routes/files.py
```diff
@@ -0,0 +1,186 @@
...
+        raise api_e.FilenameRequiredError("Filename is required")
+
+    db = get_db(library)
+    doc = get_doc(db, id)
+    doc_folder = get_folder(db, id)
+
+    file_name = normalize_path_part(upload.filename)
+    file_path = doc_folder / file_name
+    file_path = ensure_within_root(file_path, doc_folder)
+    file_path = pathlib.Path(_make_unique_file(file_path))
```
alexfikl: This ignores `add-file-name`?

papis/server/routes/files.py
```diff
@@ -0,0 +1,186 @@
...
+
+    with open(file_path, "wb") as dest:
+        shutil.copyfileobj(upload.file, dest)
+
+    doc.setdefault("files", [])
+    doc["files"].append(file_path.name)
```
alexfikl:
```suggestion
    doc.setdefault("files", []).append(file_path.name)
```
?

alexfikl (commented):
I got all the way to `papis.server.routes.file` and left a bunch of comments. I'll try to look through the rest soon :-s

jghauser:
Thanks a lot for having a look @alexfikl, will address these all soon!

papis/server/routes/cache.py
jghauser: Yeah no, that's definitely not the best way to do it. Found a way of setting it globally in `app.py` instead :)

papis/server/routes/cache.py
jghauser: Good point. I somehow thought that it would reinitialise automatically. But that's not the case. I now copy what `papis cache reset` does:

```python
    db.clear()
    db.initialize()
    _ = db.get_all_documents()
```

jghauser: Though I'm unsure whether `_ = db.get_all_documents()` is actually necessary?

papis/server/routes/documents.py
jghauser: It's because `get_library_info()` raises an appropriate error when the library doesn't exist. I can see it's not exactly elegant. Do you have some preference on how to best do this? Maybe it would be clearer if we used the first returned value (which is just the library name) in `papis.database.get()`. That might make the intent clearer?

papis/server/routes/documents.py
jghauser: Agree about having at least more info about the thing the error message is about in there (e.g. `id` if it's about a document, additionally filenames if it's about notes and files etc). I'm not sure if adding the library would be all that useful. It makes the messages a bit unwieldy. Though now I wonder whether having errors with structured data would be useful. That would be much easier for clients to consume

papis/server/routes/cache.py
alexfikl: I don't think the `get_all_documents` is required anymore.. it used to be a bug in either the `papis` or the `whoosh` backend, but it should be fixed.

papis/server/routes/documents.py
alexfikl: If this just checks if a library exists and raises an error, there should be a `raise_if_no_library(library)` or something like that, so that it's clear what we want. Or even an `if library.exists(): raise ...`?

papis/server/routes/documents.py
alexfikl: Yes! Some more structured errors would be great. I think most APIs I've played with just had an `{code: errorcode, message: msg}` return on error (we can add more metadata if it makes sense?).

alexfikl:
> Hey — I’m reaching out because I work with a small group of professional API testers (backend + automation engineers).

Thanks for reaching out! We're currently just feeling our way a bit around the API to see what's needed, so we're far from needing anything too professional in that direction. We'll keep it in mind though.

papis/server/routes/documents.py
jghauser: I mean it just does this in this particular instance, but in other instances, we do want the returned path. So the decision is between having two helper functions (one for just the error raising + one for getting the path) with clearer names or one helper function. The reason I did it like it's at the moment is because it mirrors other helper functions (that get something or other while also raising appropriate errors). One other option is to use the first value returned by `get_library_info()` (which is just the name of the library if it exists) and name it something clear à la `verified_lib_name` to show what's happening (similar to how it's currently done in `get_library()`).

papis/server/routes/documents.py
jghauser: Ok, I'll do that!

papis/server/routes/documents.py
alexfikl: I don't see an issue with having two different helper functions here, since they're doing different things. Like it is now (with the discarded outputs), it just looks like someone forgot to delete some debug code to me.

jghauser: Gotcha!

papis/server/routes/documents.py
jghauser: Thanks! And while safe to return `.`, it would lead to the creation of files in the library root which we do not want. So i'm adding an error for that :)

papis/server/routes/exporters.py
jghauser: Thanks, will change this!

papis/server/routes/documents.py
jghauser: I was thinking about this too. There are a couple of reasons I did it as it is now:
1. The export API endpoint would have to do pretty much exactly the same things as the documents endpoint. We'd have to recreate something like a `/export/libraries/{library}/documents/{id}/` structure to select the appropriate libraries and documents. I thought it'd be more elegant if this all was bundled in one place.
2. The default file format is json (simply because it's a FastAPI API), but does that mean we should have `json` again in the export endpoint? I figured it's easier to see the API simply as providing access to the docs in whatever format, and then clients can decide what to do with it (display, export, etc).
3. Finally, it was rather simple to implement multiple formats in the documents endpoint, with just two blocks that catch exports to formats other than json.

Basically, I thought it's more elegant this way and the end result works quite nicely and coherently. Now, I admit there is a wrinkle in my beautiful scheme given that we do also have an `/exporters` endpoint. I'm open to any suggestions though :).

papis/server/routes/exporters.py
jghauser: good idea!

papis/server/routes/files.py
jghauser: It would happen if someone called this function not via the http endpoint, but imported it directly in e.g. some script. This shouldn't be how it's used obviously. My LSP (or maybe our CI, can't remember) complained, so I fixed it, but if you can think of a more elegant way of handling this, let me know.

papis/server/routes/files.py
jghauser: Yeah, the idea is that the server is only responsible for ensuring paths are safe (don't contain weird characters, don't escape paths). Clients are responsible for formatting strings otherwise. This is just to keep things as simple as possible (for now). Do you think it would make sense to take a different approach?

jghauser: I just realised you made the same comment above for the documents endpoint. I'll think about how to make this align more closely with the current functionality of the commands.

papis/server/models.py
jghauser: It's not allowed to be missing. I explicitly set the `papis_id` field to make pydantic happy (which requires the `papis_id` field for `DocumentResponse`. The fallback empty string is also there because our types do not guarantee that `papis_id` exists in a document (even if it really should as you correctly say). Not sure what the best/most pythonic way of doing things here is...

papis/server/routes/documents.py
jghauser: Doesn't it just return `os.path.basename(folder)`?

alexfikl: My understanding is that that's always an absolute path (gets constructed from the library paths, which are also absolute).

jghauser: But since it just returns the last component of the path, shouldn't this get us something safe and sensible anyway? Am I missing something?

alexfikl: :facepalm: I don't know why I kept reading it as just `get_main_folder()`. Yeah, you're right, sorry for the noise.

jghauser: Haha, no worries at all!

jghauser:
New version of the server! Now comes with a `papis server` command and, as far as I know, complete coverage of Papis functionality. The endpoints also behave much more like the CLI (e.g. processing format patterns, easier adding of files, etc).

I expect the CI to fail for Windows (and maybe Mac) since I can't test it. 

Note that I'm really not an expert on much of the stuff in `papis/commands/server.py` and especially handling Windows seems to be quite difficult. I ended up adding a dependency for Windows to avoid ~50 lines just to check background process liveness on Windows...

EDIT: And this is on top of my changes in #1204 (that make `git.py` useful for the server), so that PR would need to be merged first.

jghauser:
Ha! quite proud to get so little of the CI failing :sweat_smile:. Failing sqlite tests are (I think) because of a concurrency issue, which I'll get to in  due course (together with some more general concurrency fixes for Papis needed for the API).

jghauser:
Failing sqlite tests are due to the FastAPI test harness running something in a background thread. Apparently adding `check_same_thread=False` to `sqlite3.connect` would fix that and not make Papis less thread-safe since sqlite in WAL mode should be thread-safe (and Papis is anyway already not thread-safe haha). But it does seem a bit unelegant...

All the `async def` (without `await`) are needed because that's what's telling FastAPI not to use a threadpool and instead run everything in the main event loop.

PR #1175, opened 2026-05-06
Correct capitalization of 'JabRef' in README

koppor:
Thank you for mentioning JabRef!

I took the liberty to propose our official casing. 😅

Maybe, we can think of some collaboration.


jghauser:
Thanks for the fix! :blush: And great to hear from someone else in this space. Are there any areas of collaboration you had in mind?

PR #1177, opened 2026-05-11
Use dependabot for flake.nix updates

### LINKED ISSUE #1162: switch to dependabot for upgrading nix flakes
https://github.com/dependabot/dependabot-core/issues/7340

Support recently landed, so maybe this is something to consider vs. the custom workflow that we have now.

alexfikl:
Fixes #1162


.github/actions/nix-setup/action.yml
```diff
@@ -1,25 +0,0 @@
...
-        restore-prefixes-first-match: nix-cache-
-        # purge all versions of the cache
-        purge: true
-        purge-prefixes: nix-cache-
-        # created more than 0 seconds ago relative to the start of the `Post Restore` phase
-        purge-created: 0
-        # except the version with the `primary-key`, if it exists
-        purge-primary-key: never
-        # and collect garbage in the Nix store until it reaches this size in bytes
-        gc-max-store-size: 0M
```
alexfikl, line 22: @f0rki I don't recall why this was set to 0? My understanding would be that setting it to zero never caches anything.

alexfikl:
I'll go ahead and merge this. We'll see if anything bad happens :shrug:

Issue #697, opened 2023-11-12
papis init: First-time usage is fragile and confusing


hseg:
* papis version (`$ papis --version` or commit number): 0.13

# how to reproduce the issue
Command and output of `papis --log INFO <command>`:
If `~/Documents/papers` and `$XDG_CONFIG_HOME/papis/` don't exist:
(context for below: on my system, `$XDG_CONFIG_HOME=~/.local/etc/`)
```
No sections were found in the configuration file. Adding default ones (with a default library named 'papers')!
Traceback (most recent call last):
  File "/usr/bin/papis", line 33, in <module>
    sys.exit(load_entry_point('papis==0.13', 'console_scripts', 'papis')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/click/core.py", line 1685, in invoke
    super().invoke(ctx)
  File "/usr/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/papis/commands/default.py", line 224, in run
    raise RuntimeError(
RuntimeError: Library 'papers' does not have any existing folders attached to it. Please define and create the paths in the configuration file
```

If `~/Documents/papers/` exists:
```
Creating configuration folder in '/home/gesh/.local/etc/papis'.
Creating scripts folder in '/home/gesh/.local/etc/papis/scripts'
No sections were found in the configuration file. Adding default ones (with a default library named 'papers')!
[WARNING] commands.init: Config file '/home/gesh/.local/etc/papis/config' already exists!
[...]

PR #1182, opened 2026-05-26
Only allow one `dir` per library (remove `dirs`)

jghauser:
Papis currently allows having multiple directories (`dirs`) per library. However, the feature was never documented, badly supported, and little used. This PR would remove this functionality to reduce complexity.

We'll let this sit here for a while in case there are after all people who depend on this :smile:.

### DIFF
--- a/papis/commands/default.py
+++ b/papis/commands/default.py
@@ -158,15 +158,14 @@ def run(ctx: click.Context,
     if lib is None:
         # NOTE: check if the current folder is a configured library
         libdir = os.getcwd()
-        config_lib_dirs = {
-            path: libname
+        config_lib_paths = {
+            papis.config.get_lib_from_name(libname).path: libname
             for libname in papis.config.get_libs()
-            for path in papis.config.get_lib_from_name(libname).paths
         }
 
         while libdir != (nextlibdir := os.path.dirname(libdir)):
-            if libdir in config_lib_dirs:
-                lib = config_lib_dirs[libdir]
+            if libdir in config_lib_paths:
+                lib = config_lib_paths[libdir]
                 break
 
             libdir = nextlibdir
@@ -183,21 +182,20 @@ def run(ctx: click.Context,
         raise SystemExit(1) from None
 
     configuration = papis.config.get_configuration()
-    if library.paths:
-        # Now the library should be set, let us check if there is a
-        # local configuration file there, and if there is one, then
+    if os.path.isdir(library.path):
+        # Check if there's a local configuration file in the library, and if so,
         # merge its contents
         local_config_file = papis.config.getstring("local-config-file")
-        for path in library.paths:
-            local_config_path = os.path.join(path, local_config_file)
-            papis.config.merge_configuration_from_path(local_config_path, configuration)
+        local_config_path = os.path.join(library.path, local_config_file)
+        papis.config.merge_configuration_from_path(local_config_path, configuration)
     else:
         config_file = papis.config.get_config_file()
         if os.path.exists(config_file):
             logger.error(
-                "Library '%s' does not have any folders attached to it. Please "
-                "create and add the required paths to the configuration file.",
-                library)
+                "Library '%s' directory '%s' does not exist. Please "
+                "create it or update the 'dir' setting in the "
+                "configuration file.",
+                library, library.path)
         elif ctx.invoked_subcommand != "init":
             logger.warning("No configuration file exists at '%s'.", config_file)
             logger.warning("Create a configuration file and define your "
--- a/papis/config.py
+++ b/papis/config.py
@@ -638,8 +638,7 @@ def set_lib(library: Library) -> None:
 
     if library.name not in config:
         # NOTE: can't use set(...) here due to cyclic dependencies
-        paths = [escape_interp(path) for path in library.paths]
-        config[library.name] = {"dirs": str(paths)}
+        config[library.name] = {"dir": escape_interp(library.path)}
 
     CURRENT_LIBRARY = library
 
@@ -656,14 +655,14 @@ def set_lib_from_name(libname: str) -> None:
 def get_lib_from_name(libname_or_path: str) -> Library:
     """Get a library object from a name.
 
-    :param libname: the name of a library in the configuration file or a path
-        to an existing folder that should be considered a library.
+    :param libname_or_path: the name of a library in the configuration file
+        or a path to an existing folder that should be considered a library.
     """
     config = get_configuration()
     default_settings = get_default_settings()
     libs = get_libs_from_config(config)
 
-    from papis.library import Library, from_paths
+    from papis.library import Library
 
     if libname_or_path not in libs:
         path = os.path.abspath(os.path.expanduser(libname_or_path))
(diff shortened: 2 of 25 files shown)


papis/commands/default.py
```diff
@@ -183,27 +182,12 @@ def run(ctx: click.Context,
...
-            logger.error(
-                "Library '%s' does not have any folders attached to it. Please "
-                "create and add the required paths to the configuration file.",
-                library)
-        elif ctx.invoked_subcommand != "init":
-            logger.warning("No configuration file exists at '%s'.", config_file)
-            logger.warning("Create a configuration file and define your "
-                           "libraries before using Papis. You can use "
-                           "'papis init /path/to/my/library' for a quick "
-                           "interactive setup.")
```
alexfikl, line 206: Why was this removed? Have you checked that it still works as expected with no configuration file in `papis init`?

papis/config.py
```diff
@@ -697,27 +696,47 @@ def get_lib_from_name(libname_or_path: str) -> Library:
...
 
-        lib = Library(libname, paths)
+        lib = Library(libname, path)
 
     return lib
 
 
+def get_lib_path() -> str:
+    """Get the path of the current library."""
+    return get_lib().path
```
alexfikl: I would be almost inclined to just use `get_lib().path` everywhere? What do you think?

papis/pick/__init__.py
```diff
@@ -209,8 +209,7 @@ def header_filter(lib: str) -> str:
         library = papis.config.get_lib_from_name(lib)
         return format(header_format, {
             "name": library.name,
-            "dir": library.paths[0],
-            "paths": library.paths
+            "dir": library.path,
```
alexfikl, line 212: I would leave `paths` in there, so we don't break custom user format strings? I don't imagine this is actually customized by many, if any, but still..

alexfikl (approved):
From my point of view, this is good to go :grin: Left a few nitpicks, but nothing too important.

papis/library.py
```diff
@@ -1,40 +1,17 @@
...
-            )
-
-    def path_format(self) -> str:
-        """
-        :return: a string containing all the paths in the library concatenated
-            using a colon.
-        """
-        return ":".join(self.paths)
+        #: The path with documents that form the library.
+        self.path: str = os.path.abspath(os.path.expanduser(path))
```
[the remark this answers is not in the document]
alexfikl: This is a good point. We should yell in here if it the one dir expands to more? I don't know if there are other places that should check that..

papis/config.py
```diff
@@ -697,27 +696,47 @@ def get_lib_from_name(libname_or_path: str) -> Library:
...
+
+
 def get_lib_dirs() -> list[str]:
     """Get the directories of the current library."""
-    return get_lib().paths
+    logger.warning("'papis.config.get_lib_dirs' is deprecated "
+                   "and will be removed in Papis 0.17. "
+                   "Use 'papis.config.get_lib_path' instead.")
+
+    return [get_lib().path]
```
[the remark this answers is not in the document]
alexfikl: This is a good point. All of these places should use `warnings.warn`, not the logger. These are developer facing messages.

papis/config.py
jghauser: I think that makes sense and makes things even simpler which I am always a fan of :)

papis/pick/__init__.py
jghauser: Fair enough, I'll add it back in with a note about removing it in v0.17

papis/commands/default.py
jghauser: Yeah, it seems to work just fine (tested in the container). All of this code was removed because the `Library` class now always has exactly one path (rather than a possibly empty `Sequence` of paths). Before we needed to guard against the case where we had no paths (and then guard against the case where it's ok to have no paths when running `papis init`). Now, we don't need to do that at all anymore. We always get a single path, either what's configured or the default in case of `papis init`.

alexfikl: With this PR, if I delete my configuration file and call
```
papis list -a '*'
```
it just prints nothing. Before, it gave some help to set up the configuration file and libraries (those two warnings there). I would argue that it's nice to keep those to help new users. Most of that was added to address https://github.com/papis/papis/issues/697.

jghauser: Oh, I see, thanks! I misunderstood and only looked at `papis init` :facepalm:. Will have a look again!

papis/library.py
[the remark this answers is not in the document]
jghauser: I think it might be best to add it to this to `Library`'s initializer? We'd need to slightly adjust `config.py` so that we construct a library before `if os.path.isdir(path):` on line 676, but that's very minor.

papis/commands/default.py
jghauser: Added back in!

jghauser:
I addressed the points you raised and added a check so that the `papis init` hint doesn't get forgotten again. The test is probably going to make CI fall over (but that's addressed in #1184).

papis/commands/default.py
```diff
@@ -182,29 +181,29 @@ def run(ctx: click.Context,
...
             logger.warning("No configuration file exists at '%s'.", config_file)
             logger.warning("Create a configuration file and define your "
                            "libraries before using Papis. You can use "
                            "'papis init /path/to/my/library' for a quick "
                            "interactive setup.")
 
+    configuration = papis.config.get_configuration()
+    # Now the library should be set, let us check if there is a
+    # local configuration file there, and if there is one, then
+    # merge its contents
```
alexfikl: This comment isn't exactly correct? i.e. if the warnings above fire, there's probably no library set.

papis/pick/__init__.py
```diff
@@ -209,8 +209,8 @@ def header_filter(lib: str) -> str:
...
+            "paths": library.path,
```
alexfikl:
```suggestion
            "paths": [library.path],
```

papis/library.py
```diff
@@ -1,40 +1,27 @@
...
+    def __init__(self, name: str, path: str) -> None:
         #: The name of the library, as it appears in the configuration file if
         #: defined there.
         self.name: str = name
-        #: A list of paths with documents that form the library.
-        self.paths: list[str] = list(
-            chain.from_iterable(glob.glob(os.path.expanduser(p)) for p in paths)
+        #: The path with documents that form the library.
+        self.path: str = os.path.abspath(os.path.expanduser(path))
+        if any(c in self.path for c in GLOB_CHARS):
```
alexfikl: This check feels a bit hacky. Why not just call `glob.glob` like before and error our if it returns more then one path?

alexfikl (approved):
Just a few more nitpicks, but otherwise I think it's good to go from my side.

papis/pick/__init__.py
jghauser: yup, that would be better :sweat_smile:

papis/library.py
jghauser: I can see that it's less hacky though on the other hand it wouldn't really be a check about not allowing globs, but rather about not allowing globs that resolve to multiple directories. Which I think is less neat (and not what the message says) but really I don't have a strong view on the matter haha

alexfikl: Hm.. fair enough, then I take all of that back :grin: Just let the user put whatever they want in there, and we can just yell if the folder actually doesn't exist (i.e. what you did initially..).

I see no particular reason to disallow "?" or "[]" in the paths..

papis/commands/default.py
jghauser: Do you think it makes more sense to put this stuff into into an if/else block (similar to how it was before) so the local config stuff only runs if there is a path? Or should we simply exit if the above warning or error trigger? they sound like they should terminate execution but they don't... (if we do that we should probably convert the warning into an error as well)

Though I guess the second option is kinda introducing new concerns here. New PR?

alexfikl: They "terminate execution" in the sense that no documents will be found and nothing will happen anyway, I guess :grin:

I don't think it should outright exit though, since things like `papis config --default` should maybe still work regardless?

jghauser: good point!

jghauser: I put the `if/else` back in so that merging configs is only attempted if the library path exists (which is slightly different from before when it was attempted when the `library.path` existed). But I think the new behaviour makes more sense anyway :).

alexfikl:
Code-wise this looks good to me now, no more nitpicks :grin:

jghauser:
Perfect thanks for having a look! :heart: I'll leave this here for a while to see if someone else has something to say

jghauser:
Alright, looks like no-one is going to miss this :crossed_fingers::sweat_smile:. I'll merge it now

PR #1183, opened 2026-05-26
`document.update()` should remove fields set to `None`

### LINKED ISSUE #1164: `Document` `update()` method should drop fields with value `None`
At the moment the `Document` class inherits `update()` from `dict`. This doesn't handle fields with value `None` in a special way, so that `info.yaml` can get polluted with such fields (that Papis has no use for). `papis update` handles this by deleting these fields, but it would be better to just change the `update()` method so this is handled consistently everywhere (our server will use it too for example).

jghauser:
Does what it says on the tin. We don't want fields set to `None` to be left in our `info.yaml`. This change makes behaviour consistent across the codebase so that e.g. `papis update` doesn't have to implement it separately.

fixes #1164

### DIFF
--- a/papis/document.py
+++ b/papis/document.py
@@ -408,6 +408,17 @@ class Document(dict[str, Any]):
         """
         return ""
 
+    def update(self, *args: Any, **kwargs: Any) -> None:
+        """Update the document, silently dropping keys with ``None`` values.
+
+        This overrides :meth:`dict.update` to ensure that fields with a
+        ``None`` value never leak into the document and get persisted to
+        its ``info.yaml`` file.
+        """
+        super().update(*args, **kwargs)
+        for key in [k for k in self if self[k] is None]:
+            del self[key]
+
     def copy(self) -> Document:
         """Make a shallow copy of the :class:`Document`."""
         doc = Document(data=dict(self))
--- a/tests/test_document.py
+++ b/tests/test_document.py
@@ -53,6 +53,17 @@ def test_from_folder() -> None:
     assert doc["author"] == "Russell, Bertrand"
 
 
+def test_update_drops_none_values() -> None:
+    """Document.update() must silently drop keys with None values."""
+    doc = papis.document.from_data({"title": "Hello", "author": "Turing"})
+
+    doc.update({"doi": None, "year": 1999, "abstract": None})
+    assert "doi" not in doc
+    assert "abstract" not in doc
+    assert doc["year"] == 1999
+    assert doc["title"] == "Hello"
+
+
 def test_main_features() -> None:
     doc = papis.document.from_data({
         "title": "Hello World",


papis/document.py
```diff
@@ -408,6 +408,17 @@ def __missing__(self, key: str) -> str:
...
+    def update(self, *args: Any, **kwargs: Any) -> None:
+        """Update the document, silently dropping keys with ``None`` values.
+
+        This overrides :meth:`dict.update` to ensure that fields with a
+        ``None`` value never leak into the document and get persisted to
+        its ``info.yaml`` file.
+        """
+        super().update(*args, **kwargs)
+        for key in [k for k in self if self[k] is None]:
+            del self[key]
```
alexfikl, line 420: Was this intentional? To remove all `None` keys, not just the ones in `*args, **kwargs`?

papis/document.py
```diff
@@ -408,6 +408,17 @@ def __missing__(self, key: str) -> str:
         """
         return ""
 
+    def update(self, *args: Any, **kwargs: Any) -> None:
```
alexfikl, line 411: I can still do `doc[key] = None`. Do we want to allow that?

papis/document.py
```diff
@@ -408,6 +408,17 @@ def __missing__(self, key: str) -> str:
...
+        """Update the document, silently dropping keys with ``None`` values.
+
+        This overrides :meth:`dict.update` to ensure that fields with a
+        ``None`` value never leak into the document and get persisted to
+        its ``info.yaml`` file.
+        """
+        super().update(*args, **kwargs)
+        for key in [k for k in self if self[k] is None]:
```
alexfikl, line 419: This would still allow values that are an empty string or an empty list. Is that intentional?

alexfikl (commented):
Forgot to take a look at this, sorry! Left a few comments mostly to understand what's the intention.

papis/document.py
jghauser: Yeah, I did sort of intend that since it'll nicely 'clean up' documents that might be in a messy state (and I don't see any disadvantage).

papis/document.py
jghauser: hmm would be nice if we could disallow that somehow I suppose. How would we do that?

papis/document.py
jghauser: I did intend that, yes. I don't personally see any use for it, but I figured that `None` is the only value that is guaranteed not to have any meaning. But happy to change this.

papis/document.py
alexfikl: Just implement `__setitem__` with something like (I think?)
```python
def __setitem__(self, value):
	if value is not None:
		super().__setitem__(value)
```

jghauser:
@alexfikl, sorry for merging without waiting. Should have given you a quick ping before :)

papis/document.py
alexfikl: Hm.. Yeah, I don't know? The reason I asked was because `papis update` can remove things from a list, leaving it empty. It currently checks for empty lists before writing though, so it's ok.

papis/document.py
jghauser: Nice, I'll make a quick follow up PR

papis/document.py
alexfikl: Yeah, fair enough. My thought there was
1. it's maybe expensive to go through the whole document on every update.
2. isn't it `papis doctor`'s job to clean documents?

It doesn't seem like a big deal though, so I'm fine with leaving it as is.

alexfikl:
> @alexfikl, sorry for merging without waiting. Should have given you a quick ping before :)

No worries at all! I totally forgot about this too :facepalm:

papis/document.py
jghauser: Both good points. I'll add it to other bits to do in a follow-up PR.

papis/document.py
jghauser: I'm gonna try see if what bibtex/biblatex/hayagriva do with empty strings/lists/etc. If they all treat them the same as missing fields, then I think we can just clean them out too.

jghauser:
I've been working on a follow-up PR and realised two things:
1. It does seem like empty values can be dropped. A quick look into bibtex/biblatex/hayagriva shows that they all do not distinguish between these two cases.
2. But then, if we change `__setitem__` accordingly, we can't use constructions like `doc[key] = []` and then add things to this list anymore, which is unintuitive.

I think maybe we just leave `__setitem__` alone or restrict it to dropping only fields with `None` values? Though that second option is just making things more inconsistent... I am adding a doctor check as well, which means it will be caught elsewhere. What do you think?

alexfikl:
> But then, if we change `__setitem__` accordingly, we can't use constructions like `doc[key] = []` and then add things to this list anymore, which is unintuitive.

That's fair, but we probably shouldn't be doing that anyway, haha. For example, the `papis` database backend just has an in memory dict of all the documents, so when you get one and you start modifying it in place like that it'll do some unexpected things :\

Ideally we'd make the document be a `Mapping`, not a `MutableMapping` (or full `dict`) like it is now to avoid this sort of stuff.

> I think maybe we just leave __setitem__ alone

I'm fine with leaving it alone and, in general, leaving other False-y values that aren't None alone as well.

jghauser:
> > I think maybe we just leave **setitem** alone
> 
> I'm fine with leaving it alone and, in general, leaving other False-y values that aren't None alone as well.

Do you mean also have `update()` only remove `None` values (and not empty lists, dicts, string)? I do think generally we can remove empty fields, either when we add empty values with `update()` or, alternatively, with a doctor check.

alexfikl:
> Do you mean also have `update()` only remove `None` values (and not empty lists, dicts, string)? I do think generally we can remove empty fields, either when we add empty values with `update()` or, alternatively, with a doctor check.

My main worry there is that doing something like `papis update --set key value` on a document with `None` or empty values will sneakily clean up the document for me without asking (and modify other keys besides the ones I give it on the command-line). 

It doesn't seem necessarily bad, but that's why I'm more like 80% into this and not quite sure how many "empty" values to clean up :\ Maybe we can stick to just cleaning up `None` for now and see if other values become annoying in the future? I don't feel very strongly about it though..

A doctor check that removes empty fields is definitely a good thing to have though!

EDIT: Actually `papis update` already removes non-command-line keys even without this :facepalm:, so yeah.. maybe it's fine to just clean it up, haha!

jghauser:
Sorry, I think I didn't explain well. Basically I'm thinking of doing:
1. `update()` will drop any fields where a value is empty. Fields of any type are considered empty when the value is `None` and lists/dicts/strings when they have no content. We leave the number `0` or `False` booleans alone. so `update({files = []})` will drop `files`. It will (unlike this PR) leave pre-existing empty fields alone.
2. `__setitem__` will not be changed from default.
3. A new `empty-fields` check will remove any fields that are empty.

Does that make sense?

EDIT: And it would then make sense to not have `papis update` not clean up pre-existing fields, but rather rely on doctor check (and `auto-doctor`) to do that. That would more neatly separate concerns.

alexfikl:
> Does that make sense?

Yes! I definitely misunderstood what the plan is, sorry! I'm on board with that :grin:

PR #1185, opened 2026-05-29
refactor: move non-interactive new doc creation code to `document.py`

jghauser:
This factors out all the non-interactive parts about new document creation from `add.py`. The idea is that this code can then be used elsewhere (as in the API).

Functionality should be exactly as before, except in one message: when the user is asked to confirm adding files to the document, we can no longer tell them to what the file will be renamed. Unfortunately, this requires the various fields of the document to be present and we do not have them at this stage. I don't think this matters all that much (especially since they will be shown the final name when confirming adding the document at the end). We could add a note saying that the file will be renamed if you think it's currently confusing.

I decided against changing the place of the hook and duplication check.

I will add more tests if/when we're happy with how this works.

### DIFF
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -198,37 +198,52 @@ def run(paths: list[str],
         if not os.path.exists(p):
             raise FileNotFoundError(f"File '{p}' not found")
 
-    import tempfile
-
     in_document_paths = paths
-    temp_dir = tempfile.mkdtemp()
 
     from papis.database import get as get_database
 
     db = get_database()
 
-    from papis.document import Document, describe, dump
+    from papis.document import describe, dump
+    from papis.tui.utils import confirm as ask_confirm, text_area
+    from papis.utils import open_file as open_file_viewer
 
-    tmp_document = Document(folder=temp_dir, data=data)
-    db.maybe_compute_id(tmp_document)
+    approved_files: list[str]
+    if not batch:
+        approved_files = []
+        for in_file_path in in_document_paths:
+            if open_file:
+                open_file_viewer(in_file_path)
 
-    # reference building
-    # NOTE: this needs to go before any papis.format calls, so that those can
-    # potentially use the 'ref' key in the format patterns.
-    if "ref" not in data:
-        from papis.bibtex import create_reference
+            if confirm and not ask_confirm(
+                    f"Add file '{os.path.basename(in_file_path)}' to document?"):
+                continue
 
-        new_ref = create_reference(data)
-        if new_ref:
-            logger.info("Created reference '%s'.", new_ref)
-            tmp_document["ref"] = new_ref
+            approved_files.append(in_file_path)
+    else:
+        approved_files = list(in_document_paths)
 
-    if auto_doctor:
-        from papis.commands.doctor import fix_errors
+    from papis.document import new as new_document
 
-        logger.info("Running doctor auto-fixers on document: '%s'.",
-                    describe(tmp_document))
-        fix_errors(tmp_document)
+    tmp_document = new_document(
+        data,
+        approved_files,
+        mode="link" if link else "copy",
+        file_name_format=file_name,
+        auto_doctor=auto_doctor)
+
+    # Log per-file operations
+    if link:
+        verb = "Linking"
+    elif move:
+        verb = "Moving"
+    else:
+        verb = "Copying"
+
+    for in_file, out_name in zip(
+            approved_files, tmp_document["files"], strict=True):
+        logger.info(
+            "%s '%s' to '%s'.", verb, os.path.basename(in_file), out_name)
 
     # create a nice folder name for the new document
     if base_path is None:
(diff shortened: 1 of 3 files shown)


papis/document.py
```diff
@@ -697,33 +697,93 @@ def document_sort_key(doc: Document) -> tuple[int, datetime, int, str]:
...
+        if os.path.exists(out_file_path):
+            logger.warning(
+                "File '%s' already exists. Skipping...", out_file_path)
+            continue
+
+        if link:
+            logger.info("[LN] '%s' to '%s'.", in_file_path, out_file_name)
+            symlink(in_file_path, out_file_path)
+        else:
+            action = "[MV]" if move_files else "[CP]"
```
alexfikl: I'm not sure this is any clearer than what was there before, but sure. Could remove this if the `link` and `move_files` flags are merged.

papis/document.py
```diff
@@ -697,33 +697,93 @@ def document_sort_key(doc: Document) -> tuple[int, datetime, int, str]:
...
-    :param folder_path: a main folder for the document.
     :param data: a :class:`dict` with key and values to be used as metadata
         in the document.
-    :param files: a sequence of files to add to the document.
-    :raises FileExistsError: if *folder_path* already exists.
+    :param files: a sequence of file paths to add to the document.
+    :param link: if *True*, create symbolic links to the files instead of
+        copying them.
+    :param move_files: if *True*, copy the files so that the originals can be
+        removed by the caller (the files are copied, not actually moved).
```
alexfikl: Sounds like you want to rename this to "copy_files"? Also, it sounds like the option is "link or copy", so there's no point in having two flags?

papis/document.py
```diff
@@ -697,33 +697,93 @@ def document_sort_key(doc: Document) -> tuple[int, datetime, int, str]:
...
+    :param file_name_format: a format pattern used to construct new file names
+        from the document data (see :confval:`add-file-name`).
```
alexfikl:
```suggestion
        from the document data (defaults to :confval:`add-file-name`).
```
?

alexfikl (commented):
Left a few nitpicks, but overall looks good to me.

> when the user is asked to confirm adding files to the document, we can no longer tell them to what the file will be renamed.

Yeah, I think that's fine too, as long as we still ask them what files to add in interactive mode. I recall adding the "renamed to.." part just as a small check that "add-file-name" worked as expected.

papis/document.py
```diff
@@ -697,33 +697,93 @@ def document_sort_key(doc: Document) -> tuple[int, datetime, int, str]:
...
-def new(folder_path: str,
+def new(
         data: dict[str, Any],
-        files: Sequence[str] | None = None) -> Document:
-    """Creates a complete document with data and existing files.
+        files: Sequence[str] = (), *,
+        link: bool = False,
+        move_files: bool = False,
+        file_name_format: AnyString | None = None,
+        auto_doctor: bool = False) -> Document:
```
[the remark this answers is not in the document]
alexfikl: More verbose, but yeah, this is what I meant below too :grin:

papis/document.py
jghauser: Yeah, removed it since you (and Copilot) are absolutely right about `move_files` being a lie. New version has a mode with two options: copy and link.

jghauser: I removed all this logging from `documents.py` and added it to `add.py`. It's only the cli command that can distinguish between files being copied or moved.

jghauser:
Thanks for taking a look! The new version should address your points. (I also changed the MV/CP/LN to spelling things out in the user facing messages. I think it's friendlier :blush:)

papis/commands/add.py
```diff
@@ -198,87 +198,55 @@ def run(paths: list[str],
...
-            zip(in_document_paths, renamed_file_list, strict=True)):
-        out_file_path = os.path.join(temp_dir, out_file_name)
-        if os.path.exists(out_file_path):
-            logger.warning("File '%s' already exists. Skipping...", out_file_path)
-            continue
-
+    approved_files: list[str] = []
+    for in_file_path in in_document_paths:
         if not batch and open_file:
             open_file_viewer(in_file_path)
```
alexfikl: This whole loop should be in a "if not batch.. else: approved_files = in_document_paths"?

papis/document.py
```diff
@@ -708,33 +698,87 @@ def document_sort_key(doc: Document) -> tuple[int, datetime, int, str]:
...
+        file_name_format=file_name_format, allow_remote=False)
+
+    # Copy or link files into the temporary directory
+    document_file_list: list[str] = []
+    for in_file_path, out_file_name in zip(files, renamed_files, strict=True):
+        out_file_path = os.path.join(temp_dir, out_file_name)
+        if os.path.exists(out_file_path):
+            logger.warning(
+                "File '%s' already exists. Skipping...", out_file_path)
+            continue
```
alexfikl: This really shouldn't happen anymore, right? i.e. `rename_document_files` ensures they're nicely unique and whatnot.

papis/document.py
```diff
@@ -708,33 +698,87 @@ def document_sort_key(doc: Document) -> tuple[int, datetime, int, str]:
     return sorted(docs, key=document_sort_key, reverse=reverse)
 
 
-def new(folder_path: str,
+def new(
         data: dict[str, Any],
```
alexfikl:
```suggestion
        data: DocumentLike,
```
? i.e. if you want to create a new document with the metadata from an existing one, just go ahead!

papis/commands/add.py
```diff
@@ -198,87 +198,55 @@ def run(paths: list[str],
...
+    # Log per-file operations
+    if link:
+        verb = "Linking"
+    elif move:
+        verb = "Moving"
+    else:
+        verb = "Copying"
+
+    for in_file, out_name in zip(
+            approved_files, tmp_document["files"], strict=True):
```
alexfikl, line 244: Are these guaranteed to be the same length? There's an `if ...: continue` in `papis.document.new`, but I don't think that does anything?

alexfikl (commented):
Left a few questions mostly, but overall this looks good to me!

papis/commands/add.py
jghauser: Good catch!

papis/document.py
jghauser: Yeah, you're totally right. That then also addresses the issue with `zip` below.

alexfikl (approved):
Nice! Looks good to me :grin:

jghauser:
Thanks! In it goes :blush:

PR #1188, opened 2026-06-10
Document.update(): drop None values to keep info.yaml clean

### LINKED ISSUE #1164: `Document` `update()` method should drop fields with value `None`

wali-reheman:
Fixes #1164.

Calling `document.update(data)` would write None values into the document, polluting info.yaml. This overrides update() to drop Nones upfront so both the CLI path and direct API callers benefit.

### DIFF
--- a/papis/document.py
+++ b/papis/document.py
@@ -408,6 +408,28 @@ class Document(dict[str, Any]):
         """
         return ""
 
+    def update(self, *args: Any, **kwargs: Any) -> None:
+        """Override dict.update to drop None values.
+        
+        This ensures that calling document.update(data) does not introduce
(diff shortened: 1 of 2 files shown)


alexfikl:
https://github.com/papis/papis/pull/1183 already does this, I'm afraid. If your version does anything that you feel is missing from the other PR, feel free to leave a comment there.

PR #1192, opened 2026-06-18
Feat/document update drop values v2

jghauser:
Follow-up to #1164, implementing the features discussed there.

1. `update()` drops empty fields, but leaves pre-existing fields alone
2. `empty-field` doctor check removes all empty fields
3. `empty-field` check is added to default checks (though happy to change that)

It seems bibtex/biblatex/hayagriva all handle empty fields the same as missing ones, which is why we shouldn't have them in our `info.yaml` files. Here is some evidence for this:
- [CSL suppresses all groups that are empty](https://docs.citationstyles.org/en/stable/specification.html#group).
- [BibTeX's default style template](http://mirrors.ctan.org/biblio/bibtex/base/btxbst.doc) explicitly says missing and empty values are treated the same and [BibLaTeX's punctuation tracker serves to suppress output for missing/empty fields](https://tex.stackexchange.com/questions/405173/biblatex-punctuation-and-empty-field)

### DIFF
--- a/papis/document.py
+++ b/papis/document.py
@@ -370,6 +370,20 @@ class DocHtmlEscaped(dict[str, Any]):
             .replace('"', "&quot;"))
 
 
+def is_empty_value(value: Any) -> bool:
+    """Check if a value is empty and should be treated as missing.
+
+    Returns *True* for ``None``, empty strings, empty lists, and empty
+    dicts. Numeric values (including ``0``) and booleans are never
+    considered empty.
+    """
+    if value is None:
+        return True
+    if isinstance(value, (int, float, bool)):
+        return False
+    return not value
+
+
 class Document(dict[str, Any]):
     """An abstract document in a ``papis`` library.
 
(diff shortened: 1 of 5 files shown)


papis/document.py
```diff
@@ -410,15 +424,29 @@ def __missing__(self, key: str) -> str:
...
+        never leak into the document and get persisted to its
+        ``info.yaml`` file.
         """
-        super().update(*args, **kwargs)
-        for key in [k for k in self if self[k] is None]:
-            del self[key]
+        incoming: dict[str, Any] = {}
+        for arg in args:
+            incoming.update(arg)
+        incoming.update(kwargs)
```
alexfikl:
```suggestion
        incoming = dict(*args, **kwargs)
```
? I think that does the trick as well, no?

papis/document.py
```diff
@@ -410,15 +424,29 @@ def __missing__(self, key: str) -> str:
...
+
+        filtered: dict[str, Any] = {}
+        for key, value in incoming.items():
+            if is_empty_value(value):
+                self.pop(key, None)
+            else:
+                filtered[key] = value
+
+        super().update(filtered)
```
alexfikl, line 445:
```suggestion
        super().update({k: v for k, v in incoming.items() if not is_empty_key(value)})
```
?

alexfikl:
Left two little nitpicks, but overall looks good to me! Feel free to merge it when you fix the CI and whatnot :grin:

papis/document.py
jghauser: thanks, that is indeed a lot nicer!

papis/document.py
jghauser: Yup and in a lot fewer lines :)

papis/document.py
jghauser: On second thought, it doesn't quite do the trick because empty values aren't dropped anymore? e.g. doing `update({"title" = ""})` won't remove the title?

alexfikl: I'm not sure I follow? It should be doing something like this
```python
>>> incoming = {"title": ""}
>>> {k: v for k, v in incoming.items() if v}
{}
```

alexfikl: Oooh, you did `self.pop`! Nevermind, nevermind!

jghauser: Exactly, it would have left the original value

jghauser:
I'll go ahead and merge this. Thanks for all the help!

PR #1195, opened 2026-07-01
Fix database cache to key by library name instead of `Library` identity

jghauser:
This is an issue I discovered while figuring out why the CI kept failing. I had a hacky workaround there that made Windows specifically fail. This here I think is the proper fix.

Basically, `papis.database.get_database` has a `DATABASES` cache keyed by `Library` objects. Since `Library` uses identity-based hashing and `get_lib_from_name()` creates a fresh `Library` object on every call, two lookups for the same library can land in different cache slots and produce separate `Database` instances.

When we switch back and forth between libraries (because of cross library document moving) we would get new `Library` objects. For same-library moves this meant `delete` mutated one instance while `add` mutated another. The fix is to index `DATABASES` by `library.name` (string) rather than `Library` (object). Now the same library always hits the same cache entry regardless of how the `Library` object was obtained.

### DIFF
--- a/papis/database/__init__.py
+++ b/papis/database/__init__.py
@@ -11,7 +11,7 @@ if TYPE_CHECKING:
 
 logger = papis.logging.get_logger(__name__)
 
-DATABASES: dict[Library, Database] = {}
+DATABASES: dict[str, Database] = {}
 
 
 def _instantiate_database(backend_name: str,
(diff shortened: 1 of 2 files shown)


alexfikl (approved):
Oh.. my.. God 🤦‍♂️

PR #1199, opened 2026-07-02
refactor: make `paths._make_unique_file` public

jghauser:
I use this in various places in the new FastAPI server. And it's already used in the downloaders, so should be public anyway.


alexfikl (approved):
Sounds good to me. Maybe add a little docstring for it :grin:

jghauser:
Added a docstring. Gonna merge it now! :)

PR #1200, opened 2026-07-06
Refactor/focus doctor cli code

jghauser:
This moves all non-CLI stuff from `papis/commands/doctor.py` into its own `papis/doctor.py`. Papis doctor has grown up quite a bit and now is being used in various places besides the `papis doctor` command. :blush:

The second commit moves the `suggestion_cmd` back into the CLI file since it only properly belongs to that.


jghauser:
As far as I can tell, these cyclical imports are all false alarms (they all concern deferred imports that break the cycle).

papis/commands/doctor.py
```diff
@@ -127,1097 +127,50 @@ def my_custom_check(doc) -> List[Error]:
...
-                 msg=msg,
-                 suggestion_cmd=suggestion_cmd,
-                 fix_action=fix_action,
-                 payload=payload,
-                 doc=doc)
+def _suggestion_cmd(error: Error) -> str:
+    """Build the CLI suggestion string for a given error."""
+    if error.fix_action is not None:
+        return f"papis doctor --fix -t {error.name} --doc-folder {error.path!r}"
+    return f"papis edit --doc-folder {error.path!r}"
```
alexfikl, line 155: > The second commit moves the suggestion_cmd back into the CLI file since it only properly belongs to that.

I'm not sure I appreciate these being hardcoded here. The "suggested way to fix" for an error should be decided when the error is caught.

I'm guessing you mean it's "properly belongs to the CLI" because the suggestions are CLI-centric? That sounds like a shortcoming in the current implementation, since having a fixing suggestion sounds like a good idea..

alexfikl (approved):
One little nitpick, but otherwise this looks good to me! (I'm guessing things were just moved, so I haven't looked super seriously).

> As far as I can tell, these cyclical imports are all false alarms (they all concern deferred imports that break the cycle).

Yeah, those are just noise. They weren't showing up before because the CodeQL thing already had them in its baseline from the first time they showed up.

papis/commands/doctor.py
jghauser: The way I saw it is that each check already has a fixer which can be called programmatically. How it is called manually depends on the interface, and the code here handles the CLI interface. It pretty much just translates into CLI 'run the associated fixer' (or 'edit and fix it yourself' if there is no fixer). If, for instance, we did a similar thing for a TUI (or in the API), the instructions have to be different. Do you have another way you'd like to see this done?

alexfikl: Well, this is the status quo as well, so I guess it's not a big deal for this PR. The doctor command could really use some love to be a slightly better linter :(

In a more general view, I agree that the instructions should be different based on the interface, but
1. This part of the code has no way of knowing what a good fix is, only the code that emitted the error can know. This "we can fix it or you can fix it yourself" message is not super helpful (outside of showing me some commands I can run).
2. Overloading "fix_action" to also tell us what the suggestion is is hacky.

Ideally, we'd have some sort of semantic `SuggestionType` with some metadata so the interface can decide what to do with it. E.g. `RequiresManualEdit`: CLI just prints `run 'papis edit'`, TUI asks and opens the file, a web UI makes an "Open" button not gray, etc.

alexfikl: Anyways, I forgot that the current code does this too, so nevermind! :facepalm:

jghauser: At the moment the logic is pretty simple since the suggestion is either run the fixer or you're on your own. But it'd probably be nice if we could be a bit more specific in the second case :sweat_smile:

jghauser:
Thanks for taking a look. I'll merge this now!

PR #1203, opened 2026-07-07
fix: skip local paths in DOIImporter.match

### LINKED ISSUE #1201: `papis add` crashes with `http.client.InvalidURL` when a file path contains spaces (regression of #333)
* papis version (`$ papis --version` or commit number): 0.15 (also affects current `main`, see analysis below); python-doi 0.2.0

# how to reproduce the issue

Add any local file whose path contains a space:

```
$ echo dummy > "some presentation.txt"
$ papis --log INFO add "some presentation.txt" --batch --set title Test
```

```
[INFO] importer.bibtex: Reading input file or string: 'some presentation.txt'.
[ERROR] importer.bibtex: Error reading BibTeX file or string: 'some presentation.txt'. (...)
  File ".../bin/papis", line 6, in <module>
    sys.exit(run())
  ...
  File ".../site-packages/papis/commands/add.py", line 509, in <genexpr>
    get_matching_importers_by_uri(f, include_downloaders=True)
  File ".../site-packages/papis/importer/__init__.py", line 269, in match
    importer = cls.match(uri)
  File ".../site-packages/papis/importer/doi.py", line 23, in match
    validate_doi(uri)
  File ".../site-packages/doi/__init__.py", line 55, in validate_doi
    result = json.loads(urllib.request.urlopen(request).read().decode())
  ...
  File "/usr/lib64/python3.13/http/client.py", line 1312, in _validate_path
    raise InvalidURL(f"URL can't contain control characters. {url!r} "
http.client.InvalidURL: URL can't contain control characters. '/api/handles/some presentation.txt' (found at least ' ')
```

# analysis

This is the same failure mode as #333, which was closed as fixed by #485 — but #485 replaced `urllib` with `requests` only inside papis itself. `DOIImporter.match()` delegates to `doi.validate_doi()` from the external [python-doi](https://github.com/papis/python-doi) package, which still uses `urllib` and only converts `HTTPError` into `ValueError`:

```python
# python-doi, src/doi/__init__.py (master)
try:
    result = json.loads(urllib.request.urlopen(request).read().decode())
except HTTPError:
    raise ValueError('HTTP 404: DOI not found')
```

`DOIImporter.match()` in turn only catches `ValueError`:

```python
# papis/importer/doi.py (main)
try:
    validate_doi(uri)
except ValueError:
    return None
```

For a space-free path the doi.org handle lookup just 404s → `ValueError` → importer skipped gracefully, which is why this only bites on paths with spaces (or other characters illegal in a URL path): `http.client` raises `InvalidURL` before the request is even sent, nothing catches it, and `papis add` dies.

(I'd have filed this against python-doi, but that repo has issues disabled.)

# possible fixes

- In `DOIImporter.match()`, skip URIs that are existing local paths, the way `DOIFromPDFImporter.match()` already does (`os.path.exists(uri)`). This has the nice side effect of not sending every local file path handed to `papis add` to doi.org.
- And/or broaden the `except` in `DOIImporter.match()` to also catch `http.client.InvalidURL` (or any `Exception`) so no future oddity in `validate_doi` can crash the whole `add` command.
- Longer-term, fix `validate_doi` in python-doi to URL-quote/pre-validate its input.

Happy to send a PR for the papis-side guard if that helps.

nasteffe:
A local file path handed to `papis add` is never a DOI, but
`DOIImporter.match` still sent every path to doi.org for validation.
For paths containing spaces (or other characters invalid in a URL),
`validate_doi` in python-doi raises an uncaught
`http.client.InvalidURL` and crashes the whole command.

Skip existing local paths before calling `validate_doi`, the same way
`DOIFromPDFImporter.match` already does. This also avoids a needless
network round-trip for every local file added.

Fixes #1201

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/papis/importer/doi.py
+++ b/papis/importer/doi.py
@@ -16,8 +16,8 @@ class DOIImporter(Importer):
     @classmethod
     def match(cls, uri: str) -> DOIImporter | None:
         # NOTE: a local path is never a DOI; skipping it avoids querying
-        # doi.org for every file handed to `papis add` (and paths with spaces
-        # make for an invalid URL that crashes `validate_doi` -- see #1201)
+        # doi.org (as `validate_doi` does) for every local file
+        # https://github.com/papis/papis/issues/1201
         if os.path.exists(uri):
             return None
 
--- a/papis/importer/doi.py
+++ b/papis/importer/doi.py
@@ -15,6 +15,12 @@ class DOIImporter(Importer):
 
     @classmethod
     def match(cls, uri: str) -> DOIImporter | None:
+        # NOTE: a local path is never a DOI; skipping it avoids querying
+        # doi.org for every file handed to `papis add` (and paths with spaces
+        # make for an invalid URL that crashes `validate_doi` -- see #1201)
+        if os.path.exists(uri):
+            return None
+
         from doi import validate_doi
 
         from papis.crossref import DOI_ORG_URL
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,12 @@ user can abort, skip problematic moves, or fix collisions interactively. A new
 - Remove undocumented `dirs` option (only allow one `dir` per library).
   ([#1182](https://github.com/papis/papis/pull/1182)).
 
+## Bug Fixes
+
+- Fix crash in `papis add` when a given file path contains spaces (the DOI
+  importer no longer tries to resolve local paths on doi.org)
+  ([#1201](https://github.com/papis/papis/issues/1201)).
+
 # VERSION v0.15.0 (February 8th, 2026)
 
 ## Dependency Changes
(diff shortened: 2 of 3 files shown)


papis/importer/doi.py
```diff
@@ -15,6 +15,12 @@ def __init__(self, uri: str, doi: str) -> None:
 
     @classmethod
     def match(cls, uri: str) -> DOIImporter | None:
+        # NOTE: a local path is never a DOI; skipping it avoids querying
+        # doi.org for every file handed to `papis add` (and paths with spaces
+        # make for an invalid URL that crashes `validate_doi` -- see #1201)
```
alexfikl:
```suggestion
        # NOTE: a local path is never a DOI; skipping it avoids querying
        # doi.org (as `validate_doi` does) for every local file
        # https://github.com/papis/papis/issues/1201
```
? This is also used in `papis update`, so better make the comment a bit more generic.

alexfikl (approved):
One little nitpick, but otherwise this looks good to me. Thanks for writing it up!

> Co-Authored-By: Claude Fable 5 [noreply@anthropic.com](mailto:noreply@anthropic.com)

Very fancy! The latest and the greatest at work :grin:

alexfikl:
In it goes :rocket:

PR #1204, opened 2026-07-07
Better subprocess error handling (in git and beyond)

### LINKED ISSUE #1191: Improve handling of git stdout/stderr
Git operations use `subprocess.run(...)`, but currently we don't nicely handle `git`'s output. We should especially package errors/warnings in a nice way (just like Papis internal warnings and errors). This will make these messages stand out correctly to users.

jghauser:
I was adding git functionality to the API and realised that the current `papis/git.py` functions aren't really useful (mostly because they silence errors). I then realised that in order to fix that, I have to fix the `run()` function in `utils.py`, and once that allowed us to catch stdout/stderr, there were the various call sites that needed fixing. Which is to say, this got a little out of hand. I tried to put it all into more or less coherent commits, but the PR as such touches on quite a few things.

Main points:
 - Uniform git pattern: every mutation does filesystem + database first, then attempts git tracking. Failure just means uncommitted. This means `git.py` is now index-only: `rm_cached` replaces `remove`. `mv`/`mv_and_commit` removed. Here we rely on rm+add = mv.
 - Centralized error logging: `_run_git` catches, logs git stderr through papis, re-raises. Call sites just pass.
 - `--batch` for explore cmd: Once we had error handling for `subprocess.run()`, it just made sense to add this.
 - `is_git_repo` lifted from `init.py` with optional root boundary (for server/API safety).

One question: Currently, any failures in git operations just pass, which means Papis continues, but the git state is wrong. Should we do that differently? There's two other options:
1. Raise errors. This might be unexpected for `--batch` operations and maybe too strict given that git failures don't leave the db in a bad state
2. Pass `batch` to all command `run()` functions and distinguish batch from no-batch. Maybe not worth the hassle

### DIFF
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -334,8 +334,8 @@ def run(paths: list[str],
     db.add(tmp_document)
 
     if git:
-        from papis.git import add_and_commit_resource
-        add_and_commit_resource(
+        from papis.git import add_and_commit
+        add_and_commit(
             out_folder_path, ".",
             f"Add document '{describe(tmp_document)}'")
 
--- a/papis/commands/add.py
+++ b/papis/commands/add.py
@@ -334,10 +334,14 @@ def run(paths: list[str],
     db.add(tmp_document)
 
     if git:
-        from papis.git import add_and_commit
-        add_and_commit(
-            out_folder_path, ".",
-            f"Add document '{describe(tmp_document)}'")
+        from papis.git import GitError, add_and_commit as git_add_and_commit
+
+        try:
+            git_add_and_commit(
+                out_folder_path, ".",
+                f"Add document '{describe(tmp_document)}'")
+        except GitError as exc:
+            logger.error("%s", exc)
 
     if move:
         for in_file_path in approved_files:
--- a/papis/commands/explore.py
+++ b/papis/commands/explore.py
@@ -176,10 +176,13 @@ def add(ctx: click.Context) -> None:
 
 
 @cli.command("cmd")
+@papis.cli.bool_flag(
+    "--batch",
+    help="Skip documents containing errors rather than aborting.")
 @click.pass_context
 @click.help_option("-h", "--help")
 @click.argument("command", type=papis.cli.FormatPatternParamType())
-def cmd(ctx: click.Context, command: str) -> None:
+def cmd(ctx: click.Context, command: str, batch: bool) -> None:
     """
     Run a general command on the document list.
 
@@ -203,10 +206,13 @@ def cmd(ctx: click.Context, command: str) -> None:
         try:
             run(shlex.split(fcommand))
         except subprocess.CalledProcessError as exc:
-            logger.error(
-                "Command failed for '%s': exited with code %d.",
-                fcommand, exc.returncode)
-            raise
+            msg = "Command failed for `%s`: exited with code %d."
+            if batch:
+                logger.warning(msg, fcommand, exc.returncode)
+                continue
+            else:
+                logger.error(msg, fcommand, exc.returncode)
+                raise
 
 
 @cli.command("export")
--- a/papis/commands/explore.py
+++ b/papis/commands/explore.py
@@ -98,6 +98,7 @@ Command-line interface
 from __future__ import annotations
 
 import shlex
+import subprocess
 
 import click
 
@@ -199,7 +200,13 @@ def cmd(ctx: click.Context, command: str) -> None:
     docs = ctx.obj["documents"]
     for doc in docs:
         fcommand = format(command, doc, default="")
-        run(shlex.split(fcommand))
+        try:
+            run(shlex.split(fcommand))
+        except subprocess.CalledProcessError as exc:
+            logger.error(
+                "Command failed for '%s': exited with code %d.",
+                fcommand, exc.returncode)
+            raise
 
 
 @cli.command("export")
--- a/papis/commands/external.py
+++ b/papis/commands/external.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import os
 import re
+import subprocess
 from typing import Any
 
 import click
@@ -84,4 +85,9 @@ def external_cli(ctx: click.core.Context, query: str, flags: list[str]) -> None:
 
     from papis.utils import run
 
-    run(cmd, env=environ)
+    try:
+        run(cmd, env=environ)
+    except subprocess.CalledProcessError as exc:
+        logger.error("External script '%s' exited with code %d.",
+                     cmd[0], exc.returncode)
+        ctx.exit(exc.returncode)
--- a/papis/commands/mv.py
+++ b/papis/commands/mv.py
@@ -254,30 +254,9 @@ def run(
     papis.config.set_lib_from_name(source_library)
 
     if target_lib_dir == source_lib_dir:
-        if git:
-            papis.git.mv_and_commit_resource(
-                str(main_folder_old),
-                str(main_folder_new),
-                f"Move '{papis.document.describe(document)}'",
-            )
-        else:
-            shutil.move(main_folder_old, main_folder_new)
+        shutil.move(main_folder_old, main_folder_new)
     else:
         shutil.move(doc_folder_old, main_folder_new)
-
-        if git:
-            papis.git.remove(str(source_lib_dir), str(main_folder_old))
-            papis.git.commit(
-                str(source_lib_dir),
-                f"Remove '{papis.document.describe(document)}' "
-                f"(moved to '{target_library}' library)",
-            )
-
-            papis.git.add_and_commit_resource(
-                str(target_lib_dir),
-                str(main_folder_new),
-                f"Add '{papis.document.describe(document)}'",
-            )
 
     source_db = papis.database.get()
     source_db.delete(document)
@@ -287,6 +266,32 @@ def run(
     document.set_folder(str(main_folder_new))
 
     target_db.add(document)
+
+    if git:
+        if target_lib_dir == source_lib_dir:
+            papis.git.rm_cached(str(source_lib_dir),
+                                str(main_folder_old),
+                                recursive=True)
+            papis.git.add(str(source_lib_dir), str(main_folder_new))
+            papis.git.commit(
+                str(source_lib_dir),
+                f"Move '{papis.document.describe(document)}'",
+            )
+        else:
+            papis.git.rm_cached(str(source_lib_dir),
+                                str(main_folder_old),
+                                recursive=True)
+            papis.git.commit(
+                str(source_lib_dir),
+                f"Remove '{papis.document.describe(document)}' "
+                f"(moved to '{target_library}' library)",
+            )
+
+            papis.git.add_and_commit(
+                str(target_lib_dir),
+                str(main_folder_new),
+                f"Add '{papis.document.describe(document)}'",
+            )
 
     logger.info(
         "Moved document '%s' to '%s' (library '%s').",
--- a/papis/commands/mv.py
+++ b/papis/commands/mv.py
@@ -98,7 +98,6 @@ import papis.cli
 import papis.config
 import papis.database
 import papis.document
-import papis.git
 import papis.logging
 import papis.paths
 import papis.tui.utils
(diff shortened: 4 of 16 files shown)


papis/commands/add.py
```diff
@@ -334,10 +335,13 @@ def run(paths: list[str],
...
-        add_and_commit_resource(
-            out_folder_path, ".",
-            f"Add document '{describe(tmp_document)}'")
+        try:
+            from papis.git import add_and_commit
+            add_and_commit(
+                out_folder_path, ".",
+                f"Add document '{describe(tmp_document)}'")
+        except subprocess.CalledProcessError:
+            pass
```
alexfikl: Seeing like a lot of these do a `try pass`, it feels like the API is wrong? You can either
1. Not throw at all on error and just do a `logger.error` in `git.py` like now.
2. Throw a helpful exception and let the caller do the `logger.error` or whatever they want (I usually prefer this).

papis/commands/explore.py
```diff
@@ -199,7 +203,18 @@ def cmd(ctx: click.Context, command: str) -> None:
...
+        except subprocess.CalledProcessError as exc:
+            if batch:
+                logger.warning(
+                    "Command failed for '%s': exited with code %d.",
+                    fcommand, exc.returncode)
+                continue
+            logger.error(
+                "Command failed for '%s': exited with code %d.",
+                fcommand, exc.returncode)
+            raise
```
alexfikl:
```suggestion
            else:
                logger.error(
                    "Command failed for '%s': exited with code %d.",
                    fcommand, exc.returncode)
                raise
```
?

papis/commands/external.py
```diff
@@ -84,4 +85,9 @@ def external_cli(ctx: click.core.Context, query: str, flags: list[str]) -> None:
 
     from papis.utils import run
 
-    run(cmd, env=environ)
+    try:
+        run(cmd, env=environ)
+    except subprocess.CalledProcessError as exc:
+        logger.error("External script '%s' exited with code %d.",
+                     cmd[0], exc.returncode)
+        raise SystemExit(exc.returncode) from exc
```
alexfikl:
```suggestion
	ctx.exit(exc.returncode)
```
? Not sure what the proper `click` way of exiting is.

papis/commands/mv.py
```diff
@@ -288,6 +268,35 @@ def run(
 
     target_db.add(document)
 
+    if git:
+        try:
+            if target_lib_dir == source_lib_dir:
+                papis.git.rm_cached(str(source_lib_dir),
+                                    str(main_folder_old),
+                                    recursive=True)
```
alexfikl: Remove seems to be common?

papis/commands/explore.py
```diff
@@ -199,7 +203,18 @@ def cmd(ctx: click.Context, command: str) -> None:
     docs = ctx.obj["documents"]
     for doc in docs:
         fcommand = format(command, doc, default="")
-        run(shlex.split(fcommand))
+        try:
+            run(shlex.split(fcommand))
+        except subprocess.CalledProcessError as exc:
+            if batch:
+                logger.warning(
+                    "Command failed for '%s': exited with code %d.",
```
alexfikl: `'%s'` is going to look weird if the command contains quoted strings too? Not quite sure how to show that..

papis/git.py
```diff
@@ -13,101 +21,121 @@
...
-
-    :param path: a folder with an existing git repository.
-    :param resource: a resource (e.g. ``info.yaml`` file) to add to the index.
+    try:
+        run(cmd, cwd=cwd, capture_output=True)
+    except subprocess.CalledProcessError as exc:
+        logger.error(
+            "%s:%s %s",
+            error_msg,
+            os.linesep,
```
alexfikl: Pretty sure you don't need `os.linesep` here? That looks very weird..

papis/git.py
```diff
@@ -13,101 +21,121 @@
...
 
     :param path: a folder with an existing git repository.
-    :param resource: a resource (e.g. ``info.yaml`` file) to remove from git.
     :param message: a commit message.
     """
-    add(path, resource)
-    commit(path, message)
+    logger.info("[GIT] Committing in '%s': '%s'.", path, message)
+    _run_git(["git", "commit", "-m", message], str(path),
+             f"[GIT] Failed to commit in '{path}'")
```
alexfikl: Any chance we can make this
```suggestion
    git("commit", "-m", message, str(path))
```
? That would look cool :grin:

papis/git.py
```diff
@@ -13,101 +21,121 @@
...
+            "%s:%s %s",
+            error_msg,
+            os.linesep,
+            exc.stderr.decode().strip() if exc.stderr else "",
+        )
+        raise
+
+
+def is_git_repo(
+    path: str | os.PathLike[str],
```
alexfikl: `papis.paths.PathLike`? You can modify it as needed, of course.

papis/git.py
```diff
@@ -13,101 +21,121 @@
...
+             f"[GIT] Failed to initialize repository in '{path}'")
 
 
-def remove(path: str, resource: str,
-           recursive: bool = False,
-           force: bool = True) -> None:
-    """Remove a *resource* from the git repository at *path*.
+def add(
+    path: str | os.PathLike[str],
+    resource: str | Sequence[str],
```
alexfikl:
```suggestion
    resources: str | Sequence[str],
```
?

alexfikl:
```suggestion
    resource: str | os.PathLike | Sequence[str | os.PathLike],
```
?

tests/commands/test_run.py
```diff
@@ -18,7 +18,7 @@ def test_run_run(tmp_library: TemporaryLibrary) -> None:
     libdir = papis.config.get_lib().path
     # Use a mock scriptlet
     script = os.path.join(os.path.dirname(__file__), "scripts.py")
-    run(libdir, command=[sys.executable, script, "ls"])
+    run(libdir, command=[sys.executable, script, "ls", os.devnull])
```
alexfikl: ?

tests/test_git.py
```diff
@@ -0,0 +1,155 @@
...
+if TYPE_CHECKING:
+    from papis.testing import TemporaryLibrary
+
+pytestmark = pytest.mark.skipif(
+    not shutil.which("git"),
+    reason="Test requires 'git' executable to be in the PATH",
+)
+
+
+def _git(cmd: list[str], cwd: str) -> subprocess.CompletedProcess[str]:
```
alexfikl: Why isn't this using `papis.git._run_git`? might be worth a comment at least.

alexfikl (approved):
Left a few questions and nitpicks, but overall no complaints :grin:

tests/commands/test_run.py
jghauser: The script was actually always failing because it expects exactly two arguments (command + argument). It's just that before, we had no idea it was failing haha. We don't really need the script here at all, so I'm just replacing it with `run(libdir, command=[sys.executable, "-c", "pass"])`, which should be far more transparent.

jghauser: Or do you think running an actual script adds anything?

tests/test_git.py
jghauser: It's because we need to inspect stdout etc in the tests. the `papis.git._run_git` function either returns `None` or raises (because we're using our own logging). I'm adding a comment!

tests/commands/test_run.py
alexfikl: > The script was actually always failing because it expects exactly two arguments (command + argument).

Ah :facepalm: Yeah, fair enough. Then this is fine! I may fix that `scripts` thing when I remember..

tests/test_git.py
alexfikl: Sounds like `_run_git` is just not doing the right thing then. Similar to those `try ... pass` things everywhere. I can't think of any reason not to make it more flexible..

papis/git.py
jghauser: This is about `_run_git` not requiring a `git` argument, right? It's not about the info message?

papis/commands/add.py
jghauser: Good point! I'll rethink it (also regarding your comment about the tests).

papis/git.py
alexfikl: Yeah, yeah, just the calling convention from `git(cmd: Sequence[str], ...)` to `git(*args: str, ...)`.

papis/commands/external.py
jghauser: Your suggestion seems to be correct!

papis/commands/mv.py
jghauser: What do you mean here? Do you mean 'remove + commit'? If so, I didn't add a helper because often we also want to add something (e.g. changes to `info.yaml`). Or do you mean 'mv + commit? If so, I didn't do that because it (to me) implies that we're moving files, and I want to be clear that this doesn't happen (hence, calling the rm function `rm_cached`).

papis/commands/explore.py
jghauser: You're right. Using backticks now.

papis/git.py
jghauser: Done!

papis/git.py
jghauser: Gonna standardise all functions to use `resources` and use (papis versions of) `PathLike | Sequence[PathLike]` (which already is a `Path | str`).

jghauser: See above. I assume you didn't mean `os.PathLike` but the papis versions?

jghauser:
I think I've addressed all your concerns @alexfikl. I didn't do what copilot asked me to (normalising paths to relative paths) since I didn't really find any real evidence that this is needed. Happy to add that too if you think it'd be good.

papis/commands/mv.py
alexfikl: I just meant that the if and else branch had the same `rm_cached` that could be hoisted out? Unless I misread some of the arguments..

alexfikl:
> I didn't do what copilot asked me to (normalising paths to relative paths) since I didn't really find any real evidence that this is needed. Happy to add that too if you think it'd be good.

I think (?) what it was going for is that if you pass an absolute path then it can be outside of the repo and git will fail with some `fatal: path is outside of repository`. I'm not sure that needs an extra check on our side?

papis/commands/mv.py
jghauser: :facepalm: gotcha!

papis/commands/add.py
```diff
@@ -111,6 +111,7 @@
 import papis.cli
 import papis.config
 import papis.logging
+from papis.git import GitError, add_and_commit as git_add_and_commit
```
alexfikl: For what it's worth, I tried reducing these top level imports because each one will make `papis CMD --help` and other commands just a bit slower. It's probably not a big deal, but worth keeping in mind.

papis/git.py
```diff
@@ -1,113 +1,157 @@
...
+) -> list[str]:
+    """Convert a resource or sequence of resources to a list of strings."""
+    if isinstance(resources, (str, os.PathLike)):
+        return [str(resources)]
+    return [str(r) for r in resources]
 
-    logger.info("Adding '%s'.", path)
-    run(["git", "add", resource], cwd=path)
 
+def git(cwd: PathLike, *args: str) -> str:
```
alexfikl: I'm not sure I like the `cwd` as the first argument. It feels counter-intuitive because most places take a `cwd=cwd` or a `-C ...` flag.

I don't feel strongly about it though 🤷 I'm guessing this is done because the other functions also take a path as the first argument already?

jghauser:
> I think (?) what it was going for is that if you pass an absolute path then it can be outside of the repo and git will fail with some `fatal: path is outside of repository`. I'm not sure that needs an extra check on our side?

Oh, I see. Yeah I don't think we need a check for that. We (Papis) are the callers, and I don't think there's a possibility of such paths being generated in the code. I think this adds complexity for not much (or any) gain.

alexfikl:
> Oh, I see. Yeah I don't think we need a check for that. We (Papis) are the callers, and I don't think there's a possibility of such paths being generated in the code. I think this adds complexity for not much (or any) gain.

Yeah, it would definitely be just a way to catch bugs on our side, but I guess the git error will already point in the right direction.

jghauser:
> Yeah, it would definitely be just a way to catch bugs on our side, but I guess the git error will already point in the right direction.

Yeah. Either git will fail or this check will fail. I don't have strong opinions on the matter :shrug:

papis/git.py
```diff
@@ -1,113 +1,157 @@
...
-    run(["git", "commit", "-m", message], cwd=path)
+    try:
+        result = run(["git", *args], cwd=str(cwd), capture_output=True)
+    except subprocess.CalledProcessError as exc:
+        stderr = exc.stderr.strip() if exc.stderr else "(no output)"
+        raise GitError(
+            f"[GIT] 'git {' '.join(args)}' failed in '{cwd}': {stderr}",
+            exc.returncode,
+        ) from exc
+    assert result is not None  # ``wait=True`` in ``run()`` guarantees this
```
alexfikl: I'm guessing this is to please `mypy`? You can add `overloads` to `run` to teach it directly how this works, but that can be done any time..

alexfikl (approved):
This looks good to me!

Left a few comments, but none of them should keep you from merging, I think.

papis/commands/add.py
jghauser: oooh, good point, thanks! I forgot about this and was trying to tidy things up while I'm at it. I'll revert it and keep it in mind in the future as well

papis/git.py
jghauser: Yeah, I didn't think much about it. Happy to make it a `cwd=cwd` thing. You're fine with the other functions that have a `path` as first argument?

papis/git.py
jghauser: Yeah exactly. I added the overload thing, but it's a first for me. If you could have a look that would be excellent :blush:

alexfikl: If `mypy` is happy, it's probably fine. It looks good to me :grin:

jghauser: Thanks!

jghauser:
Thanks a lot, @alexfikl! :heart: Merging this now!

PR #1205, opened 2026-07-08
feat: derive author_list from author in papis.document.new

### LINKED ISSUE #1202: `papis add --set author ...` should derive `author_list` (like `doctor --checks keys-missing --fix` does)
* papis version (`$ papis --version` or commit number): 0.15 (behavior unchanged on current `main`)

# feature request: derive `author_list` when `add` is given only a flat `author`

## current behavior

Adding a document manually with `--set` (a talk, a report, anything with no DOI/ISBN to import from):

```
$ papis add slides.pdf --set author "Jane Doe" --set title "Some talk" --set year 2026 --batch
...
[WARNING] format.python: Could not format pattern '{doc[author_list][0][family]}{doc[year]}' for document ... (Caught exception 'IndexError: string index out of range'. ...)
[ERROR] paths: Could not format path for document. (Caught exception 'FormatFailedError: {doc[year]}_{doc[author_list][0][family]}_...'. ...)
[INFO] paths: Falling back to 'papis_id' as a reference folder name.
```

`--set author` only sets the flat `author` string, so any `ref-format` / `add-folder-name` that uses `{doc[author_list][0][family]}` (the only way to express "surname of first author" in a format pattern) fails, and the document lands in a bare-hash folder like `f0f35b823273534c2a7c3b684d6c83f8` with a fallback ref.

The pieces to fix this already exist — `papis doctor --checks keys-missing --fix` parses `author` into `author_list` via `papis.document.split_authors_name` — but the user has to know to run it as a second step, then rename/re-ref the document afterwards.

## proposed behavior

In `papis add` (and arguably `papis update`), after collecting the `--set` data and before formatting the ref/folder name: if `author` is present and `author_list` is not, derive it with `split_authors_name`, exactly like `doctor`'s `keys-missing` fixer does:

```python
if "author" in data and "author_list" not in data:
    data["author_list"] = papis.document.split_authors_name([data["author"]])
```

(and possibly the symmetric `author_list`-without-`author` case, though in practice every importer that sets `author_list` also sets `author`.)

This would make manual `--set`-only adds produce the same well-formed records the importers do, with no behavior change for documents where an importer or the user already provides `author_list`.

Happy to send a PR if this direction is acceptable.

nasteffe:
Fixes #1202.

Documents created with only a flat `author` string (e.g. via `papis add --set author ...`) were missing the structured `author_list`, so any `ref-format` / `add-folder-name` using `{doc[author_list][0][family]}` failed and fell back to a bare `papis_id` folder name and reference.

As discussed in #1202, this adds the "little if" to `papis.document.new()`: derive `author_list` with `split_authors_name` at document construction, **before** the reference and folder name are generated — the same completion step `new()` already performs for `ref`. An explicitly provided `author_list` is never overwritten, so importers are unaffected.

Includes a test (derivation, no-overwrite, and no-author cases) and a CHANGELOG entry (placed under "Other noteworthy features" to avoid colliding with the new "Bug Fixes" section #1203 introduces; happy to move it once that lands).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,10 @@ user can abort, skip problematic moves, or fix collisions interactively. A new
 
 - Remove undocumented `dirs` option (only allow one `dir` per library).
   ([#1182](https://github.com/papis/papis/pull/1182)).
+- Derive `author_list` from a flat `author` when creating a new document
+  (e.g. `papis add --set author ...`), so formats using
+  `{doc[author_list][0][family]}` work without a manual `papis doctor` pass
+  ([#1202](https://github.com/papis/papis/issues/1202)).
 
 # VERSION v0.15.0 (February 8th, 2026)
 
--- a/papis/document.py
+++ b/papis/document.py
@@ -787,6 +787,12 @@ def new(
     db = get_database()
     db.maybe_compute_id(doc)
 
+    # Derive the structured author list if only a flat author was given
+    # (e.g. `papis add --set author ...`), so that formats using
+    # `author_list` (ref-format, add-folder-name, ...) work below
+    if "author" in doc and "author_list" not in doc:
+        doc["author_list"] = split_authors_name(doc["author"])
+
     # Create a BibTeX reference if missing
     if "ref" not in doc:
         from papis.bibtex import create_reference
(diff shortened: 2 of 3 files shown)


alexfikl:
Thanks for the fix! In it goes :rocket:

PR #1207, opened 2026-07-10
Add IEEE importer

Parsa2820:
## Summary
- Add an `ieee` importer that wraps the IEEE downloader and falls back to DOI-from-PDF + Crossref when IEEE's BibTeX API is blocked.
- Update the IEEE downloader to POST the current `/rest/search/citation/format` endpoint, parse the JSON `data` field, and improve identifier extraction (`abstract/document`, `arnumber=`, `ieee:`).
- Cover match/identifier/BibTeX POST/importer fallback paths with unit tests and wire IEEE into importer matching tests.
## Test plan
- [X] `pytest tests/downloaders/test_ieee.py tests/test_importer.py`
- [X] From an institutional network: `papis add --from ieee https://ieeexplore.ieee.org/document/<id>`
- [X] Same URL with `--download` and confirm PDF lands in the document folder
- [X] Without institutional access: confirm a clear warning (or suggest `papis add --from doi`)


alexfikl:
From a quick look at this, I don't see the point in having both an importer and a downloader for IEEE.
1. If the `/rest/search/citation/format` endpoint is public and accessible, we should just have an importer for it. Similar to how e.g. DBLP works. When does this fail? Is the API documented anywhere?
2. If it doesn't work for some reason, then the fallback to "something else" should be handled in a more general way, not just for the IEEE documents. Where did you hit this?

PR #1210, opened 2026-07-25
Update test for new citeproc default style

alexfikl:
They changed from an older `harvard1` style to a newer `harvard-cite-them-right` style, that messed up our check.

### DIFF
--- a/tests/test_csl.py
+++ b/tests/test_csl.py
@@ -9,7 +9,7 @@ if TYPE_CHECKING:
 
 
 def test_csl_export(tmp_config: TemporaryConfiguration) -> None:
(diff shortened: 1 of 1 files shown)


tests/test_csl.py
```diff
@@ -21,10 +21,25 @@ def test_csl_export(tmp_config: TemporaryConfiguration) -> None:
         "title": "The Theory of Everything",
         "journal": "Nature",
         "year": 2350,
+        "pages": "1-24",
```
alexfikl, line 24: For what it's worth, added the pages here because the new `harvad-cite-them-right` style appends `[PREPRINT]` if the field is missing.

PR #1212, opened 2026-07-25
fix(export): make YAML export appendable

### LINKED ISSUE #1038: Fix line breaks / formatting on `export --append`
Followup for #1036 

Context: https://github.com/papis/papis/pull/1036#issuecomment-3039877652, https://github.com/papis/papis/pull/1036#issuecomment-3040499663, https://github.com/papis/papis/pull/1036#pullrequestreview-2991052748

To summarize: on `export --append`, the resulting outfile is badly formatted in most cases, specifically:

- `bibtex`: no newline at end-of-file, so appended text doesn't have a proper linebreak (see https://github.com/papis/papis/pull/1036#issuecomment-3039877652)
- `csl`: no newline (as above)
- `json`:
    1. no newline (as above)
    2. instead of extending existing list, starts a new one (see https://github.com/papis/papis/pull/1036#issuecomment-3040499663)
- `yaml`: doesn't insert `---` in between documents on append
- `typst`: can't export multiple docs (?)
    - otherwise quite sane, although I suspect it would look better with a blank line in between on append

Sanjays2402:
Closes #1038 (the YAML part).

`export --append` concatenates the exporter output onto the existing file, but the YAML exporter emitted no document separator, so the result parsed as a single merged document — appending a second document silently overwrote the first one's keys rather than adding to the file.

Passing `explicit_start=True` to `yaml.dump_all` prefixes every document with `---`, which makes the output safe to concatenate. New `test_export_yaml_append` fails on `main` (1 document parsed instead of 2) and passes with the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,10 @@ user can abort, skip problematic moves, or fix collisions interactively. A new
 - Fix crash in `papis add` when a given file path contains spaces (the DOI
   importer no longer tries to resolve local paths on doi.org)
   ([#1201](https://github.com/papis/papis/issues/1201)).
+- The YAML exporter now starts each document with an explicit `---` marker, so
+  that `papis export --format yaml --append` produces a valid multi-document
+  file instead of merging the documents together
+  ([#1038](https://github.com/papis/papis/issues/1038)).
 
 # VERSION v0.15.0 (February 8th, 2026)
 
--- a/papis/exporters/yaml.py
+++ b/papis/exporters/yaml.py
@@ -12,8 +12,12 @@ def exporter(documents: list[Document]) -> str:
 
     from papis.document import to_dict
 
+    # NOTE: 'explicit_start' adds a '---' marker before every document, so that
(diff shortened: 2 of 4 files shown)


alexfikl:
This looks good to me! Can you update the existing failing test? I think it just requires a `---` at the beginning of that `re.match`.

Sanjays2402:
Updated in 2b54bec — `test_explore_yaml_cli` now expects a leading `---` line, exactly as you suggested. That was the only test asserting on the raw YAML text.

`tests/commands/test_explore.py` and `tests/commands/test_export.py` are 12 passed locally; reverting the regex change puts it back to the same failure CI saw.

alexfikl:
This looks good to me. Can you rebase? That should fix the CI errors and I'll go ahead and merge it.

Sanjays2402:
Rebased onto main — 4183075. Clean, no conflicts, and the two commits are unchanged in content apart from the new base.

tests/commands/test_explore.py and tests/commands/test_export.py are 12 passed on the rebased tree.

alexfikl:

PR #1219, opened 2026-08-05
feat: more robust pdf fetching for DOI importer

alipatti:
Improves the DOI importer's file fetching by resolving the DOI redirect, checking whether the resolved URL matches an existing downloader,  and then delegating to that downloader if so. If it doesn't match, falls back to existing implementation.

Fixes not fetching pdfs for NBER DOIs (and I imagine also other publishers). See discussion in #1218.

Airport waiting room implementation and could certainly use another pair of eyes (perhaps copilot's)


alipatti:
Accidentally branched off the NBER branch but since that's merged into main we should be fine.

alexfikl:
Hm, I don't think I like this logic very much: (1) it seems specific to what NBER is doing (I don't think many other places redirect to a PDF?) and (2) puts a lot of smarter "pick me a downloader" at the wrong level (I would expect this in `get_matching_importers_by_uri` or something instead).

For the DOI importer, `fetch_files` could just look at the metadata more carefully instead of just at `doc-url-key-name`? For example, the NBER document has this metadata: https://api.crossref.org/works/10.3386/w35546. There's a `resource->primary->URL` in there that has a link to the PDF file. Other documents seem to have it in `link->1` (e.g. https://api.crossref.org/works/10.1515/9783111705446-002).

I'm not sure if the best place for this is here or in `papis/crossref.py` instead though..

PR #1225, opened 2026-08-09
Fix/new doc no empty files field

jghauser:
Fix `document.new()` so  we don't end up with an empty list of files when creating a doc without files.

### DIFF
--- a/papis/document.py
+++ b/papis/document.py
@@ -829,7 +829,8 @@ def new(
 
         document_file_list.append(out_file_name)
 
-    doc["files"] = document_file_list
+    if document_file_list:
+        doc["files"] = document_file_list
     doc.save()
 
     return doc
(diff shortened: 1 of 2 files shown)


papis/document.py
```diff
@@ -829,7 +829,7 @@ def new(
 
         document_file_list.append(out_file_name)
 
-    doc["files"] = document_file_list
+    doc.update({"files": document_file_list})
```
alexfikl: This feels a bit obfuscated, but I guess it's fine 🤷

alexfikl (approved):

papis/document.py
jghauser: Should I add a note explaining why we're doing it like this?

alexfikl: I would prefer doing a `if document_file_list: doc["files"] = document_file_list`, but it's really just a nitpick. Feel free to merge without it!

jghauser: I changed it and now a test is failing :rofl:

jghauser:
Failing test seems to have been transient (and also didn't show on my machine). Can't see how it would have been related to the change. Merging now :).

PR #1226, opened 2026-08-09
Allow passing `PAPIS_DATABASE_BACKEND` to the container when running `papis-run-container-tests`

jghauser:
I wanted to be able to pass the `PAPIS_DATABASE_BACKEND` env to the container when running the tests so I can test the different backends.

This also surfaced a bug with the version strings being incompatible with PEP440 requirements.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -157,7 +157,8 @@
                   +
                   # bash
                   ''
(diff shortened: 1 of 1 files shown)


alexfikl (approved):
Looks good to me! 

Thanks for fixing that `flake` error. I apparently just fixed the `then` branch the other day 🤦

PR #1228, opened 2026-08-13
Unmask config lookup failure, fix test

### LINKED ISSUE #1227: Default config behaviour incoherent
Papis's defaulting behaviour re absent configs is inconsistent across the codebase.
This is the root cause of the second half of https://github.com/papis/papis/issues/948#issuecomment-5254457810.

Specifically, `papis/config.py` defaults to having a minimal configuration, _including_ containing a default library at eg `~/Documents/papers`.
However, in contrast, in eg `papis/commands/default.py`, there seems to be an expectation that the paths given to papis either don't exist or refer to initialized libraries.
This causes trouble in `tests/commands/test_default.py::test_no_config_shows_init_hint`, which expects the `TemporaryConfiguration` to give it a filesystem that is entirely unconfigured. But since `papis/config.py` defaults to a configuration containing a library, if that path coincidentally exists it will cause the test to misbehave.

Multiple aspects are wrong in this picture:
- Most critically, tests should be completely isolated from their build systems. The fact that the existence or nonexistence of a path in `$HOME` affects the test is unacceptable. It seems like `pyfakefs` should solve this.
- papis shouldn't be cavalierly assuming that random paths are in a known state (so the default of `~/Documents/papers` should only crop up in an initial configuration wizard, not in actual program logic)
- The check for the existence of a configuration in `papis/commands/default.py` doesn't match my expectations of its logic -- surely you want to check you A) have a configuration B) that the requested library exists in it and only then C) check that the library exists on the filesystem?

hseg:
The default command's lookup order for config files was inverted, first looking up the local config file, and only then checking if the global config file exists. Moreover, it would entirely skip the check for the global config file if the selected library existed.
In addition, the test that was supposed to exercise this path was resetting the temporary configuration fixture it was receiving, so that if in the test environment the default library path `~/Documents/papers` existed, the error would be erroneously masked.
Finally, despite the absence of the global configuration file being an error, papis would only warn about it and not abort.

Instead, reverse the conditionals to get the correct behaviour and make this situation an error.
As for the test, resetting the configuration is actually *correct*, since there is no way in regular operation to mutate papis's internal state to monkeypatch in the expected erroneous configuration.
Rather, explicitly set papis to try to load a nonexistent config file instead.

Closes: #1227

### DIFF
--- a/papis/commands/default.py
+++ b/papis/commands/default.py
@@ -197,12 +197,11 @@ def run(ctx: click.Context,
                 "configuration file.",
                 library, library.path)
     elif ctx.invoked_subcommand != "init":
-        logger.error("No configuration file exists at '%s'.", config_file)
-        logger.error("Create a configuration file and define your "
-                     "libraries before using Papis. You can use "
-                     "'papis init /path/to/my/library' for a quick "
-                     "interactive setup.")
-        raise click.FileError(config_file, hint="No configuration file exists there")
+        logger.warning("No configuration file exists at '%s'.", config_file)
+        logger.warning("Create a configuration file and define your "
+                       "libraries before using Papis. You can use "
+                       "'papis init /path/to/my/library' for a quick "
+                       "interactive setup.")
 
     # read in configuration from command-line
     sections = configuration.sections()
--- a/papis/commands/default.py
+++ b/papis/commands/default.py
@@ -182,26 +182,27 @@ def run(ctx: click.Context,
         raise SystemExit(1) from None
 
     configuration = papis.config.get_configuration()
-    if os.path.isdir(library.path):
-        # Check if there's a local configuration file in the library, and if so,
-        # merge its contents
-        local_config_file = papis.config.getstring("local-config-file")
-        local_config_path = os.path.join(library.path, local_config_file)
-        papis.config.merge_configuration_from_path(local_config_path, configuration)
-    else:
-        config_file = papis.config.get_config_file()
-        if os.path.exists(config_file):
+    config_file = papis.config.get_config_file()
+    if os.path.isfile(config_file):
+        if os.path.isdir(library.path):
+            # Check if there's a local configuration file in the library, and if so,
+            # merge its contents
+            local_config_file = papis.config.getstring("local-config-file")
+            local_config_path = os.path.join(library.path, local_config_file)
+            papis.config.merge_configuration_from_path(local_config_path, configuration)
+        else:
             logger.error(
                 "Library '%s' directory '%s' does not exist. Please "
                 "create it or update the 'dir' setting in the "
                 "configuration file.",
                 library, library.path)
-        elif ctx.invoked_subcommand != "init":
-            logger.warning("No configuration file exists at '%s'.", config_file)
-            logger.warning("Create a configuration file and define your "
-                           "libraries before using Papis. You can use "
-                           "'papis init /path/to/my/library' for a quick "
-                           "interactive setup.")
+    elif ctx.invoked_subcommand != "init":
+        logger.error("No configuration file exists at '%s'.", config_file)
+        logger.error("Create a configuration file and define your "
+                     "libraries before using Papis. You can use "
+                     "'papis init /path/to/my/library' for a quick "
+                     "interactive setup.")
+        raise click.FileError(config_file, hint="No configuration file exists there")
 
     # read in configuration from command-line
     sections = configuration.sections()
--- a/tests/commands/test_default.py
+++ b/tests/commands/test_default.py
@@ -1,24 +1,46 @@
 from __future__ import annotations
 
+import os
 import shutil
+
+import platformdirs
+import pytest
 
 from papis.testing import PapisRunner, TemporaryConfiguration, TemporaryLibrary
 
 
-def test_no_config_shows_init_hint(tmp_config: TemporaryConfiguration) -> None:
+@pytest.mark.parametrize("with_library", [False, True])
+def test_no_config_shows_init_hint(tmp_config: TemporaryConfiguration,
+                                   monkeypatch: pytest.MonkeyPatch,
+                                   with_library: bool) -> None:
     import papis.config
     from papis.commands.default import run as cli
 
-    shutil.rmtree(tmp_config.configdir)
-    shutil.rmtree(tmp_config.libdir)
-    papis.config.reset_configuration()
+    with monkeypatch.context() as m:
+        # NOTE: this is the default directory set in
+        #   papis.config::Configuration.__init__
+        # See https://github.com/papis/papis/issues/1227
+        m.setattr(platformdirs, "user_documents_dir",
+                  lambda: os.path.join(tmp_config.tmpdir, "Documents"))
+        libdir = os.path.join(platformdirs.user_documents_dir(), "papers")
 
-    cli_runner = PapisRunner()
-    result = cli_runner.invoke(cli, ["list", "-c", tmp_config.configfile])
+        if with_library:
+            os.makedirs(libdir)
+        else:
+            assert not os.path.isdir(libdir)
 
-    assert result.exit_code == 1
-    assert "No configuration file exists at" in result.output
-    assert "papis init" in result.output
+        shutil.rmtree(tmp_config.configdir)
+        papis.config.reset_configuration()
+
+        cli_runner = PapisRunner()
+        result = cli_runner.invoke(cli, ["list"])
+
+        assert result.exit_code == 0
+        assert "No configuration file exists at" in result.output
+        assert "papis init" in result.output
+
+        if with_library:
+            os.rmdir(libdir)
 
 
 def test_default_cli(tmp_library: TemporaryLibrary) -> None:
--- a/tests/commands/test_default.py
+++ b/tests/commands/test_default.py
@@ -10,12 +10,13 @@ def test_no_config_shows_init_hint(tmp_config: TemporaryConfiguration) -> None:
     from papis.commands.default import run as cli
 
     shutil.rmtree(tmp_config.configdir)
+    shutil.rmtree(tmp_config.libdir)
     papis.config.reset_configuration()
 
     cli_runner = PapisRunner()
-    result = cli_runner.invoke(cli, ["list"])
+    result = cli_runner.invoke(cli, ["list", "-c", tmp_config.configfile])
 
-    assert result.exit_code == 0
+    assert result.exit_code == 1
     assert "No configuration file exists at" in result.output
     assert "papis init" in result.output
 
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -4,9 +4,9 @@ sphinx:
   configuration: doc/source/conf.py
 
 build:
-  os: ubuntu-22.04
+  os: ubuntu-24.04
   tools:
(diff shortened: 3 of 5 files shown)


papis/commands/default.py
```diff
@@ -182,26 +182,27 @@ def run(ctx: click.Context,
...
-                           "libraries before using Papis. You can use "
-                           "'papis init /path/to/my/library' for a quick "
-                           "interactive setup.")
+    elif ctx.invoked_subcommand != "init":
+        logger.error("No configuration file exists at '%s'.", config_file)
+        logger.error("Create a configuration file and define your "
+                     "libraries before using Papis. You can use "
+                     "'papis init /path/to/my/library' for a quick "
+                     "interactive setup.")
+        raise click.FileError(config_file, hint="No configuration file exists there")
```
alexfikl: Why is this a hard error now?

My understanding was that papis would work with no configuration, you'd just need to put your library `user_documents_dir() / "papers"`. I'm not sure that's documented anywhere though, but I thought it was on purpose..

tests/commands/test_default.py
```diff
@@ -10,12 +10,13 @@ def test_no_config_shows_init_hint(tmp_config: TemporaryConfiguration) -> None:
     from papis.commands.default import run as cli
 
     shutil.rmtree(tmp_config.configdir)
+    shutil.rmtree(tmp_config.libdir)
```
alexfikl: Maybe add a little comment for this? It's reasonably clear why the config is deleted in a "no config" test, but not so much the `libdir`.

alexfikl (commented):
> In addition, the test that was supposed to exercise this path was resetting the temporary configuration fixture it was receiving

You mean the `papis.config.reset_configuration()`? 

Looking at this fix, it seems like the actual error is just that the testing is broken and allows escaping the temporary directory set up by `TemporaryConfiguration`. I'll look a bit into how to fix that (I thought `papis.config.set_config_file` would do the trick, but apparently not).

Then making the "no config" case a hard error is a different issue.

hseg:
(Can't respond to in-code comments via email because github, responding globally)

> > In addition, the test that was supposed to exercise this path was resetting the temporary configuration fixture it was receiving
>
> You mean the *papis.config.reset_configuration()*?

Yes.

> Looking at this fix, it seems like the actual error is just that the testing is broken and allows escaping the temporary directory set up by *TemporaryConfiguration*. I'll look a bit into how to fix that (I thought *papis.config.set_config_file* would do the trick, but apparently not).

It's both the test _and_ the default command's logic being wrong - the test allows escaping the temporary fixture, the default command's test order has circumstances where it masks the error, so that to trigger a failure you need to hit that masking.

> Then making the "no config" case a hard error is a different issue.

Fair enough - though the absence of documentation of that usecase is what -- possibly unclearly -- motivated my original issue.
I understand papis's workflow as being analogous to git - you have several silos [=repositories, libraries] you work in, and the tool maintains some metadata for each to support its functionality. The difference between the two is that git expects CWD to be inside the silo, whereas papis expects you to call it from an arbitrary location on the system and thus needs to be passed a choice of silo. In this model, it makes no sense to work without reference to a silo - the ability to omit it supported by papis's default-library mechanism is a user convenience, nothing more.
This also means that the defaults suggested by papis init are just that, suggestions, and that papis should refuse to run without initialization.

In contrast, it seems you want to also have papis support an automagical simplified usecase, where the no-configuration setup already picks a sensible default silo and always points there. But that requires different behavior - in particular, it now makes it reasonable to have uninitialized silos (I didn't have a ~/Documents before starting to use papis), and so the guidance on when to initialize becomes much trickier.

I'm fine reverting that change, but I'd like a decision and clarity as to what direction papis will take. Also, such a decision should also indicate when papis should guide users to initialize libraries - perhaps it should distinguish between the no-config and the empty config case?

The deletion of the libdir was a half-baked idea to try and catch the error-masking behavior of the previous implementation, as well as conceptually setting up a call to papis with a completely empty system and a nonexistent expectation that the config file exists. A better solution would require a specification of when precisely papis should issue the initialization hint.

alexfikl:
> You mean the *papis.config.reset_configuration()*?
> Yes.

Ah, ok, that doesn't have much to do with the `TemporaryConfiguration` fixture, but yeah, the `rmdir(configdir)` and that essentially reset the configuration to the `~/Documents/papers` path.

> Fair enough - though the absence of documentation of that usecase is what -- possibly unclearly -- motivated my original issue.

Does `papis init` still work with that hard error? My understanding is that if there is no config and I do `papis init`, it will go in `default.py`, exit with the hard error and give the user no chance to init (except manually).

> I understand papis's workflow as being analogous to git - you have several silos [=repositories, libraries] you work in, and the tool maintains some metadata for each to support its functionality. The difference between the two is that git expects CWD to be inside the silo, whereas papis expects you to call it from an arbitrary location on the system and thus needs to be passed a choice of silo. In this model, it makes no sense to work without reference to a silo - the ability to omit it supported by papis's default-library mechanism is a user convenience, nothing more.

I was thinking of it in the same way as Firefox automatically downloading files to `~/Downloads` if you don't tell it to ask (and probably other similar software). In that view, papis will automatically put new documents in that `papers` library. I guess it would need to create it too if that's the idea..

I'm not sure it's very intuitive though, so fair enough. @jghauser Any thoughts?

alexfikl:
@hseg What do you think about this update for the test? It fails before the fixes in this PR and passes with them (without the hard error part). It also monkeypatches `platformdirs` so that the test doesn't escape the `tmpdir`.
```patch
diff --git a/tests/commands/test_default.py b/tests/commands/test_default.py
index 13672c74..b0ee41fb 100644
--- a/tests/commands/test_default.py
+++ b/tests/commands/test_default.py
@@ -1,23 +1,46 @@
 from __future__ import annotations
 
+import os
 import shutil
 
+import platformdirs
+import pytest
+
 from papis.testing import PapisRunner, TemporaryConfiguration, TemporaryLibrary
 
 
-def test_no_config_shows_init_hint(tmp_config: TemporaryConfiguration) -> None:
+@pytest.mark.parametrize("with_library", [False, True])
+def test_no_config_shows_init_hint(tmp_config: TemporaryConfiguration,
+                                   monkeypatch: pytest.MonkeyPatch,
+                                   with_library: bool) -> None:
     import papis.config
     from papis.commands.default import run as cli
 
-    shutil.rmtree(tmp_config.configdir)
-    papis.config.reset_configuration()
+    with monkeypatch.context() as m:
+        # NOTE: this is the default directory set in
+        #   papis.config::Configuration.__init__
+        # See https://github.com/papis/papis/issues/1227
+        m.setattr(platformdirs, "user_documents_dir",
+                  lambda: os.path.join(tmp_config.tmpdir, "Documents"))
+        libdir = os.path.join(platformdirs.user_documents_dir(), "papers")
 
-    cli_runner = PapisRunner()
-    result = cli_runner.invoke(cli, ["list"])
+        if with_library:
+            os.makedirs(libdir)
+        else:
+            assert not os.path.isdir(libdir)
 
-    assert result.exit_code == 0
-    assert "No configuration file exists at" in result.output
-    assert "papis init" in result.output
+        shutil.rmtree(tmp_config.configdir)
+        papis.config.reset_configuration()
+
+        cli_runner = PapisRunner()
+        result = cli_runner.invoke(cli, ["list"])
+
+        assert result.exit_code == 0
+        assert "No configuration file exists at" in result.output
+        assert "papis init" in result.output
+
+        if with_library:
+            os.rmdir(libdir)
 
 
 def test_default_cli(tmp_library: TemporaryLibrary) -> None:
```

tests/commands/test_default.py
```diff
@@ -10,12 +10,13 @@ def test_no_config_shows_init_hint(tmp_config: TemporaryConfiguration) -> None:
...
     papis.config.reset_configuration()
 
     cli_runner = PapisRunner()
-    result = cli_runner.invoke(cli, ["list"])
+    result = cli_runner.invoke(cli, ["list", "-c", tmp_config.configfile])
```
alexfikl: I'm not sure this `-c ...` does what it should? That flag is marked with `click.Path(exists=True)`, so it should fail before it even gets inside the function.

jghauser:
I don't have a strong preference between the two options (working without config assuming the `~/Documents/papers` dir and requiring a config) as long as it's documented. Though maybe it would be better to error than to create directories (with `papis add`) that the user doesn't expect (or co-opt an existing directory that the user might be using for some other purpose). So maybe failing early and pointing the user to `papis init` is the least surprising and safest path.

We could make `~/Documents/papers` the default suggested dir in `papis init`, but that's a different matter.

alexfikl:
> Though maybe it would be better to error than to create directories (with `papis add`) that the user doesn't expect (or co-opt an existing directory that the user might be using for some other purpose). So maybe failing early and pointing the user to `papis init` is the least surprising and safest path.

I think this is more or less what it currently does? i.e. if the library dir doesn't exist, we don't do any operations at all and error pretty fast (in `papis add` and such).
```
> papis --lib debug list -a
[ERROR] commands.default: Library 'debug' directory '<path>/debug-unknown' does not exist. Please create it or update the 'dir' setting in the configuration file.
[WARNING] commands.list: No documents retrieved
```

I don't think this interacts very well with the "no config" case though. That just seems to show that there's no config, but doesn't complain about the library missing. This is definitely worth fixing :facepalm: 

> We could make ~/Documents/papers the default suggested dir in papis init, but that's a different matter.

The current "default" is the current directory, for what it's worth. I'm not sure either choice is particularly better..

jghauser:
> I think this is more or less what it currently does? i.e. if the library dir doesn't exist, we don't do any operations at all and error pretty fast (in `papis add` and such).

We do create for some reason -- I've checked. I think it might be because apparently `shutil.move()` (in `document.move()`) falls back to `shutil.copytree()` when `os.rename()` fails. And `shutil.copytree()` creates missing folders. Or at least so says AI, this is a bit beyond me :sweat_smile:. In any case, such oddities are kinda an argument for just failing when there's no config.

alexfikl:
> We do create for some reason -- I've checked. I think it might be because apparently `shutil.move()` (in `document.move()`) falls back to `shutil.copytree()` when `os.rename()` fails. And `shutil.copytree()` creates missing folders. Or at least so says AI, this is a bit beyond me 😅. In any case, such oddities are kinda an argument for just failing when there's no config.

TIL.. that's a pretty unfortunate default to have. Should remember never to use `shutil.move` :facepalm: at least not without checking that the parent exists..

hseg:
Haven't yet found the time to look into your remarks, and am heading into a busy week. Hopefully next week will be more doable to get this finalized.

alexfikl:
> Haven't yet found the time to look into your remarks, and am heading into a busy week. Hopefully next week will be more doable to get this finalized.

I can take this over if you don't mind? I was just going to modify the test like in the patch I sent and remove the hard error for a missing config (we can have a separate PR for that?).

hseg:
Would appreciate it, thank you!

alexfikl:
@hseg Will go ahead and merge this when the CI passes. It should hopefully fix the test failures you were seeing in the arch package (at least that `mwe.sh` is passing on this branch and failing on main for me :grin:).

The "no config" case is still quite brittle and untested though. I'll try to look into that a bit during the weekend.

hseg:
Can confirm the tests work here, thanks!

PR #1240, opened 2026-09-09
Share a single bibliography when exporting multiple CSL documents

### LINKED ISSUE #1232: Broken index in some CSL styles
Some styles export a `[NN] ....` sort of string. However, since we export one document at a time, the number is always `NN=1`.

`citeproc-py` supports exporting multiple documents just fine, so we need to fix that.

chiliec:
## What

Closes #1232. Numbered CSL styles (e.g. IEEE) always rendered every exported
entry as `[1]` instead of `[1]`, `[2]`, `[3]`, ...

## Why

`exporter()` called `export_document()` once per document, and each call built
its own `BibliographySource` / `CitationStylesBibliography`. citeproc's
`citation-number` variable is scoped to a bibliography, so it restarted at 1 for
every document — hence the repeated `[1]`. (`citeproc-py` handles multiple
documents in one bibliography fine, as the issue noted.)

## Fix

Added `export_documents()` which builds a single shared `BibliographySource`,
registers every citation into one `CitationStylesBibliography`, and returns the
per-document rendered strings. `export_document()` is now a thin wrapper over it,
so the single-document path is unchanged.

## Tests

Added `test_csl_export_multiple_documents` using a minimal in-repo numbered CSL
style (no network) with two documents, asserting `[1] Doc One` / `[2] Doc Two`.

Validation (Python 3.11, `citeproc-py` 0.11.1):

```
$ python -m pytest tests/test_csl.py -q
3 passed

# genuine RED -> GREEN (revert only papis/exporters/csl.py, keep the test):
$ git stash push papis/exporters/csl.py && python -m pytest tests/test_csl.py::test_csl_export_multiple_documents
AssertionError: assert '[1] Doc One\n\n[1] Doc Two' == '[1] Doc One\n\n[2] Doc Two'
$ git stash pop && python -m pytest tests/test_csl.py -q
3 passed
```

Also confirmed manually with the real IEEE style from Zotero: before the fix two
articles both rendered `[1]…`, after the fix `[1]…` then `[2]…`.

`ruff check` and `mypy` on the changed files both pass. Happy to adjust.


papis/exporters/csl.py
```diff
@@ -200,30 +201,40 @@ def export_document(doc: Document,
...
         logger.warning("Reference with key '%s' not found in the bibliography",
                        item.key)
 
-    bib.cite(citation, callback=warn)
+    citations = [Citation([CitationItem(doc["ref"])]) for doc in documents]
+    for citation in citations:
+        bib.register(citation)
+
+    for citation in citations:
+        bib.cite(citation, callback=warn)
```
alexfikl, line 218: Why are these two loops? Does `citeproc` need all of them to be registered?

papis/exporters/csl.py
```diff
@@ -200,30 +201,40 @@ def export_document(doc: Document,
...
+    return [str(item).replace("..", ".") for item in items]
+
+
+def export_document(doc: Document,
+                    style_name: str | None = None,
+                    formatter_name: str | None = None) -> str:
+    results = export_documents([doc],
+                               style_name=style_name,
+                               formatter_name=formatter_name)
+    return results[0] if results else ""
```
alexfikl, line 237: I don't think this is needed? Just update the test to use `export_documents`.

papis/exporters/csl.py
```diff
@@ -180,12 +180,13 @@ def export_document(doc: Document,
         logger.error("Cannot find style '%s'. You can download or create this "
                      "style yourself and place it in '%s'.",
                      os.path.basename(style_name), get_styles_folder())
-        return ""
+        return [""] * len(documents)
```
alexfikl, line 183: Why does this do a `* len(documents)`? I would expect it to just return `[]` in that case.

alexfikl (commented):
Left a few small comments, but overall this looks good and is welcome :grin:

chiliec:
Thanks @alexfikl! Addressed in 6eeaeff:

- **`export_document` shim** — removed; the test now calls `export_documents([doc])` directly.
- **`[""] * len(documents)`** — all three early-return paths now return `[]`.
- **The two loops** — yes, this is deliberate: citeproc-py needs *all* citations `register`ed before any `cite` call, because bibliography formatting (disambiguation, `citation-number`, subsequent-author-substitute, etc.) depends on the full set being known up front. Registering and citing in a single pass would make each document's formatting see only the documents before it. That's also what the upstream citeproc examples do. Happy to add a short comment there if you'd like.

`tests/test_csl.py` passes (3/3).

alexfikl:

PR #1246, opened 2026-09-15
modify arxiv importer to support alphaxiv

joelsleeba:
Recently I came across alphaxiv which is a community front-end for arxiv. This commit modifies the arxiv importer to parse alphaxiv urls and use the existing pipeline of importing arxiv preprints.


papis/importer/arxiv.py
```diff
@@ -28,7 +28,16 @@ def url(self) -> str:
...
+        from papis.arxiv import (
+            ARXIV_ABS_URL,
+            find_arxivid_in_alphaxiv_url,
+            find_arxivid_in_text,
+            is_arxivid,
+        )
+
+        arxivid = find_arxivid_in_alphaxiv_url(uri)
+        if arxivid:
+            return ArxivImporter(f"{ARXIV_ABS_URL}/{arxivid}", arxivid)
```
alexfikl, line 41: I would be more comfortable checking arXiv first and then this, since I assume it will be less common?

papis/arxiv.py
```diff
@@ -247,6 +283,10 @@ def find_arxivid_in_text(text: str) -> str | None:
 
     :returns: an arXiv identifier, if any could be found, or *None* otherwise.
     """
+    aid = find_arxivid_in_alphaxiv_url(text)
+    if aid:
+        return aid
+
```
alexfikl, line 289: Here too, check arxiv first and then fallback to this.

alexfikl (changes requested):
Two small nitpicks, but overall this looks ok to me.

> Recently I came across alphaxiv which is a community front-end for arxiv. This commit modifies the arxiv importer to parse alphaxiv urls and use the existing pipeline of importing arxiv preprints.

It seems to be more than a frontend? The first paper on the site is https://www.alphaxiv.org/abs/2609.recurrent-looped-transformer and it doesn't have a valid arXiv ID.

papis/importer/arxiv.py
alexfikl: Actually, since `find_arxivid_in_text` falls back to `find_arxivid_in_alphaxiv_url`, is this even needed?

